PR #19, opened 2026-06-08
feat(hermes): add tokensave memory provider

ScriptedAlchemy:
## Summary
- Add a generated Hermes `TokensaveMemoryProvider` that exposes `fact_store` and `fact_feedback` through Tokensave's existing memory tools.
- Configure Hermes installs with `memory.provider: tokensave`, including local activation guidance, named-profile installs, and `--all-profiles` install/uninstall support.
- Add Hermes-focused tests for provider registration, loader-compatible schema shape, local/profile/all-profile config, and uninstall cleanup.

## Test plan
- `cargo fmt`
- `cargo test --test agent_test hermes`
- `cargo test`
- `git diff --check`


src/agents/hermes.rs
```diff
@@ -618,10 +776,24 @@ def make_handler(name: str):
...
 from . import schemas, tools
 
+try:
+    from agent.memory_provider import MemoryProvider
+except Exception:
+    class MemoryProvider:
+        pass
+
+MEMORY_TOOL_MAP = {
+    "fact_store": "tokensave_fact_store",
```
ScriptedAlchemy: Check if this is fully integrated with our holllograic memory interface and Hermes memory provider apis. Check what i offer which is beyond 2 tools.

ScriptedAlchemy: Fixed in b00ac60. The Hermes memory provider now advertises the individual holographic memory operations (add/search/probe/related/reason/contradict/update/remove/list), feedback, and memory_status, with those aliases routed through the existing tokensave fact-store/status implementations.

PR #33, opened 2026-06-13
fix(lcm): preserve Hermes compaction replay

ScriptedAlchemy:
## Summary
- Ensure threshold-triggered LCM compaction can summarize short oversized conversations instead of preserving the entire fresh tail.
- Preserve the Hermes bridge contract for oversized `tracedecay_lcm_compress` responses by compacting payloads without losing replay/control fields.
- Add regression coverage for the short-huge compaction case, oversized MCP response contract, and generated Hermes args-file handling.

## Test plan
- `cargo test --test session_lcm_compression_test`
- `cargo test --test mcp_handler_test lcm -- --nocapture`
- `cargo test --test hermes_lcm_bridge_test`
- `cargo fmt --check`
- Disposable Hermes `compress_context(..., force=false)` harness verified `compressed_backlog`, session rotation, and no no-change warning.

### DIFF
--- a/docs/USER-GUIDE.md
+++ b/docs/USER-GUIDE.md
@@ -247,6 +247,29 @@ provider config, and LCM session storage. If you pass
 targets the named profile instead of the project plugin directory; use this when
 you want to run the command from a project but update a Hermes profile.
 
+#### Verifying Hermes plugin and context-engine changes
+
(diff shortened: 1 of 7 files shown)


ScriptedAlchemy:
Closing to split this work into smaller stacked pull requests.

PR #50, opened 2026-06-18
fix(dashboard): modernize frontend build and LCM UI

ScriptedAlchemy:
Modernizes the dashboard frontend build and UI onto a single Rust-based toolchain (Rspack/Rsbuild) and cleans up the inconsistencies called out in a frontend review. Prod-embed contract is unchanged: \`src/dashboard/assets.rs\` still \`include_bytes!\`/\`include_str!\` the \`dist/\` files at compile time, so the shipped binary serves the UI with **no Node/Rspack dependency at launch** — Rspack runs only at build time.

## Build (Rspack, no esbuild in the pipeline)
- \`dashboard/build.mjs\` now builds the shell + holographic/graph/savings/lcm plugins with \`@rspack/core\` (single-file IIFE per plugin; React externalized onto the host SDK via in-tree shims; exact \`dist/\` paths preserved; Hermes-wrapper concatenation preserved).
- Holographic CSS compiled with **real Tailwind v4** (\`@tailwindcss/node\` + \`@tailwindcss/oxide\`; \`@layer theme\`/\`base\` stripped so the plugin never clobbers host \`:root\` vars, wrapped in \`@layer hermes-plugin\`). Deletes the ~390-line hand-rolled utility subset (the fragile \`[class*=\"xl:grid-cols-[minmax(0,1.25fr)_minmax(0,1fr)]\"]\` attribute selectors). A small CSS compactor replaces esbuild for minify (preserves \`@supports color-mix\`).
- esbuild is gone from the build (the esbuild fallback builder + \`build:esbuild\` script are removed); esbuild remains a devDep only for the unit-test bundler helper.

## Plugins
- **LCM**: the 2020-line hand-written vanilla IIFE is ported to TSX (entry/App/components/markdown/helpers), built like the other plugins. Bundle 86 KB → 48 KB. All behavior + \`hermes-lcm-*\` classes preserved.
- **Canonical \`cn\`** (\`lib/cn.ts\`): one class-name joiner for the shell + lib + holographic's self-contained copy (pinned by \`test/shell-sdk.test.mjs\`).
- **Shared UI primitives** (\`lib/primitives.tsx\` + \`.css\`): \`EmptyState\`, \`ErrorPanel\`, \`SkeletonLines\`, \`Stat\`, \`BarList\` — adopted in graph as the reference pattern (delivered via a new \`buildPlugin\` \`primitives\` CSS-prepend option so the CSS rides into both standalone and Hermes).

## Shell / a11y
- **Light-theme collapse**: the ~120-line block of 20 per-component \`[data-theme=\"light\"]\` overrides → 32 semantic tokens that flip with theme. Dark + light computed values are byte-identical. New components no longer need a manual light override.
- a11y: \`CoverageGauge\` gets \`role=\"img\"\` + \`aria-label\`; savings view-switch is a proper \`role=\"tablist\"\`/\`role=\"tab\"\`/\`aria-selected\`; \`VIEW_TABS\` hoisted to module scope.

## Dev server
- New \`npm run dev\` (\`dashboard/dev/run.mjs\`) — Rsbuild HMR dev server proxying \`/api/*\` to a running \`tracedecay dashboard\` (\`TRACEDECAY_DEV_API\` / \`TRACEDECAY_DEV_PORT\`). The dev entry builds the SDK on \`window\` then imports plugin entries directly. NOTE: in some sandboxes both Rsbuild Tailwind-v4 integrations (\`@rsbuild/plugin-tailwindcss\` and \`@tailwindcss/postcss\`) hit a native segfault in \`createRsbuild()\`, so the shipped dev server is \`pluginReact()\`-only — holographic renders unstyled in dev there; the prod build is the source of truth for its Tailwind styles. Verify \`npm run dev\` in your env.

## Verification
- \`npm test\`: 100/100 node + 12/12 vitest.
- \`npm run build\`: all 14 artifacts; cargo embed + dashboard serve all routes 200 (served byte sizes match Rspack output).
- Real-browser Playwright smoke (desktop + narrow): Holographic/Similarity/Curation/Code Graph/LCM all render incl. the new TSX LCM bundle.

## Not in scope (follow-ups)
- Adopt shared primitives in savings/lcm (holographic stays self-contained).
- Consolidate the prod \`@rspack/core\` script onto idiomatic Rsbuild multi-environment builds.
- Migrate the unit-test bundler off esbuild (would remove the last esbuild devDep at the cost of slower tests).

Branch based on \`profile-storage\` so the diff is UI-only. Retarget to \`master\` if you prefer.


ScriptedAlchemy:
Closing this as superseded by the split PR stack that has already merged into master (#56, #57, #58, #59, #60, #61) plus the release-plz v0.0.5 release work. This aggregate branch is still based on profile-storage and its current tree would now undo/remove current master release automation artifacts, so it is no longer a safe merge target.

PR #121, opened 2026-06-24
feat: support linked worktree store resolution

spa5k:
Previously, each linked worktree needed its own TraceDecay initialization to behave reliably. This change lets TraceDecay recognize linked git worktrees as part of the already-initialized repository, then use that initialized repo context while scanning and syncing the current worktree's files and branch.

```mermaid
flowchart TD
    A[Main checkout] -->|tracedecay init once| B[Initialized TraceDecay repo context]
    C[Linked worktree: feature-a] -->|git common dir match| B
    D[Linked worktree: feature-b] -->|git common dir match| B
    C -->|sync/search from worktree root| E[Feature-a branch graph]
    D -->|sync/search from worktree root| F[Feature-b branch graph]
    B --> E
    B --> F
```

## Summary

- Resolve linked git worktrees through the initialized repository store by git common directory.
- Keep worktree paths as the active project root while sharing graph, session, and branch storage.
- Add focused coverage for sibling and nested worktrees.

## Motivation

Codex, Conductor, Cursor, and plain `git worktree` workflows create separate working directories for the same repository. TraceDecay should recognize those linked worktrees as the same initialized repo context while still operating on the invoking worktree's files.

## Changes

- Stop project-root discovery from walking past the current git worktree root.
- Use registry/git-common-dir store lookup in init/sync/hook guards and session DB resolution.
- Make branch commands resolve metadata through `TraceDecay::open` when possible.
- Document linked worktree behavior in the README, user guide, and changelog.

### DIFF
--- a/.changeset/worktree-store-resolution.md
+++ b/.changeset/worktree-store-resolution.md
@@ -1,0 +1,5 @@
+---
+"tracedecay": patch
+---
+
+Support linked git worktrees resolving the initialized repository store by git common directory.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+
+- support linked git worktrees resolving the initialized repository store by git common directory
+
 ## [0.0.12](https://github.com/ScriptedAlchemy/tracedecay/compare/v0.0.11...v0.0.12) - 2026-06-24
 
 ### Fixed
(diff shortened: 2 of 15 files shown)


spa5k:
I wanted to confirm this a bit, earlier i was adding .tracedecay to the .worktreeinclude so that the tracedecay is initialized in each worktree, but I thought we can skip both if we just handle it directly through the native worktree setup. That's why I created this PR, so that I dont need to bother running `tracedecay init` or handle the worktreeinclude

spa5k:
Do you generally manually run it per worktree or agent runs it for you?

ScriptedAlchemy:
no, the idea is one unified thing across all branches and worktreesx, so this is right

PR #210, opened 2026-07-02
Harden Cursor MCP integration: degraded serve mode + doctor log diagnostics

ScriptedAlchemy:
## Summary

Hardening informed by a real failure debugged on 2026-07-02: Cursor spawns plugin MCP servers with cwd = user home, headless agent-session scopes can pass a literal `${workspaceFolder}`, and — critically — **Cursor never retries a failed MCP scope**, so a single `serve` startup exit turned every later tool call in the session into "Timed out waiting for connection" until the user toggled the server or reloaded the window.

This PR is scoped to what happens **after** path/config resolution fails; the literal-`${workspaceFolder}` path interpretation itself is handled separately on `codex/serve-literal-workspacefolder` (the two branches touch adjacent but disjoint code in the `serve` startup path).

- **`serve` no longer exits on recoverable startup resolution failures** (uninitialized project, ambiguous global fallback, bad explicit `--path`). It completes the MCP handshake, lists the real tools, and answers each `tools/call` with an actionable error naming the failure, the fix, and the `tracedecay tool …` CLI fallback. Every tool call retries resolution, so the server **recovers in-session** after `tracedecay init` — no toggle or window reload. Truly unrecoverable transport failures (stdin/stdout I/O errors) still exit.
- **`tracedecay doctor --agent cursor` now catches this failure class**: a best-effort scan of Cursor's recent MCP logs (`~/.config/Cursor/logs`, macOS/Windows equivalents) for literal-`${workspaceFolder}` spawn errors, `Connection failed: MCP error -32000` lines, and degraded-mode notices, with the concrete remediation (toggle the MCP server in Cursor Settings → MCP or reload the window). Also warns when the installed plugin bundle version differs from the binary (`tracedecay update-plugin`). Missing/odd platform layouts degrade silently to no findings.
- **Docs**: `cursor-plugin/README.md` gains a "Troubleshooting a dead MCP scope" section; changelog entries added.

Verified against this machine's real Cursor logs: doctor reports 75 literal-placeholder spawns, 39 failed connections, and a true-positive bundle-version mismatch. A manual `serve --path '${workspaceFolder}'` from a home-like cwd now completes the handshake and returns the actionable error instead of dying.

## Test plan

- [x] New stdio integration tests in `mcp_suite::mcp_cli_serve_test`: degraded handshake + actionable tool error for an explicit uninitialized `--path` (still never serves the global-fallback project), ambiguity stays alive and reports "Multiple tracedecay projects found", and in-session recovery after `tracedecay init` (next tool call served for real).
- [x] New `agents::cursor_logs` unit tests against fixture log files mirroring the real 2026-07-02 failure: literal placeholder + `-32000` detection, tracedecay-only filtering of the shared `mcpprocess.log`, newest-sessions window, silent no-log behavior, platform log roots.
- [x] `plugin_version_staleness` unit test.
- [x] `cargo check --all-targets`, `cargo clippy --all-targets` (no new warnings), `cargo fmt --check`.
- [x] `mcp_cli_serve_test` (15/15), doctor + `agents::` unit tests (154/154), `update_plugin_test` (19/19) — all pass. Hermetic via the `.cargo/config.toml` `TRACEDECAY_DATA_DIR` pin.

Do not merge yet — a thermo review pass runs afterwards.


ScriptedAlchemy:
Addressed the review rounds in the latest push:

- **Codex P2 (recovery never re-runs discovery):** fixed. `ServeProjectResolver` re-runs the full startup resolution ladder on every degraded tool call — explicit path, fresh cwd walk-up, remembered MCP initialize roots, then the global registry (unique-only in discarded-template mode). Pinned by `serve_degraded_mode_test::literal_template_from_home_recovers_after_project_registration` ($HOME scope → init+register → next call served).
- **Duplicated MCP dispatch:** protocol shapes now single-sourced in the `mcp` module (`McpMethod` classifier, `initialize_result`, `resources_list_result`) shared by `McpServer::handle_request` and the new `mcp::degraded` responder.
- **Pipelined handoff loss:** recovery keeps serving on the same `ReplayTransport` (new in `mcp::transport`), so requests buffered behind the recovery-triggering call are answered, not dropped (`pipelined_requests_survive_recovery_handoff`).
- **Marker constant:** doctor's log scanner imports `serve::DEGRADED_SERVE_STDERR_MARKER` instead of re-typing it; Cursor doctor helpers consolidated into `agents::cursor_diagnostics`.
- **Structure:** `main.rs`'s Serve arm is now one `serve::run_serve()` call; degraded stdio tests split into `serve_degraded_mode_test.rs` on the shared harness; PR #206 merged in and its exit-pinning template tests reconciled to stay-alive semantics (no-silent-fallback and ambiguity assertions preserved).

PR #295, opened 2026-07-05
fix: harden automation self-improvement outputs

ScriptedAlchemy:
## Summary
- harden TraceDecay automation self-improvement outputs and managed-skill exports
- make memory curator and session reflector auto-apply honor `auto_apply_memory_ops` even when dashboard approval remains enabled
- export managed skills with Codex-loadable `name`/`description` frontmatter and stop defaulting generated skills to Cursor
- fix native skill overlay manifests to record final paths, not temp staging paths
- make generic markdown tables report empty visible fields instead of rendering blank
- update codex handoff readiness for auto-applied records and dashboard backend persistence coverage

## Renderer audit notes
`tracedecay_context` is the good markdown baseline. Remaining renderer opportunities: custom renderers for run artifacts, skill list/view, project/fact/LCM outputs, and callers/impact instead of generic JSON-ish markdown.

## Thermo review status
Subagents are reviewing current branch structure, automation policy, managed-skill exports, renderer behavior, tests, and CI/base shape. Early high-conviction finding: native managed-skill export should split host SKILL frontmatter from TraceDecay storage metadata instead of emitting both through one renderer.

## Tests
- cargo fmt --check
- cargo test -p tracedecay codex_handoff_reports_auto_applied_records_without_approval_gate -- --nocapture
- cargo test -p tracedecay table_with_no_visible_columns_is_not_blank -- --nocapture
- cargo test -p tracedecay --test automation_runner_test session_reflector_runner_auto_applies_facts_when_dashboard_approval_is_required -- --nocapture
- cargo test -p tracedecay --test automation_runner_test memory_curator_runner_auto_applies -- --nocapture
- cargo test -p tracedecay --test agent_suite skill_targets_test -- --nocapture
- cargo test -p tracedecay --test agent_suite test_cursor_install_exports_active_managed_skills -- --nocapture

Note: ran `cargo clean` as requested; it removed the repo target symlink, then I restored `target -> /scratch/cargo-target/tracedecay` per AGENTS.md.

### DIFF
--- a/.claude/skills/inspecting-automation-cycles/SKILL.md
+++ b/.claude/skills/inspecting-automation-cycles/SKILL.md
@@ -18,7 +18,7 @@ whether generated output was adopted.
 3. For failures or suspicious skips, open the relevant artifact with
    `tracedecay_automation_run_artifact_view` or
    `tracedecay tool automation_run_artifact_view --args ...`.
-4. Inspect pending human review queues:
+4. Inspect pending self-managed review/apply state:
    `tracedecay automation facts list`, dashboard approvals, and
    `tracedecay_skill_list --state pending`.
 5. Check adoption evidence: `tracedecay analytics diagnostics --all --no-sync`,
@@ -49,6 +49,6 @@ whether generated output was adopted.
 
 ## Deliverable
 
-Report task/status counts, the exact run or artifact ids inspected, pending
-approval queues, adoption gaps, and the next concrete command for any mutation
+Report task/status counts, the exact run or artifact ids inspected, staged
+fact/skill state, adoption gaps, and the next concrete command for any mutation
 the user should choose.
--- a/.claude/skills/inspecting-automation-cycles/SKILL.md
+++ b/.claude/skills/inspecting-automation-cycles/SKILL.md
@@ -1,0 +1,54 @@
+---
+name: inspecting-automation-cycles
+description: 'TraceDecay Dev: Use when auditing TraceDecay automation loops, skipped runs, memory-curator/session-reflector/skill-writer output, pending approvals, or run artifacts.'
+---
+
+# TraceDecay Dev: Inspecting Automation Cycles
+
+TraceDecay automation is a loop, not a single artifact: config schedules jobs,
+runs produce artifacts, dashboards queue approvals, and usage analytics prove
+whether generated output was adopted.
+
+## Workflow
+
+1. Start with `tracedecay automation config get` to identify enabled tasks,
+   schedules, locks, profile paths, and approval gates.
+2. List recent runs with `tracedecay automation runs list --limit 100`; group
+   by task and status before opening individual artifacts.
+3. For failures or suspicious skips, open the relevant artifact with
+   `tracedecay_automation_run_artifact_view` or
+   `tracedecay tool automation_run_artifact_view --args ...`.
+4. Inspect pending human review queues:
+   `tracedecay automation facts list`, dashboard approvals, and
+   `tracedecay_skill_list --state pending`.
+5. Check adoption evidence: `tracedecay analytics diagnostics --all --no-sync`,
+   `tracedecay sessions search "mcp__tracedecay" --provider all`, and managed
+   skill usage counts.
+
+## Reading Results
+
+| Signal | Meaning | Next step |
+|---|---|---|
+| `scheduler_interval_not_elapsed` | Healthy throttling | Count only, do not fix. |
+| `scheduler_lock_active` | Another run owns the loop | Check age before calling stale. |
+| `no_new_session_activity` | Nothing new to process | Verify transcript ingest if surprising. |
+| `validation_gate` artifact | Proposed mutation was staged | Report approval command or dashboard path. |
+| Many pending fact proposals | Queue needs curation | Use `tracedecay:project-memory`. |
+| Active managed skills with zero use | Adoption telemetry gap | Use `tracedecay:diagnosing-analytics`. |
+
+## Guardrails
+
+- Prefer read-only inspection. Do not approve, reject, delete, or apply queued
+  facts unless the user explicitly asked for mutation.
+- Do not treat skipped runs as failures until grouped by skip reason and age.
+- Avoid parallel `tracedecay_skill_view` calls against one profile while
+  automation may write usage ledgers. If a usage read reports a truncated JSON
+  or EOF parse error, retry once after `tracedecay_skill_list` succeeds.
+- Do not read `.tracedecay` databases directly; use CLI, dashboard APIs, or
+  MCP tools.
+
+## Deliverable
+
+Report task/status counts, the exact run or artifact ids inspected, pending
+approval queues, adoption gaps, and the next concrete command for any mutation
+the user should choose.
--- a/.claude/skills/inspecting-automation-cycles/SKILL.md
+++ b/.claude/skills/inspecting-automation-cycles/SKILL.md
@@ -1,25 +1,25 @@
 ---
 name: inspecting-automation-cycles
-description: 'TraceDecay Dev: Use when auditing TraceDecay automation loops, skipped runs, memory-curator/session-reflector/skill-writer output, pending approvals, or run artifacts.'
+description: 'TraceDecay Dev: Use when auditing TraceDecay automation loops, skipped runs, memory-curator/session-reflector/skill-writer output, apply policy, or run artifacts.'
 ---
 
 # TraceDecay Dev: Inspecting Automation Cycles
 
 TraceDecay automation is a loop, not a single artifact: config schedules jobs,
-runs produce artifacts, dashboards queue approvals, and usage analytics prove
-whether generated output was adopted.
+runs produce artifacts, dashboards expose outcomes/telemetry, and usage
+analytics prove whether generated output was adopted.
 
 ## Workflow
 
 1. Start with `tracedecay automation config get` to identify enabled tasks,
-   schedules, locks, profile paths, and approval gates.
+   schedules, locks, profile paths, and apply policy.
 2. List recent runs with `tracedecay automation runs list --limit 100`; group
    by task and status before opening individual artifacts.
 3. For failures or suspicious skips, open the relevant artifact with
    `tracedecay_automation_run_artifact_view` or
    `tracedecay tool automation_run_artifact_view --args ...`.
-4. Inspect pending self-managed review/apply state:
-   `tracedecay automation facts list`, dashboard approvals, and
+4. Inspect model-managed memory outcomes plus configured managed-skill review
+   queues: `tracedecay automation facts list`, dashboard telemetry, and
    `tracedecay_skill_list --state pending`.
 5. Check adoption evidence: `tracedecay analytics diagnostics --all --no-sync`,
    `tracedecay sessions search "mcp__tracedecay" --provider all`, and managed
@@ -32,14 +32,15 @@ whether generated output was adopted.
 | `scheduler_interval_not_elapsed` | Healthy throttling | Count only, do not fix. |
 | `scheduler_lock_active` | Another run owns the loop | Check age before calling stale. |
 | `no_new_session_activity` | Nothing new to process | Verify transcript ingest if surprising. |
-| `validation_gate` artifact | Proposed mutation was staged | Report approval command or dashboard path. |
-| Many pending fact proposals | Queue needs curation | Use `tracedecay:project-memory`. |
+| `validation_gate` artifact | Mutation passed validation | Inspect apply-policy state or dashboard artifact. |
+| Many fact proposal records | Inspect validation/apply telemetry | Use `tracedecay:project-memory`. |
 | Active managed skills with zero use | Adoption telemetry gap | Use `tracedecay:diagnosing-analytics`. |
 
 ## Guardrails
 
-- Prefer read-only inspection. Do not approve, reject, delete, or apply queued
-  facts unless the user explicitly asked for mutation.
+- Prefer read-only inspection. Do not mutate fact records.
+- Do not approve, reject, delete, or apply managed-skill drafts unless the user
+  explicitly asked for mutation.
 - Do not treat skipped runs as failures until grouped by skip reason and age.
 - Avoid parallel `tracedecay_skill_view` calls against one profile while
   automation may write usage ledgers. If a usage read reports a truncated JSON
@@ -49,6 +50,6 @@ whether generated output was adopted.
 
 ## Deliverable
 
-Report task/status counts, the exact run or artifact ids inspected, staged
-fact/skill state, adoption gaps, and the next concrete command for any mutation
-the user should choose.
+Report task/status counts, the exact run or artifact ids inspected, apply-policy
+state, adoption gaps, and the next concrete command for any mutation the user
+should choose.
--- a/.gitignore
+++ b/.gitignore
@@ -4,8 +4,7 @@
 .tracedecay
 .mcp.json
 !plugin/.mcp.json
-.claude/*
-!.claude/skills/
+.claude/settings.local.json
 .cursor/
 docs/superpowers
 /local-only
--- a/.gitignore
+++ b/.gitignore
@@ -4,7 +4,8 @@
 .tracedecay
 .mcp.json
 !plugin/.mcp.json
-.claude/
+.claude/*
+!.claude/skills/
 .cursor/
 docs/superpowers
 /local-only
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -20,3 +20,16 @@ CARGO_TARGET_DIR="${RUNNER_TEMP:-/tmp}/tracedecay-cargo-target" \
 TRACEDECAY_DATA_DIR="${RUNNER_TEMP:-/tmp}/tracedecay-test-profile/.tracedecay" \
 cargo test-all
 ```
+
+## Learned User Preferences
+
+- Do not merge a batch of PRs until aggregate verification is stable; a single flaky pass is not enough.
+- Delegate code edits to execution-focused subagents; use planning/review-focused agents for planning, review, and thinking.
+- When orchestrating parallel agents, the lead dictates exact scoped edits, subagents execute, and the lead reviews diffs before any push.
+- Subagents should not invent scope beyond what the lead dictated.
+
+## Learned Workspace Facts
+
+- Parallel branch work uses git worktrees under `.worktrees/` in the repo root (for example `.worktrees/codex-cli-args-stdin`).
+- Integration/default branch is `master`, not `main` (GitHub: ScriptedAlchemy/tracedecay).
+- Multi-PR merge verification: build a detached scratch worktree on `origin/master`, merge all target branches, then run tests with isolated `CARGO_TARGET_DIR` and `TRACEDECAY_DATA_DIR` paths.
--- a/src/automation/managed_skill_model.rs
+++ b/src/automation/managed_skill_model.rs
@@ -56,7 +56,6 @@ impl SkillInstallTarget {
 
 pub fn default_managed_skill_targets() -> Vec<SkillInstallTarget> {
     vec![
-        SkillInstallTarget::Cursor,
         SkillInstallTarget::Codex,
         SkillInstallTarget::Claude,
         SkillInstallTarget::Agents,
@@ -64,6 +63,21 @@ pub fn default_managed_skill_targets() -> Vec<SkillInstallTarget> {
         SkillInstallTarget::Kimi,
         SkillInstallTarget::Kiro,
     ]
+}
+
+fn managed_skill_description(summary: &str) -> String {
+    let trimmed = summary.trim();
+    if trimmed
+        .get(..8)
+        .is_some_and(|prefix| prefix.eq_ignore_ascii_case("use when"))
+        || trimmed
+            .get(..19)
+            .is_some_and(|prefix| prefix.eq_ignore_ascii_case("use this skill when"))
+    {
+        trimmed.to_string()
+    } else {
+        format!("Use when {trimmed}")
+    }
 }
 
 #[derive(Debug, Clone, Copy, PartialEq, Eq, Serialize, Deserialize)]
@@ -283,6 +297,12 @@ impl ManagedSkill {
     pub fn render_skill_markdown(&self) -> String {
         let mut output = String::new();
         output.push_str("---\n");
+        let _ = writeln!(output, "name: {}", self.metadata.id);
+        let _ = writeln!(
+            output,
+            "description: {}",
+            frontmatter_string(&managed_skill_description(&self.metadata.summary))
+        );
         let _ = writeln!(output, "id: {}", self.metadata.id);
         let _ = writeln!(
             output,
--- a/src/automation/managed_skill_model.rs
+++ b/src/automation/managed_skill_model.rs
@@ -9,6 +9,7 @@ use crate::errors::Result;
 use super::managed_skill_format::{frontmatter_string, source_key, state_key, target_key};
 use super::managed_skill_validation::{
     validate_managed_skill, validate_native_skill_markdown, validate_support_file,
+    MAX_NATIVE_SKILL_DESCRIPTION_CHARS, MAX_NATIVE_SKILL_NAME_CHARS,
 };
 
 pub const MAX_MANAGED_SUPPORT_FILES: usize = 20;
@@ -70,7 +71,7 @@ pub fn default_managed_skill_targets() -> Vec<SkillInstallTarget> {
 
 fn managed_skill_description(summary: &str) -> String {
     let trimmed = summary.trim();
-    if trimmed
+    let description = if trimmed
         .get(..8)
         .is_some_and(|prefix| prefix.eq_ignore_ascii_case("use when"))
         || trimmed
@@ -80,7 +81,41 @@ fn managed_skill_description(summary: &str) -> String {
         trimmed.to_string()
     } else {
         format!("Use when {trimmed}")
+    };
+    truncate_frontmatter_chars(&description, MAX_NATIVE_SKILL_DESCRIPTION_CHARS)
+}
+
+fn native_skill_name(id: &str) -> String {
+    let mut normalized = String::with_capacity(id.len().min(MAX_NATIVE_SKILL_NAME_CHARS));
+    for byte in id.bytes() {
+        match byte {
+            b'a'..=b'z' | b'0'..=b'9' => normalized.push(byte as char),
+            b'-' | b'_' if !normalized.ends_with('-') => normalized.push('-'),
+            _ => {}
+        }
     }
+
+    let trimmed = normalized.trim_matches('-');
+    let truncated = truncate_frontmatter_chars(trimmed, MAX_NATIVE_SKILL_NAME_CHARS);
+    let name = truncated.trim_end_matches('-');
+    if name.is_empty() {
+        "skill".to_string()
+    } else {
+        name.to_string()
+    }
+}
+
+fn truncate_frontmatter_chars(value: &str, max_chars: usize) -> String {
+    if value.chars().count() <= max_chars {
+        return value.to_string();
+    }
+
+    value
+        .chars()
+        .take(max_chars)
+        .collect::<String>()
+        .trim_end()
+        .to_string()
 }
 
 #[derive(Debug, Clone, Copy, PartialEq, Eq, Serialize, Deserialize)]
@@ -347,7 +382,7 @@ impl ManagedSkill {
     pub fn render_native_skill_markdown(&self) -> Result<String> {
         let mut output = String::new();
         output.push_str("---\n");
-        let _ = writeln!(output, "name: {}", self.metadata.id);
+        let _ = writeln!(output, "name: {}", native_skill_name(&self.metadata.id));
         let _ = writeln!(
             output,
             "description: {}",
--- a/src/automation/managed_skill_model.rs
+++ b/src/automation/managed_skill_model.rs
@@ -7,7 +7,9 @@ use sha2::{Digest, Sha256};
 use crate::errors::Result;
 
 use super::managed_skill_format::{frontmatter_string, source_key, state_key, target_key};
-use super::managed_skill_validation::{validate_managed_skill, validate_support_file};
+use super::managed_skill_validation::{
+    validate_managed_skill, validate_native_skill_markdown, validate_support_file,
+};
 
 pub const MAX_MANAGED_SUPPORT_FILES: usize = 20;
 pub const MAX_MANAGED_SUPPORT_FILE_BYTES: usize = 64 * 1024;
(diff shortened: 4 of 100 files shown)


.gitignore
```diff
@@ -4,7 +4,8 @@
 .tracedecay
 .mcp.json
 !plugin/.mcp.json
-.claude/
+.claude/*
+!.claude/skills/
```
ScriptedAlchemy: **🔴 CONFIRMED — Narrowing '.claude/' to '.claude/*' with '!.claude/skills/' re-includes the entire .claude/skills tree for every contributor, not just the five repo skills this PR adds, so previously-ignored local/personal Claude skills become trackable**

A contributor (or TraceDecay's own tooling) keeps machine-local skills under .claude/skills/ — previously invisible to git; after this PR they show as untracked noise in every git status and a habitual 'git add .' or 'git add -A' commits private local skill instructions (which can embed absolute paths, workflow details, or org-specific prompts) to the public repo.

**Verifier evidence:** At d0c7e9e9, .gitignore lines 7-8 read `.claude/*` and `!.claude/skills/` (base origin/master had `.claude/` which ignored the whole tree). The negation is directory-wide with no per-skill scoping, and no other pattern in the file matches paths under .claude/skills/, so ALL files under .claude/skills/ — not just the five SKILL.md files the PR adds (confirmed via `git diff --name-only origin/master...d0c7e9e9 -- .claude/`: inspecting-automation-cycles, interpreting-tracedecay-diagnostics, introsp

_max-effort adversarial review · finder+independent verifier · Claude Code_

src/automation/runner.rs
```diff
@@ -356,22 +357,16 @@ async fn finalize_session_reflector_success(
...
+            .to_json();
+    if let Some(object) = session_fact_apply_policy.as_object_mut() {
+        object.insert(
+            "applied_proposal_ids".to_string(),
+            json!(applied_proposal_ids),
+        );
+        object.insert("applied_fact_ids".to_string(), json!(applied_fact_ids));
+    }
     let report = json!({
         "status": if auto_apply_facts && accepted_count > 0 { "auto_applied" } else { "needs_approval" },
```
ScriptedAlchemy: **🔴 CONFIRMED — Session reflector report claims status 'auto_applied' and dry_run=false based on config alone, even when zero fact proposals were actually applied**

auto_apply_memory_ops on, accepted_count>0, but every proposal record comes back from auto_apply_session_fact_proposals in a non-Applied state (none were PendingApproval, or apply produced non-Applied outcomes): auto_apply_facts is true so report.status='auto_applied' and dry_run=false, and MemoryApplyPolicy::session_facts's decision() reports 'auto_apply_allowed' — while mutates_store=false and applied_proposal_ids is empty. Dashboards/run artifacts show a run that 'auto applied' facts that never landed in the store.

**Verifier evidence:** runner.rs line 337: `let auto_apply_facts = MemoryApplyPolicy::should_apply(config, accepted_count);` where should_apply is only `accepted_count > 0 && config.auto_apply_memory_ops` (apply_policy.rs line 115). Lines 371-372: `"status": if auto_apply_facts && accepted_count > 0 { "auto_applied" } else { "needs_approval" }, "dry_run": !(auto_apply_facts && accepted_count > 0),` — computed from config/validation alone, never from apply results. Concrete triggers where zero facts land: (1) fact_prop

_max-effort adversarial review · finder+independent verifier · Claude Code_

src/doctor/registry_drift.rs
```diff
@@ -63,6 +63,146 @@ pub(super) async fn registry_drift_findings(
...
+        .manifest_path
+        .parent()
+        .map(|parent| parent.join(crate::config::CONFIG_FILENAME));
+    let mut config_rewritten = None;
+    if let Some(config_path) = config_path {
+        if config_path.is_file() {
+            match reconcile_config_root_dir(&config_path, canonical, &finding.registry_value) {
+                Ok(true) => config_rewritten = Some(config_path),
+                Ok(false) => {}
+                Err(message) => return Err(message),
```
ScriptedAlchemy: **🔴 CONFIRMED — When the manifest rewrite succeeds but the sibling config.json rewrite fails, reconcile_one_store_root returns Err and the already-performed manifest mutation is dropped from the reconciled report**

Doctor heal pass on a drifted store where config.json is unreadable/read-only: store_manifest.json is rewritten to the canonical root (real mutation on disk), then reconcile_config_root_dir errors; the function returns Err so the heal report lists only a warning and zero reconciled stores — the operator is told nothing was healed while the manifest was in fact rewritten, and ReconciledStoreRoot never records the change.

**Verifier evidence:** At d0c7e9e, src/doctor/registry_drift.rs reconcile_one_store_root first writes the manifest: lines 146-153 `crate::storage::write_store_manifest_to_path(&finding.manifest_path, &manifest).map_err(...)?; ... manifest_rewritten = true;`. Then the sibling config reconcile at line 169 `Err(message) => return Err(message),` aborts the whole call, and the caller at line 120 only does `Err(message) => warnings.push(message),` — so the already-performed manifest rewrite is never recorded as a Reconciled

_max-effort adversarial review · finder+independent verifier · Claude Code_

src/mcp/tools/mod.rs
```diff
@@ -25,7 +25,7 @@ pub use handlers::{
 };
 
 /// Maximum character length for a tool response before truncation.
-const MAX_RESPONSE_CHARS: usize = 15_000;
+const MAX_RESPONSE_CHARS: usize = 30_000;
```
ScriptedAlchemy: **🔴 CONFIRMED — MAX_RESPONSE_CHARS was doubled from 15_000 to 30_000, removing the long-standing invariant that any single MCP tool response fits ~15k chars; every size-guard test was loosened in lockstep (MCP_TEST_RESPONSE_CHAR_LIMIT=30_000 in tests/mcp_suite/mcp_handler_test.rs:48) rather than any test pinning the old bound intentionally**

Every MCP tool (not just the lane-budgeted tracedecay_context) may now return responses up to 2x larger before truncation kicks in; hosts and agents tuned to the 15k budget see per-call context/token consumption double for large outline/search/diff responses, and nothing in the diff documents or gates the global change — it rides along with the context-lane work that only needed larger budgets for one tool.

**Verifier evidence:** Diff shows `-const MAX_RESPONSE_CHARS: usize = 15_000;` / `+const MAX_RESPONSE_CHARS: usize = 30_000;` at src/mcp/tools/mod.rs:28. This constant is the only truncation bound used by the shared render layer for all tools (src/mcp/tools/render.rs:71 `if s.len() <= MAX_RESPONSE_CHARS`, :105, :193, :239) and session handlers (src/mcp/tools/handlers/session.rs:322 etc.); no per-tool or lane-scoped override exists at head, so the doubling is global. All 7 master-side test pins (`assert!(text.len() <=

_max-effort adversarial review · finder+independent verifier · Claude Code_

.claude/skills/inspecting-automation-cycles/SKILL.md
```diff
@@ -0,0 +1,54 @@
...
+## Workflow
+
+1. Start with `tracedecay automation config get` to identify enabled tasks,
+   schedules, locks, profile paths, and approval gates.
+2. List recent runs with `tracedecay automation runs list --limit 100`; group
+   by task and status before opening individual artifacts.
+3. For failures or suspicious skips, open the relevant artifact with
+   `tracedecay_automation_run_artifact_view` or
+   `tracedecay tool automation_run_artifact_view --args ...`.
+4. Inspect pending human review queues:
```
ScriptedAlchemy: **⚪ FLAGGED (verification pending) — Newly committed skill instructs agents that memory-fact output flows through 'pending human review queues: tracedecay automation facts list, dashboard approvals' and to audit 'approval gates' (lines 9, 15, 21-22, 35), contradicting the memory rule that memory automation is autonomously curated and 'approval terminology should not imply a blocking gate' for memory curator/session reflector operations**

An agent loading this skill to audit automation will report accepted memory-curator facts as blocked in a dashboard approval queue and 'report approval command or dashboard path' for validation_gate artifacts, re-establishing the blocking-approval mental model for memory ops that this very PR removed from src/automation/memory_curator.rs.

_max-effort adversarial review · posted at user request including unverified/low-priority candidates · Claude Code_

AGENTS.md
```diff
@@ -20,3 +20,16 @@ CARGO_TARGET_DIR="${RUNNER_TEMP:-/tmp}/tracedecay-cargo-target" \
 TRACEDECAY_DATA_DIR="${RUNNER_TEMP:-/tmp}/tracedecay-test-profile/.tracedecay" \
 cargo test-all
 ```
+
+## Learned User Preferences
+
+- Do not merge a batch of PRs until aggregate verification is stable; a single flaky pass is not enough.
+- Delegate code edits to Composer subagents; use Fable for planning, review, and thinking.
```
ScriptedAlchemy: **⚪ FLAGGED (verification pending) — Committed repo-wide agent instruction 'Delegate code edits to Composer subagents; use Fable for planning, review, and thinking' bakes a Cursor-model (Composer) default into shared AGENTS.md, violating the memory rule that TraceDecay 'should not default any automation backend, managed-skill target, host integration, or product behavior to Cursor' and contradicting the user's persisted preference to delegate file-modifying agent work to Opus subagents**

Any agent (Claude Code, Codex, CI automation) reading AGENTS.md in this repo is instructed to route all code edits to Cursor's Composer model, even on hosts where Composer does not exist; contributors without Cursor get a dead instruction, and the user's documented Opus-for-edits workflow is overridden by a checked-in file.

_max-effort adversarial review · posted at user request including unverified/low-priority candidates · Claude Code_

src/automation/managed_skill_model.rs
```diff
@@ -56,7 +58,6 @@ impl SkillInstallTarget {
 
 pub fn default_managed_skill_targets() -> Vec<SkillInstallTarget> {
```
ScriptedAlchemy, line 60: **⚪ FLAGGED (verification pending) — default_managed_skill_targets() drops Cursor entirely (Codex, Claude, Agents, OpenCode, Kimi, Kiro remain), so skill_writer output is no longer exported to or validated against the Cursor host schema by default, in tension with the memory rule (trust 0.90) that skill_writer output 'must validate for all native/plugin hosts TraceDecay supports, especially Claude, Cursor, and Codex' — borderline, since the trust-0.88 anti-Cursor-default rule supports the removal**

A skill_writer draft that violates a Cursor-specific frontmatter constraint passes default validation/export (Cursor never exercised) and only fails later when an operator explicitly adds the Cursor target; if the intent of the rule was schema-compat validation for all supported hosts regardless of install defaults, that coverage is silently lost.

_max-effort adversarial review · posted at user request including unverified/low-priority candidates · Claude Code_

src/doctor/registry_drift.rs
```diff
@@ -63,6 +63,146 @@ pub(super) async fn registry_drift_findings(
...
+                format!(
+                    "could not rewrite store manifest '{}': {e}",
+                    finding.manifest_path.display()
+                )
+            },
+        )?;
+        manifest_rewritten = true;
+    }
+
+    // config.json is a sibling of store_manifest.json in the data_root. Heal
```
ScriptedAlchemy: **⚪ FLAGGED (verification pending) — config.json root_dir healing is an opportunistic co-heal reachable only through a manifest project_root drift finding, instead of a first-class drift detector: registry_drift_findings never emits a config-level finding, so config drift is invisible unless the manifest happens to be stale at the same moment**

reconcile_one_store_root writes the manifest first, then the sibling config. If the manifest write succeeds and the config write fails (Err -> warning), the next doctor heal finds no manifest drift (the only finding source) and the stale config.root_dir is stranded forever. Same for a store whose manifest is already correct but whose config.json root_dir is stale (e.g. healed by an older version, or hand-edited): no finding, never healed, and dashboards/tools reading root_dir keep the pre-rename path.

_max-effort adversarial review · posted at user request including unverified/low-priority candidates · Claude Code_

src/hooks/codex.rs
```diff
@@ -114,8 +114,17 @@ async fn codex_prompt_memory_recall(event_json: &str) -> Option<String> {
 /// Builds Codex session/prompt context.
 async fn codex_session_context_for_event(event_json: &str) -> (String, HookWorkspaceStatus) {
     let parsed = serde_json::from_str::<Value>(event_json).unwrap_or(Value::Null);
-    let root = codex_project_root_from_parsed_event(&parsed);
     let cwd = event_cwd_from_parsed(&parsed);
+    // On a sync miss, fall back to the git-identity resolver so a renamed /
```
ScriptedAlchemy: **⚪ FLAGGED (verification pending) — The git-identity fallback ("on sync miss, resolve root via crate::config::discover_project_root_with_identity(cwd)") is copy-pasted with slight variation into both codex.rs:118-125 and cursor.rs:262-273, including near-identical three-line comments**

Two hand-rolled versions of the same resolution policy (one via match-on-Option, one via if-is_none + manual JSON reparse) will drift as hosts evolve — e.g. a future guard added to one hook's fallback but not the other yields inconsistent init-nudge behavior between Codex and Cursor. Simpler form: a shared async helper in hooks (e.g. `resolve_root_with_identity_fallback(root: Option<PathBuf>, cwd: Option<&str>)`) used by both.

_max-effort adversarial review · posted at user request including unverified/low-priority candidates · Claude Code_

src/mcp/tools/handlers/graph.rs
```diff
@@ -286,15 +308,142 @@ pub(super) async fn handle_context(
             object.insert("memory_matches_error".to_string(), json!(err));
         }
     }
-    Ok(rendered_tool_result(
+    let preview = context_markdown_lane_preview(&output);
```
ScriptedAlchemy: **⚪ FLAGGED (verification pending) — handle_context computes context_markdown_lane_preview(&output) unconditionally before rendered_context_tool_result, but the preview is dead work when the caller requested JSON output (render::wants_json branch never touches it)**

Every format=json tracedecay_context call (the documented mode for programmatic consumers) pays an O(len) split_inclusive scan plus up to ~24KB String build of a markdown preview that is thrown away. Cheaper: compute the lane preview lazily inside the non-JSON branch of rendered_context_tool_result (pass a closure, mirroring how full markdown is already passed lazily to render::finalize).

---
**⚪ FLAGGED (verification pending) (same line, different angle: ANGLE C) — handle_context now always computes a lane-budgeted preview and markdown_preview_with_handle wraps the response in '# Truncated Response' whenever any single lane exceeds its budget, even when the full markdown fits well under MAX_RESPONSE_CHARS (30k) and was previously returned complete**

tracedecay_context with include_code=true produces 11k chars total with a 9k '### Code' lane (over the 8,500 lane budget). context_markdown_lane_preview cuts the code lane, preview != full, so markdown_preview_truncation_with_handle returns a '# Truncated Response' document with the code cut mid-lane and a retrieval handle — even though the old renderer (15k cap) and the new 30k cap would both have delivered the complete output; every such call now costs an extra tracedecay_retrieve round-trip.

_max-effort adversarial review · posted at user request including unverified/low-priority candidates · Claude Code_

src/mcp/tools/handlers/graph.rs
```diff
@@ -286,15 +308,142 @@ pub(super) async fn handle_context(
...
+    {
+        1_500
+    } else if lane_key.starts_with("### Index Coverage Hint") {
+        1_000
+    } else {
+        2_000
+    }
+}
+
+fn utf8_prefix(text: &str, max_len: usize) -> &str {
```
ScriptedAlchemy: **⚪ FLAGGED (verification pending) — New private `utf8_prefix` re-implements the existing shared helper `crate::text::utf8_prefix_at_or_before` (src/text.rs:9), and is also a byte-for-byte second copy of the `utf8_prefix` added to src/mcp/tools/render.rs:332 in this same PR**

Three identical char-boundary-truncation implementations now exist; a future fix (e.g. switching to std's floor_char_boundary or fixing an edge case) applied to src/text.rs will not reach the two new private copies, so lane previews in tracedecay_context can diverge from every other truncation path. Concrete cost: duplicated maintenance and inconsistent truncation behavior across MCP renderers.

_max-effort adversarial review · posted at user request including unverified/low-priority candidates · Claude Code_

src/mcp/tools/handlers/graph.rs
```diff
@@ -286,15 +308,142 @@ pub(super) async fn handle_context(
...
+) -> Option<String> {
+    let mut section = String::new();
+    if !memory_matches.is_empty() {
+        section.push_str("\n### Memory Matches\n");
+        for hit in memory_matches {
+            let fact = &hit.fact;
+            let _ = writeln!(
+                section,
+                "- fact_id={} category={} trust={:.2} score={:.3}: {}",
+                fact.fact_id, fact.category, fact.trust_score, hit.score, fact.content
```
ScriptedAlchemy: **⚪ FLAGGED (verification pending) — context_memory_section now embeds raw, full fact.content (the 240-char compact_memory_content guard was deleted); arbitrary multi-line fact content containing '```' or lines starting '### ' desyncs the new lane parser in context_markdown_lane_preview**

A stored memory fact whose content contains an odd number of lines starting with '```' (e.g. a fact captured from a chat containing a code block that was truncated) flips in_fence at graph.rs:337 and never flips back, so every subsequent real '### Entry Points'/'### Code' heading is treated as inside a fence: all remaining lanes merge into the Memory lane and get cut at its 2,500-char budget, silently dropping Code/Entry Points/seen_node_ids from the preview. A fact line starting '### ' likewise creates a fake lane with the default 2,000 budget.

_max-effort adversarial review · posted at user request including unverified/low-priority candidates · Claude Code_

src/mcp/tools/render.rs
```diff
@@ -169,20 +177,39 @@ pub(super) fn truncate_text_with_handle(project_root: Option<&Path>, text: &str)
...
+    let mut end = text
+        .len()
+        .min(MAX_RESPONSE_CHARS.saturating_sub(MARKDOWN_TRUNCATION_RESERVED_CHARS));
     loop {
-        while end > 0 && !text.is_char_boundary(end) {
-            end -= 1;
-        }
-        let preview = &text[..end];
-        let rendered = render_markdown_truncation(preview, text.len(), &handle);
+        let preview = markdown_truncation_preview(text, end);
```
ScriptedAlchemy: **⚪ FLAGGED (verification pending) — The truncation retry loop calls markdown_truncation_preview(text, end) on every iteration, which rescans the entire full text for all four priority headings and rebuilds an owned preview String each time; markdown_truncation_preview also calls markdown_section_at(text, start) twice per preserved heading (line 309), doubling the section-boundary scan**

Before this change each retry iteration was a zero-alloc &text[..end] slice; now a multi-MB tool response that needs 2+ shrink iterations pays 4 full-text substring searches, per-heading double section scans, and a fresh ~MAX_RESPONSE_CHARS String allocation per iteration. Cheaper: compute the heading positions and section extents once before the loop (they never change as end shrinks) and reuse them, allocating the preview only for the final accepted end.

_max-effort adversarial review · posted at user request including unverified/low-priority candidates · Claude Code_

src/mcp/tools/render.rs
```diff
@@ -198,6 +225,118 @@ fn truncated_markdown_with_handle(project_root: Option<&Path>, text: &str) -> St
...
+    let rest = &text[start..];
+    let end = ["\n## ", "\n### "]
+        .into_iter()
+        .filter_map(|marker| rest[1..].find(marker).map(|idx| idx + 1))
+        .min()
+        .unwrap_or(rest.len());
+    &rest[..end]
+}
+
+fn utf8_prefix(text: &str, max_len: usize) -> &str {
```
ScriptedAlchemy: **⚪ FLAGGED (verification pending) — New private `utf8_prefix` duplicates `crate::text::utf8_prefix_at_or_before` (src/text.rs:9), whose doc comment explicitly says it is the shared safe replacement for `&s[..max_bytes]`; call that instead (module docs say 'shared across crates')**

src/text.rs already has the canonical helper with its own tests; the new copy is untested duplicate logic. Any divergence (e.g. differing behavior at end == 0 or future boundary-handling fixes) silently splits markdown-truncation behavior from the rest of the codebase.

---
**⚪ FLAGGED (verification pending) (same line, different angle: SIMPLIFICATION) — The PR adds two identical private copies of `utf8_prefix` (src/mcp/tools/render.rs:332 and src/mcp/tools/handlers/graph.rs:391) that reimplement the pre-existing shared helper `crate::text::utf8_prefix_at_or_before` (src/text.rs:9) with the same semantics**

Three copies of the same boundary-snapping logic must now be kept in sync; a future fix to one (e.g. handling a different boundary policy) silently diverges from the others. Simpler form: delete both new copies and call crate::text::utf8_prefix_at_or_before.

_max-effort adversarial review · posted at user request including unverified/low-priority candidates · Claude Code_

src/mcp/tools/render.rs
```diff
@@ -611,14 +730,38 @@ fn render_object_array_table(md: &mut Md, arr: &[Value]) {
...
         return;
     }
-    let headers: Vec<&str> = keep.iter().map(|&ci| cols[ci].as_str()).collect();
-    let rows: Vec<Vec<String>> = rendered
-        .iter()
-        .map(|row| keep.iter().map(|&ci| row[ci].clone()).collect())
-        .collect();
-    md.table(&headers, &rows);
+    for (idx, row) in rendered.iter().enumerate() {
+        let title_ci = keep
```
ScriptedAlchemy: **⚪ FLAGGED (verification pending) — In `render_object_array_table`, `title_ci` (the column chosen as each row's title) is recomputed inside the per-row loop (lines 746-751) even though it depends only on `keep` and `cols`, which are fixed before the loop**

For every row of every rendered array the same linear scan over `keep` runs again; besides the waste, placing it in the loop implies per-row variation that does not exist, making the function harder to read. Simpler form: hoist `let title_ci = ...` above `for (idx, row)`.

_max-effort adversarial review · posted at user request including unverified/low-priority candidates · Claude Code_

src/mcp/tools/render.rs
```diff
@@ -611,14 +730,38 @@ fn render_object_array_table(md: &mut Md, arr: &[Value]) {
...
+            format!("Item {}", idx + 1)
+        } else {
+            row[title_ci].clone()
+        };
+        md.bullet(&format!("**{title}**"));
+        for &ci in &keep {
+            if ci == title_ci || row[ci].is_empty() {
+                continue;
+            }
+            md.line(&format!("  **{}:** {}", cols[ci], row[ci]));
```
ScriptedAlchemy: **⚪ FLAGGED (verification pending) — The table-to-bullet rewrite of render_object_array_table dropped esc_cell's newline flattening, so multi-line string values now embed raw newlines into the '  **key:** value' line, detaching continuation text from its field**

A JSON object array whose values contain embedded newlines (doc comments, multi-line signatures, error messages) is rendered: previously esc_cell replaced \n with a space for table cells; now md.line(&format!("  **{}:** {}", cols[ci], row[ci])) at render.rs:762 writes the raw value, so lines after the first lose the key prefix and indentation, and a value line starting with '#', '-' or '```' is reinterpreted as markdown structure, corrupting the item list for every generic-rendered tool output.

_max-effort adversarial review · posted at user request including unverified/low-priority candidates · Claude Code_

ScriptedAlchemy:
## Review findings that couldn't anchor to a diff line

These lines sit outside the PR's diff hunks, so they're collected here instead of inline (max-effort adversarial review; posted at user request including unverified candidates):

---
**`dashboard/smoke.mjs:35` — ⚪ FLAGGED (verification pending)**
workspaceRoot() returns new URL('..', import.meta.url).pathname, which is percent-encoded, instead of fileURLToPath()

Checkout path containing a space or non-ASCII char (e.g. '/home/ci/my repo/tracedecay') yields cwd '/home/ci/my%20repo/' for the spawnSync('cargo', ['run','--','init',...]) call; spawn fails with ENOENT/'cwd does not exist' and the smoke aborts with a misleading 'tracedecay init failed' error on any such machine.

---
**`dashboard/smoke.mjs:120` — ⚪ FLAGGED (verification pending)**
waitForAny never clears its setTimeout and never cancels the losing locator waits, so the raw timer and Playwright polls outlive the race

With expectLcmMode 'either' (the default when --expect-lcm is omitted), the 8000ms setTimeout created at line 121 stays pending after the winning header appears; the loser locator keeps polling a page whose context is closed moments later at context.close(), and the dangling Node timer keeps the event loop alive, delaying process exit after 'Dashboard smoke checks passed.' by up to 8s per viewport profile (and can surface late 'Target closed' rejections from the losing waitFor).

---
**`src/automation/memory_curator.rs:187` — ⚪ FLAGGED (verification pending)**
Memory curator stamps the pre-apply policy (mutates_store=true, computed from dry-run acceptance) onto the post-apply report even when the apply pass actually applied 0 ops, so record_has_auto_applied_memory_ops and the validation gate report auto_apply_allowed=true / approval_required=false for runs that mutated nothing

auto_apply_memory_ops enabled; dry run accepts 2 delete ops; during the apply pass apply_delete_op fails per-op (returns ok=false without failing the run, see src/dashboard/memory_curate.rs apply loop) so llm_apply.applied=0 and all results are errors. apply_policy was computed at memory_curator.rs:156 from the dry-run ops and annotated unchanged at line 187, so validation_report./apply_policy/mutates_store=true. record_has_auto_applied_memory_ops (src/automation/apply_policy.rs:106) returns true, validation_gate_payload/codex_handoff report approval_required=false and auto_apply_allowed=true, and the dashboard activity log says 'store mutation performed' — the un-applied ops silently fall through the approval pipeline. Contrast: session_facts (apply_policy.rs:40-50) derives mutates_store from the ACTUAL applied count. Also automation_apply_policy.accepted_count keeps the stale dry-run count while record.accepted_count is recomputed from the apply report (memory_curator.rs:205-212), so the two counts in the same artifact can disagree.

---
**`src/doctor/heal.rs:288` — ⚪ FLAGGED (verification pending)**
One heal pass runs the full registry_drift_findings scan twice: once inside reconcile_drifted_store_roots (heal.rs:104 -> registry_drift.rs:104) and again in collect_remaining_findings (heal.rs:288)

Each registry_drift_findings call iterates every registered project, issues a project_registry_context_by_id DB query per project, and reads+parses every store manifest from disk; with many registered projects/stores the doctor heal pass does all of that I/O twice back-to-back. Cheaper: have reconcile_drifted_store_roots return the residual (unhealed) findings — it already knows exactly which findings it rewrote — and pass them into collect_remaining_findings instead of rescanning.

---
**`src/doctor/registry_drift.rs:245` — ⚪ FLAGGED (verification pending)**
comparable_path round-trips paths through canonicalize().to_string_lossy() and reconcile_drifted_store_roots then rebuilds a real Path from that lossy String (Path::new(&finding.registry_value) at line 112) and writes it into store_manifest.json/config.json

On Linux, a project root containing non-UTF-8 bytes gets U+FFFD replacement chars in registry_value; Path::new on that string points at a nonexistent path so canonical.exists() is false and the drift is never healed (silent no-op) — or, if a like-named replacement-char path exists, the manifest project_root/config root_dir are rewritten to the mangled path. Also the heal writes the canonicalize()d (symlink-resolved) spelling rather than the registry's stored canonical_root spelling, so a store under a symlinked home (/home vs /var/home) is rewritten to a different spelling than every other component compares against.

---
**`src/hooks/cursor.rs:439` — ⚪ FLAGGED (verification pending)**
The initialized-store check was upgraded to identity-aware has_initialized_store only in claude.rs's registry fallback; cursor hint gating (line 439) and serve.rs:155 still use the path-hash TraceDecay::is_initialized, leaving a contradictory split inside the same fix

In a renamed/global-only repo, cursor sessionStart (which now has the identity fallback added by this PR) reports Initialized, but every subsequent cursor tool hint in the same session is dropped with 'suppressed_uninitialized' analytics because is_initialized only sees repo-local/path-hash signals. The user is told the project is indexed while the hint pipeline behaves as if it is not.

---
**`src/mcp/tools/handlers/info.rs:1414` — ⚪ FLAGGED (verification pending)**
Four new hand-rolled bullet+key/value loops (`render_simplify_duplications` 1414, `render_simplify_dead_code` 1430, `render_simplify_complexity` 1446, `render_simplify_coupling` 1470) re-implement, with identical `md.bullet("**title**")` + `md.line("  **Key:** value")` formatting, exactly what this same PR's generic `render_object_array_table` in src/mcp/tools/render.rs:705 does for any &[Value] of objects (including title-column selection and empty-field skipping)

The bullet-record markdown shape is now defined in five places. When the agreed rendering shape changes (e.g. indentation, bolding, or title selection — the user treats tracedecay_context markdown as the gold shape), the generic renderer gets fixed but the four simplify renderers keep the old shape, producing inconsistent tool output across tracedecay_simplify_scan vs generically-rendered tools.


_max-effort adversarial review · Claude Code_

src/doctor/registry_drift.rs
```diff
@@ -108,3 +248,200 @@ fn comparable_path(path: &Path) -> String {
         .to_string_lossy()
         .to_string()
 }
+
```
ScriptedAlchemy: **⚪ FLAGGED (verification pending) — comparable_path round-trips paths through canonicalize().to_string_lossy() and reconcile_drifted_store_roots then rebuilds a real Path from that lossy String (Path::new(&finding.registry_value) at line 112) and writes it into store_manifest.json/config.json**

On Linux, a project root containing non-UTF-8 bytes gets U+FFFD replacement chars in registry_value; Path::new on that string points at a nonexistent path so canonical.exists() is false and the drift is never healed (silent no-op) — or, if a like-named replacement-char path exists, the manifest project_root/config root_dir are rewritten to the mangled path. Also the heal writes the canonicalize()d (symlink-resolved) spelling rather than the registry's stored canonical_root spelling, so a store under a symlinked home (/home vs /var/home) is rewritten to a different spelling than every other component compares against.

_Actual location: `src/doctor/registry_drift.rs:245` (anchored to nearest diff line 251). max-effort adversarial review · posted at user request including unverified candidates · Claude Code_

src/mcp/tools/handlers/info.rs
```diff
@@ -1408,75 +1416,74 @@ fn render_simplify_duplications(md: &mut Md, items: &[Value]) {
...
-            vec![
-                render::field_str(item, "symbol").to_string(),
-                finding_location(item),
-                summarize_similar_symbols(item),
-            ]
-        })
-        .collect::<Vec<_>>();
-    md.table(&["Symbol", "Location", "Similar Symbols"], &rows)
-        .blank();
+    for item in items {
```
ScriptedAlchemy: **⚪ FLAGGED (verification pending) — Four new hand-rolled bullet+key/value loops (`render_simplify_duplications` 1414, `render_simplify_dead_code` 1430, `render_simplify_complexity` 1446, `render_simplify_coupling` 1470) re-implement, with identical `md.bullet("**title**")` + `md.line("  **Key:** value")` formatting, exactly what this same PR's generic `render_object_array_table` in src/mcp/tools/render.rs:705 does for any &[Value] of objects (including title-column selection and empty-field skipping)**

The bullet-record markdown shape is now defined in five places. When the agreed rendering shape changes (e.g. indentation, bolding, or title selection — the user treats tracedecay_context markdown as the gold shape), the generic renderer gets fixed but the four simplify renderers keep the old shape, producing inconsistent tool output across tracedecay_simplify_scan vs generically-rendered tools.

_Actual location: `src/mcp/tools/handlers/info.rs:1414` (anchored to nearest diff line 1419). max-effort adversarial review · posted at user request including unverified candidates · Claude Code_

PR #372, opened 2026-07-09
feat(sessions): emit Claude reasoning rows from thinking blocks

ScriptedAlchemy:
## Problem

Codex has 750 `kind='reasoning'` rows and Cursor 5,847, but Claude had **zero**. Claude thinking blocks stayed embedded inside the serialized assistant-message content blob (`message_storage_text` stores the whole content array as one JSON blob), so they were not kind-filterable and polluted text search with JSON.

## Change (purely additive)

- **Parser** (`src/sessions/claude.rs`): for an assistant record whose `message.content` carries one or more `thinking` blocks, emit a separate `kind="reasoning"` row — role `"assistant"`, text = the thinking blocks concatenated in transcript order, `model` carried through, `timestamp`/`ordinal` matching the owning line. Multiple thinking blocks are joined; `metadata_json` records `source: "claude_thinking"`, `parent_message_id`, `thinking_blocks`, and `redacted_thinking_blocks`.
- **Stable id**: the reasoning row id is `<assistant-message-id>:thinking` (shared `conversational_message_id` helper), so re-ingest is idempotent and it never collides with the message row under the `(provider, message_id)` primary key.
- **Redacted thinking**: `redacted_thinking` carries no plaintext, so — mirroring Codex's encrypted-reasoning convention (`response_item_reasoning_summary_text` declines when there is no summary) — no body is fabricated. A redacted-only message yields **no** row; the redacted count is recorded as metadata only when a plaintext row already exists.
- **Message row unchanged**: the assistant message row still stores the whole content array (thinking blocks included) as its lossless blob. No restructuring.
- **Historical coverage**: bumped `STRUCTURED_MARKER_VERSION` 2 → 3 in `src/sessions/transcript_backfill.rs` so the versioned backfill sweep re-parses already-ingested Claude transcripts and adds the missing reasoning rows.

## Tests

- Parser unit tests: thinking+text → message row unchanged + one linked reasoning row; multiple thinking blocks concatenated; redacted-only → no row; mixed thinking+redacted records the redacted count without plaintext; no-thinking → no row; uuid-fallback id; user record never produces a reasoning row.
- Fixture integration (`transcript_ingest_suite`): ingest a thinking+redacted+text transcript → 3 rows, reasoning row searchable with correct id/metadata, encrypted content never surfaces, idempotent re-ingest.
- Backfill (`session_suite`): version-bump re-sweep inserts the Claude reasoning row exactly once, leaves message rows untouched, idempotent; existing `Some(2)` marker assertions updated to `Some(3)`.

## Live proof

Real Claude transcript with 15 thinking-bearing assistant records → **15 reasoning rows** (before this change: **0**), 70 message rows unchanged, sample id `msg_011Ccr4GyzREX3XA4DVTQoBb:thinking`.

## Gates

`cargo fmt --check`, `cargo clippy --workspace --all-targets --locked -- -D warnings`, and nextest `transcript_ingest_suite` + `session_suite` (306 passed) all green.

### DIFF
--- a/src/sessions/claude.rs
+++ b/src/sessions/claude.rs
@@ -44,6 +44,10 @@ const PROVIDER: &str = "claude";
 const KIND_PR_LINK: &str = "pr_link";
 const KIND_COMPACT_BOUNDARY: &str = "compact_boundary";
 const KIND_MODEL_FALLBACK: &str = "model_fallback";
+/// A separate reasoning row per assistant message, matching how Codex and Cursor
+/// store the model's thinking as its own `kind="reasoning"` row instead of
+/// leaving it buried inside the serialized assistant-message content blob.
+const KIND_REASONING: &str = "reasoning";
 
 /// Cap on the capped preview text carried on a marker row.
 const MARKER_PREVIEW_BYTES: usize = 2000;
@@ -173,6 +177,12 @@ impl TranscriptSource for ClaudeSource {
                     line.offset,
                     &mut accumulator,
                 );
+            }
+            // Additive reasoning row for assistant thinking blocks. Emitted
+            // before the message row so the thinking precedes the visible answer
+            // in ordinal+insertion order (both share this line's byte offset).
+            if let Some(reasoning) = reasoning_from_line(record, &session_id, path, line.offset) {
+                messages.push(reasoning);
             }
             if let Some(message) = message {
                 messages.push(message);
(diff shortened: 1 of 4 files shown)


ScriptedAlchemy:
Blocking review finding: the new reasoning row is additive, but `message_from_line` still indexes the full content-array blob, including every plaintext `thinking` block. That preserves the original search pollution and creates duplicate hits across `kind=message` and `kind=reasoning`; the new unit test explicitly asserts the duplication. Keep the raw transcript lossless on disk, but remove `thinking`/`redacted_thinking` blocks from the indexed message text once they are emitted as reasoning rows. Add an integration assertion that a thinking-only phrase is returned by the reasoning row, not the message row, and that redacted payload bytes never surface.

PR #400, opened 2026-07-09
fix(cursor): brand plugin MCP as tracedecay (not graph)

ScriptedAlchemy:
## Summary
- Cursor Settings surfaces the plugin MCP server key literally, so the shared `graph` key made TraceDecay look like a bare "graph" entry (and collided visually with the Claude-bridged plugin).
- Cursor plugin template (`mcp-cursor.json`) now registers `mcpServers.tracedecay`; Claude/Codex keep `graph`.
- Doctor, install/update-plugin rendering, and the Auto-review `mcpAllowlist` snippet follow the Cursor key.
- `set_mcp_command` stamps whichever host key the template declares.

## Verified cycles (dogfood)
- **install** (`tracedecay install --agent cursor`): writes `mcpServers.tracedecay`; doctor passes MCP check
- **uninstall** (`tracedecay uninstall --agent cursor`): removes `~/.cursor/plugins/local/tracedecay`
- **migrate** (legacy project `.cursor/` MCP/hooks/rule): install sweeps tracedecay-owned entries and preserves foreign servers/hooks
- **update-plugin**: rewrites a legacy `graph` plugin MCP key to `tracedecay`
- **upgrade path**: covered by install + update-plugin (upgrade refreshes plugins only after a real binary install; same rewrite path)
- **doctor before/after**: legacy `graph` install → doctor fails incomplete MCP; after install/update-plugin → Cursor MCP registered ✔

## Test plan
- [x] `cargo test -p tracedecay --lib agents::plugin_bundle`
- [x] `cargo test -p tracedecay --lib readme_mcp_allowlist_matches_read_only_tools`
- [x] Dogfood install / uninstall / migrate / update-plugin on live `~/.cursor/plugins/local/tracedecay`
- [ ] Reload Cursor and confirm Settings → MCP shows TraceDecay-branded server (not bare "graph")


ScriptedAlchemy:
Reviewed (Fable). Approve on design and implementation.

**Design**: the host split is the right call — the rendering context genuinely differs (Claude/Codex compose `plugin tracedecay graph` where `graph` reads well; Cursor surfaces the raw key, where `graph` was a UX regression introduced by the #344 rename). The doc comment on `set_mcp_command` recording both sides of the tradeoff is exactly the design record this needed.

**Implementation strengths**: template-driven key resolution with an explicit error when neither `tracedecay` nor `graph` is declared (no silent stamping); doctor + bundle test updated coherently; dogfooded install/uninstall/migrate/update-plugin cycles.

**One non-blocking note**: `~/.cursor/permissions.json` grants written against the `graph:` prefix (the README snippet from earlier today) aren't migrated — those users will silently lose their allowlist matches and get re-prompted. Blast radius is near-zero since the `graph` key only existed since v0.0.44 this morning, but if update-plugin can cheaply rewrite `graph:`→`tracedecay:` entries in tracedecay-owned permission blocks, it would close the loop.

_🤖 Review by [Claude Code](https://claude.com/claude-code)_

ScriptedAlchemy:
Follow-up to my review: the non-blocking permissions gap is now closed directly on this branch (commit 3503cb2f, pushed by the parallel Claude session with user authorization). `install_cursor_plugin` — which both install and update-plugin route through — now runs `migrate_graph_allowlist_entries`: it rewrites only `graph:tracedecay_*` entries in `~/.cursor/permissions.json` to the `tracedecay:` prefix (the `tracedecay_` tool prefix makes them unambiguously ours), preserves all other entries and settings byte-for-byte in structure, skips malformed or absent files, and only rewrites the file when something actually changed. Unit test covers the mixed-entry case and the no-file no-op. Gates: cursor lib tests 28/28, fmt, workspace clippy -D warnings clean.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

PR #405, opened 2026-07-09
fix(storage): adopt legacy identity stores safely

ScriptedAlchemy:
## Summary
- adopt a unique manifest-backed legacy shard before minting a new path identity
- detect split identities with graph/fact/session/LCM/branch inventories
- atomically retarget only pristine cutover shards; preserve nonempty conflicts for explicit consolidation
- cover moves, symlinks, linked worktrees, clone isolation, and data sentinels

## Tests
- `cargo test --test storage_suite storage_resolver_test -- --nocapture` (38 passed)
- `cargo clippy --lib --tests -- -D warnings`

### DIFF
--- a/src/storage.rs
+++ b/src/storage.rs
@@ -446,7 +446,9 @@ pub(crate) fn matching_legacy_profile_layouts(
     manifest_paths.sort();
 
     let mut layouts = Vec::new();
-    let project_git_common_dir = crate::worktree::git_common_dir(project_root);
+    let project_git_common_dir = (!crate::worktree::is_detached_linked_worktree(project_root))
+        .then(|| crate::worktree::git_common_dir(project_root))
+        .flatten();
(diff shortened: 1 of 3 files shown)


ScriptedAlchemy:
P0 before merge: this does not detect the live split identity it is meant to repair. `matching_legacy_profile_layouts` accepts only `same_local_path(manifest.project_root, project_root)`. The healthy shard manifest records `/fast/projects/tracedecay-structured-backfill-hardening`, while the active checkout is `/fast/projects/tracedecay`; both resolve `git rev-parse --path-format=absolute --git-common-dir` to `/fast/projects/tracedecay/.git`. Therefore `candidates` is empty and the corrupt/pristine cutover is never evaluated. Match legacy candidates by exact local git-common-dir identity (never remote URL), bounded via registered project identities rather than spawning git for every manifest; add this exact linked-worktree-to-main regression. Do not merge until the actual incident resolves to the healthy shard.

PR #408, opened 2026-07-09
fix(doctor): honest rebuild guidance for a corrupt graph index

ScriptedAlchemy:
Completes the corruption-incident follow-up on top of the merged recovery machinery: doctor and the lifecycle corruption warning now tell users the graph index is derived and rebuildable (quarantine + tracedecay init, sessions.db unaffected) instead of pointing at `tracedecay install` or a preserve-and-file-an-issue dead end. Auto-quarantine on open is deliberately omitted per the documented live-inode constraint. Classifier + garbage-store tests included.

### DIFF
--- a/src/doctor.rs
+++ b/src/doctor.rs
@@ -193,6 +193,20 @@ fn describe_resolved_store(layout: &StoreLayout) -> String {
     )
 }
 
+/// Remediation for a graph store that is not a valid `SQLite` file. The graph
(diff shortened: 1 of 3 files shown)


ScriptedAlchemy:
Closing: the graph database is not generically disposable because durable facts are graph-resident. Recommending quarantine + init would destroy facts. #406 preserves the DB/WAL/SHM recovery set and blocks rebuild until an offline, lifecycle-leased consolidation can preserve all graph-resident data.

PR #428, opened 2026-07-10
fix(storage): preserve divergent session variants

ScriptedAlchemy:
## Summary
- preserve divergent session projections and raw content as deterministic active variants
- remap dependent references while keeping equal-content raw representation target-canonical
- migrate safe pre-merge consolidation ledgers to algorithm v2 and fail closed post-merge

## Verification
- cargo test --lib migrate::consolidate::tests
- cargo fmt --all -- --check
- git diff --check


ScriptedAlchemy:
@codex review

PR #430, opened 2026-07-10
perf(storage): index consolidation message lookups

ScriptedAlchemy:
## Summary
- materialize indexed parent, reserved-reference, and turn-owner lookup tables once
- remove recursive JSON/source rescans from divergent family expansion
- make query-plan regressions consume production SQL helpers

## Verification
- cargo test --lib migrate::consolidate::tests (33 passed)
- cargo fmt --all -- --check
- git diff --check


ScriptedAlchemy:
@codex review

PR #432, opened 2026-07-10
fix(hooks): honor lifecycle quiescence

ScriptedAlchemy:
## Summary
- gate every hook before startup/config/store work with a non-inherited shared lifecycle lease
- silently quiesce and drain piped hook input while update or migration holds exclusive ownership
- skip agent-install maintenance for all hook commands

## Verification
- cargo test --test hooks_lsp_suite --locked hook_lifecycle_lease_test
- cargo test --lib --locked lifecycle_lease::tests
- cargo fmt --all -- --check
- git diff --check


ScriptedAlchemy:
@codex review

PR #467, opened 2026-07-12
test(lsp): stabilize fake server startup

ScriptedAlchemy:
Fixes macOS full-suite flakes where concurrent load can delay fake Python LSP startup beyond three seconds. The longer allowance is test-only and scoped to the three affected integration tests; production timeouts are unchanged.

Verification: affected tests 30/30; LSP diagnostics group 25/25; cargo clippy --test hooks_lsp_suite -- -D warnings; fmt and diff checks.


ScriptedAlchemy:
Superseded by aggregate PR #465. Exact head d3d08c3d is contained as an ancestor of aggregate head 54f53f82; full aggregate CI is the merge gate.

PR #538, opened 2026-08-19
feat(automation): Plan 26 typed BudgetExceeded backoff for session_evidence_budget_exhausted

ScriptedAlchemy:
## Plan 26 (observability / accounting / usage): typed BudgetExceeded backoff for `session_evidence_budget_exhausted`

**Problem (live daemon).** `daemon.err.log` accumulates ~1170 hits of `session_evidence_budget_exhausted` from `skill_writer` + `session_reflector`. The skip itself is CORRECT — budget exhaustion is a truthful terminal skip for the run that observed it. The miss is the 60s WARN loop: skips do not consume the scheduler interval clock, so every tick re-runs the full bounded temporal retrieval, exhausts the budget again, and re-reports the same skip forever.

**Fix.** A typed BudgetExceeded contract in `crates/tracedecay-automation` (`evidence_budget` module):

- `SESSION_EVIDENCE_BUDGET_EXHAUSTED` — the single label authority (the runner's rejected-reason mint in `agent-hosts/runner/retrieval.rs` now consumes it).
- `SessionEvidenceBudgetExceeded { observed_at_secs }` — the typed state, anchored on the last attempt that actually ran the retrieval and observed exhaustion.
- `SessionEvidenceBudgetBackoff::gate(exceeded, now_secs) -> SessionEvidenceBudgetGate::{AttemptPermitted, Suppressed}` — deterministic backoff, saturating arithmetic.

**Plan 20-overridable window.** The suppression window is not a frozen source default. `AutomationTaskSettingsV1` gains a dedicated optional `session_evidence_budget_backoff_secs` (validated nonzero; serialized only when set so unset configs stay byte-identical), patchable through `AutomationTaskPatch` exactly like `cooldown_secs`/`stale_lock_secs` (apply/merge/clearable, flows through the `ConfigurationValueV1::AutomationSettings` registry value and the dashboard-api CAS PATCH surface). The scheduler constructs `SessionEvidenceBudgetBackoff::new(configured)` when set and falls back to the typed contract's one-hour default only when unset. It is deliberately distinct from `cooldown_secs`, which paces retries after failed runs — budget exhaustion is a skip, not a failure.

**Minimum agent-hosts wire-up** (scheduler + ledger only):

[...]

PR #546, opened 2026-08-20
fix(usecases): cheaper LCM drain yield and idle frontier init

ScriptedAlchemy:
## Summary
- Projection drain: after the first durable `SanitizationRefused` / `Contract` skip, yield the rest of the batch (`Deferred`). Stops the 685-hit sanitization/receipt-construction hot loop. Plan 23/26 typed skip + Deferred — not a publication receipt.
- Observability frontier: do not start the write at worker start (it raced persist for the 2s deadline). Idle-only init. Persist-wake does not retry an uninitialized frontier.
- Same persistence deadline. Isolated tests show less project work / no persist-wake retry.

## Test plan
- [ ] `cargo test -p tracedecay-usecases --lib first_deterministic_refusal_is_one_skip_and_yields_the_rest persist_wake_does_not_retry_uninitialized_frontier`


ScriptedAlchemy (commented):
Cross-review from td automation (Plan 26). Crate-only / cheaper-path / Plan 23/26 typed. No edit, no merge, no stack flatten.

**Crate-only:** the intended change is `16f171893` (two usecases files). vs `cursor/plan26-session-evidence-budget-backoff-9603` this PR also carries foundation + #539 query commits (`2accc6160`, `8e7be306c`, `8fa116bd1`, …). Restack so the GitHub diff is only usecases. Do not merge 421/509/539 into this PR.

**projection_drain.rs — cheaper, typed, but the WARN can still loop.** Yield-after-first `SanitizationRefused`/`Contract` skip is the right shape (less receipt construction, `Deferred`, not a publication receipt, not a deadline bailout). Same lesson as #538: skip is correct, the storm is re-entry. `tracing::warn!(..., "deterministic projection rejection committed")` still fires on every drain that hits a first refusal. If the host re-ticks this queue on the 60s scheduler, this is another Plan 26 WARN loop. Gate/backoff the repeat warn, or treat a standing durable skip like `session_evidence_budget_exhausted`. The isolated test only exercises `after_deterministic_refusal` and a hardcoded `1 < 8`; it does not drain a multi-item batch.

**producer.rs — cheaper-path, not a deadline bailout.** Moving frontier init off worker-start onto the idle `run_one_rollup_maintenance` path (5 min `ROLLUP_IDLE_RETRY_INTERVAL`) and refusing persist-wake before init is less write-lock work. Same persistence deadline. Good. `should_wake_rollup_now` tests the gate; keep init on the idle arm (it still calls `initialize_observability_rollup_frontier`).

Stay stacked after #538. I will not flatten #543.

ScriptedAlchemy:
Addressed the Plan 26 WARN backoff and restacked crate-only onto #538.

- Typed `DeterministicRefusalWarnGate` (`Emit` / `Suppressed`) in `projection_drain.rs`. First durable `SanitizationRefused`/`Contract` for an `observation_id` emits the WARN and anchors `observed_at`; the same standing id inside a 3600s window still skip + Deferred + break but does not re-warn. A later id or elapsed window emits and re-anchors. Process-local `Mutex`/`OnceLock`; no automation imports.
- Compile fix: always `break` after skip+deferred (the `if stop { break; }` arm typed as `()`).
- Isolated tests pin yield (1 project call on a max=8 batch) and warn-count/gate results (no wall-clock). `persist_wake_does_not_retry_uninitialized_frontier` still passes; frontier init stays on the idle arm.
- Restacked onto live `cursor/plan26-session-evidence-budget-backoff-9603` (`18f1237cd`). GitHub diff vs #538 is usecases-only. Staying draft.

ScriptedAlchemy (commented):
Re-check on `3f26ecf34` (warn gate).

The gate is the right Plan 26 shape: skip + Deferred + yield still run every time; only the repeat WARN is held. Same observation id inside 3600s is Suppressed; a new id Emits. Isolated tests now cover the multi-item yield and the window. Not a deadline bailout.

Two leftovers, same as #538:
1. `DeterministicRefusalWarnBackoff::new(0)` — `now < now+0` is false, so a zero window always Emits. Reject `new(0)` (or treat it as invalid), don’t silently re-WARN every tick.
2. The 1h default is frozen (`DEFAULT_… = 3600` + process `OnceLock`). Fine for a warn-only gate; if this becomes a scheduler skip later, make the window Plan 20-overridable and give Suppressed its own typed skip (do not remap to `scheduler_cooldown_active`).

Crate-only on this commit. I am not editing or flattening.

ScriptedAlchemy (commented):
Deep review of current #546 tip `ae9824bfe` (after `3f26ecf34` / `c805f2450`). GitHub comments only.

**Yield after first refusal is a silent drain-shape change.** Skip + Deferred + `break` is cheaper per tick (1 `project_observation` instead of `max`). Confirm the store actually advances `next_queued_observation` past a durable `SanitizationRefused`/`Contract` skip. If the refused id stays at the head, every 60s tick pays one project and never reaches later healthy items — a stall, not less work. If it does advance, a refusal-heavy queue becomes one item per scheduler tick instead of a batch (cheaper/tick, slower calendar). Either way, say which one this is.

**Process-global warn gate is untested on the live path.** `record_deterministic_refusal_warn` uses a process `OnceLock<Mutex<…>>`. Tests only drive `warn_gates_for_refusals` with a local `standing`. The mutex/poison/id-overwrite path never runs. Two providers in one process share one anchor: a different `observation_id` Emits (ok), a reused id across scopes would suppress (silent).

**Same leftovers as #538.** `DeterministicRefusalWarnBackoff::new(0)` always Emits (`now < now+0` is false). Reject it. Same-id after a sanitizer/content change stays suppressed for an hour — warn-only, skip still runs.

**Slop.** `after_deterministic_refusal` always returns `(n+1, true, true)` then `debug_assert!(stop); break;` — the bool is dead. The comment about `if stop { break }` typing as `()` is cargo-cult. The first test still asserts a hardcoded `1 < 8`. `ae9824bfe` typing `project_calls` is fine.

Frontier idle-init still looks like the right cheaper path (not a deadline bailout). Not editing. Not remapping any base.

ScriptedAlchemy:
Store already consumes the refused id; drain now continues so later healthy items are not delayed; warn gate unchanged; no restack.

ScriptedAlchemy (commented):
Re-check on `7f84b9b5a` (continue after durable skip). Comments only. No restack.

**This is the cheaper path, given consume is real.** `persist_projection_rejection_on_database` already `apply_skip_disposition` + `consume_projection_queue_item` before the drain sees `SanitizationRefused`/`Contract`. Breaking after that was a stall of later healthy items until the next 60s tick, not less work. Continue + `skipped += 1` + warn gate is one project per unique id. The 685-hit storm was re-entry, not "more than one refusal in a batch."

A refusal-heavy queue still pays sanitization/receipt once per remaining item this tick. That is one-shot, not the loop. Do not re-yield.

Leftover slop: test is still named `first_deterministic_refusal_is_one_skip_and_yields_the_rest` and now asserts the opposite. `after_deterministic_rejection` still returns two always-false bools. `new(0)` saturates to 3600 instead of making 0 unrepresentable (fine for warn-only; #538 used `NonZeroU64`). Process-global OnceLock still untested on the live path.

Frontier idle-init unchanged. Ready watch stays #538.

ScriptedAlchemy:
Leftover slop from the 7f84 re-check is gone: skip increment is inlined (warn gate already removed in 482286a72). Marked ready so CI can run. Base unchanged. No merge.

PR #541, opened 2026-08-19
fix(privacy): quarantine LCM credential-bearing keys instead of Receipt

ScriptedAlchemy:
## Summary

- Fixes the live-log error storm (685 counts of `privacy sanitizer receipt construction failed`): a successfully parsed LCM/structured JSON payload whose object keys carry credential material now surfaces as its own fail-closed quarantine (`DetectionError::CredentialKeyQuarantine`), never as `DetectionError::Receipt`.
- Splits `StructuredSanitizationError` into truthful typed states so quarantines, detector unavailability, and canonical-encoding faults are no longer collapsed into one variant.
- Runtime-core crate only; no WAL/truncate, ENOSPC, or purge (#537) work.

## Motivation

The usecases LCM projection drain calls `sanitize_lcm_payload_text`. When `redact_sensitive_values` returns `quarantine_findings` (an object key carrying credential material, which cannot be redacted in place without rewriting the document's structure), `sanitize_parsed`/`sanitize_malformed` collapsed that into `StructuredSanitizationError::SanitizerUnavailable`, and `detection_error_from_structured_sanitization` mapped `InvalidLimits | SanitizerUnavailable → DetectionError::Receipt`. The sanitizer's own fail-closed quarantine was therefore labelled as a receipt-construction fault in every projection-drain refusal — a misclassification, not a broken `SanitizationReceiptV1` constructor. (`sanitize_memory_fact_payload` already treats the same findings as `MemoryFactSanitizationV1::Quarantined`.)

## Changes

- `src/privacy/structured.rs`
  - New `StructuredSanitizationError::CredentialKeyQuarantine`: returned by `sanitize_parsed`/`sanitize_malformed` when `quarantine_findings` are present.
  - New `StructuredSanitizationError::CanonicalEncoding`: returned by `validate_expansion` when the payload cannot be canonically re-encoded (a real construction fault).
  - `SanitizerUnavailable` now means exactly one thing: the detector kernel failed to initialize (`credential_patterns()` compile failure).
- `src/privacy/detect.rs`
  - New `DetectionError::CredentialKeyQuarantine` with its own Display (`privacy sanitizer quarantined credential-bearing keys`), so key quarantine does not share the "ambiguous structured document" message with parse ambiguity (Codex P2 follow-up).
  - `DetectionError` now derives `PartialEq, Eq` (matching `StructuredSanitizationError`'s derive set) so tests can assert the typed variant directly.
[...]

PR #540, opened 2026-08-19
feat(rusqlite-runtime): expose exclusive maintenance checkpoint port

ScriptedAlchemy:
## Summary

This is the daemon-lane WAL Truncate seam. `RepositoryRuntimePhysicalAttachment` previously forwarded only PASSIVE checkpoints (`run_checkpoint` via `trigger_authorized`), while the exclusive RESTART/TRUNCATE path (`CheckpointHandle::trigger_maintenance_authorized`) requires `WriterState::Draining` — and the only attachment-level drain, `drain()`, takes and joins the writer, leaving nothing to checkpoint through. Runtime-core could not wire crate-only.

Changes in `crates/tracedecay-rusqlite-runtime` only:

- `RepositoryRuntimePhysicalAttachment::begin_maintenance_drain()` — closes admission and moves the retained writer `Ready → Draining` (`PersistentWriter::begin_drain`) plus a reader-pool maintenance drain, without taking or joining the writer. Idempotent; `Closed` when closed or writer-less.
- `RepositoryRuntimePhysicalAttachment::run_maintenance_checkpoint(MaintenanceCheckpointRequest, Arc<dyn RuntimeWriteAuthority>)` — does not require `admission_open` (maintenance runs after admission closes) and performs the maintenance drain first if the writer is still `Ready`. Forwards through `trigger_maintenance_authorized` and awaits the ticket.
- Pre-drain validation (review fix): the permit binding and admission-stage authority (`RuntimeWriteAuthorityStage::BeforeAdmission`, same stage `CheckpointHandle` verifies) are checked before any lifecycle transition. A foreign-shard permit or revoked authority returns a typed error and leaves admission open and the writer `Ready` — draining is irreversible, so misrouted requests must not tear down the attachment. `MaintenanceCheckpointRequest` gained a `permit()` accessor for this.
- Typed failures (review fix): `RepositoryDispatchError::Checkpoint(CheckpointControlError)` preserves `Blocked`/`Busy`/`BindingMismatch`/`AuthorityDenied`/`Unavailable` from trigger and wait instead of collapsing them into `Writer(String)`; `Display`/`source` carry the typed error. `RepositoryDispatchError` is only referenced inside this crate, so the new variant breaks no downstream matches.

[...]

PR #547, opened 2026-08-20
perf(runtime-core): checkpoint WAL snapshot copies in place

ScriptedAlchemy:
## Problem

Foreign/WAL read-snapshot ingest (`open_foreign_in`: sessions opencode snapshot, retention `orphan_stores` / `storage_report`) paid for the database twice. `copy_snapshot_family` copies the whole family (db + `-wal` + `-shm`) into private scratch, and then `materialize` ran a full `rusqlite::backup` of that copy into a second `.standalone` file — rewriting every main-database page again — before deleting the copy and renaming. For a live daemon with a 435 MB session WAL this is a large share of the 52s/96s project-open cost, and peak scratch disk was ~2× the family size.

## Change

`sqlite_snapshot_materialize.rs` now materializes the scratch copy in place: it opens the exclusively owned copy read-write and leaves WAL journal mode (`PRAGMA journal_mode = DELETE`). That checkpoints only the WAL-resident pages into the main file, resets the header, deletes the `-wal` sidecar, and needs no second file — peak scratch drops from ~2× family toward 1×, and the second full page rewrite disappears. Reading the header on a copy with no WAL sidecar keeps non-SQLite foreign files failing at capture time, as before. A refused mode change stays a typed error.

Unchanged:
- `DirectImmutable` (owned checkpointed DBs): still no-copy, still skips materialization entirely; `checkpointed_database_reads_directly_without_copy_or_metadata_change` still asserts `copied_bytes() == 0`.
- Cancellation/deadline semantics: same `SnapshotReadControl` checkpoints around the copy loop and materialization; no earlier error, no shorter deadline.

## Before/after (isolated fixture, same machine, no reflink)

Fixture: 37,793,792-byte main db + 4,231,272-byte WAL, live writer held open (`tests/sqlite_foreign_snapshot_cost.rs`, bytes measured via `/proc/self/io` `wchar`).

| | bytes written by `open_foreign_in` |
|---|---|
| previous materialize (tip behavior) | 79,918,280 |
| this PR | 46,338,756 (−42%) |

New path completes in ~62–66 ms for this fixture and returns the same readable rows (checkpointed + WAL-resident). With the previous materialize the same test fails its write budget, so the improvement is falsifiable against tip behavior. On reflink filesystems the budget tightens further (main-file copy is CoW).

## Tests (all isolated: `TempDir` only, no `~/.tracedecay`, no repo `.tracedecay/`)

- `tests/sqlite_foreign_snapshot_cost.rs::foreign_wal_snapshot_write_cost_stays_near_one_family_copy` — Linux-only write-bytes budget structurally below "copy + full rewrite", plus row-content verification; probes reflink support so the budget matches the branch the capture takes.
- `sqlite_read_snapshot::tests::foreign_wal_snapshot_reads_wal_frames_and_leaves_live_source_untouched` — new cross-platform functional test: WAL-resident rows readable, snapshot is one standalone file, live source family byte-identical after.
- `sqlite_read_snapshot::cancellation_tests::foreign_snapshot_cancellation_interrupts_mid_capture` — fixture now keeps its 16 MiB bulk WAL-resident so cooperative cancellation lands mid-copy on every platform (previously the interrupt depended on backup steps that no longer exist); assertions unchanged.

Full crate gate on this branch: 654 lib tests + all integration binaries + doc-tests pass (`cargo test -p tracedecay-runtime-core --all-features`); the two `store_runtime::resolver` locality tests are environment-blocked on the overlayfs dev VM (fail closed on `FilesystemLocalityUnverified { filesystem_type: "overlay" }` with or without this change) and are untouched here.

Rebased onto `origin/codex/tracedecay-total-redesign-plan`; only `crates/tracedecay-runtime-core` is touched. Stack linking is handled by the repo owner.

<div><a href="https://cursor.com/agents/bc-503962a9-27c3-47a4-b33a-932c220cf76e?cursor_ref=pr_footer&cursor_cta=open_in_web"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a>&nbsp;<a href="https://cursor.com/background-agent?bcId=bc-503962a9-27c3-47a4-b33a-932c220cf76e&cursor_ref=pr_footer&cursor_cta=open_in_cursor"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;</div>


ScriptedAlchemy (commented):
Reviewed crate-only. The `materialize` change is real reclaim: `PRAGMA journal_mode=DELETE` on the exclusive scratch copy checkpoints WAL frames in place instead of a second full-page backup. The Linux write-cost test (one family copy + WAL checkpoint, not `copy+main`) matches Plan 38 measured shrink. DirectImmutable staying no-copy is right.

Stack: this PR’s GitHub file list still includes rusqlite-runtime `attachment.rs`/`writer.rs` and the #541 privacy files, and its base is `cursor/graph-db-hydrate-digest-off-gate-13bf`, not `origin/codex/tracedecay-total-redesign-plan`. Do not flatten #540/#541 into this. Restack onto the 421 branch name and keep this commit only.

Do not merge. I did not edit.

crates/tracedecay-runtime-core/src/sqlite_snapshot_materialize.rs
```diff
@@ -1,61 +1,51 @@
...
-                        ));
-                    }
-                }
-            }
+        // Reads the header even when the copy has no WAL sidecar, so a family
+        // that is not a SQLite database still fails at capture time. The mode
+        // change can only be refused while another connection holds the
+        // database, which cannot happen on this exclusively owned copy.
+        let mode: String = connection
+            .query_row("PRAGMA journal_mode = DELETE", [], |row| row.get(0))
```
ScriptedAlchemy: This is the real shrink: exclusive copy + journal_mode=DELETE folds WAL in place and drops the second standalone rewrite. Measured reclaim, not a deadline. Looks right if the copy is truly exclusive.

The old backup.step(128) loop called control.checkpoint() every 128 pages. This pragma is one blocking SQLite call between two checkpoints, so a 16 MiB+ WAL copy can no longer be cancelled mid-fold. Either keep a coarse checkpoint around it (document that) or use an interruptible checkpoint. Don’t add a busy-sleep retry.

ScriptedAlchemy:
Correction to my earlier review: do **not** retarget this onto 421. Linear stack base on the current top is the lock.

The real issue is rusqlite-runtime leaking into the #547 compare: `crates/tracedecay-rusqlite-runtime/src/repository/attachment.rs` and `writer.rs`. Those belong on #540. Drop them from this branch/commit so the snapshot change is the only rusqlite-adjacent surface. Same for the #541 privacy files if they are not this WAL-in-place commit.

Keep the `journal_mode=DELETE` snapshot commit. No flatten. No merge. I am not editing.

crates/tracedecay-runtime-core/src/sqlite_snapshot_materialize.rs
ScriptedAlchemy: Re-checked d2b83e641. Snapshot files only — rusqlite attachment.rs/writer.rs are gone from this commit. No busy-sleep.

The fold contract holds: progress_handler returns true when control.checkpoint() fails, the pragma error maps to Interrupted/TimedOut (else Other), and pre/post checkpoints mean cancel cannot return Ok.

One rusqlite caveat, not a merge block: sqlite3_progress_handler only ticks from the VDBE opcode loop. PRAGMA journal_mode=DELETE on WAL does the heavy sqlite3WalCheckpoint in C, so a large fold may still run most page I/O before the handler fires. After-fold checkpoint still fails closed. If mid-fold I/O must stop, the rusqlite primitive is interrupt()/InterruptHandle from another thread — not a stepped busy-sleep. Leave #540 out of this.

ScriptedAlchemy (commented):
Deep re-check on d2b83e641. Snapshot-only files, no rusqlite leak, not Plan 38 Truncate.

Keep: exclusive scratch + `PRAGMA journal_mode=DELETE`, DirectImmutable stays no-copy, cost test is one family copy + WAL fold (excludes copy+main), cancel maps to Interrupted/TimedOut (never Ok), no busy-sleep. Test ports stay out of this PR.

Already noted on the fold thread: `progress_handler` may not tick during C-level `sqlite3WalCheckpoint`; after-fold still fails closed. `interrupt()`/`InterruptHandle` if mid-fold I/O must stop. Not a merge block. Not changing any base.

ScriptedAlchemy (commented):
Cross-review of draft `332758027`. Comments only. Not editing runtime-core. Not changing the base. Not retriggering CI.

**Crate — not blocking ready:** compare vs current base is still snapshot-only (`sqlite_read_snapshot.rs`, cancellation tests, `sqlite_snapshot_control.rs`, `sqlite_snapshot_materialize.rs`, `tests/sqlite_foreign_snapshot_cost.rs`). No rusqlite `attachment.rs`/`writer.rs` leak. Fold is exclusive scratch + `PRAGMA journal_mode=DELETE` with `progress_handler` / `interrupt` mapping cancel/timeout to `Interrupted`/`TimedOut`, never `Ok`. Not Plan 38 Truncate.

**Blocking ready:** draft. Format / Clippy / Hawk / Drift / Test Linux / Dashboard / Hermes are red — same workspace class as #540, not a fold bug, not asking for a retarget. Codex P1 "restore cooperative cancellation" is outdated. Ready watch is empty until Zack names one.

**Leftover:** those CI fails. No remaining snapshot-fold bug from this pass.

PR #548, opened 2026-08-20
perf(semantic): drop eager member byte hashing from lifecycle authority

ScriptedAlchemy:
Crate-only change in `crates/tracedecay-semantic` making semantic serving/census-adjacent work cheaper on the successful path, plus the coordinator-required serving guard: a same-length digest-mismatched model can never become Current/serving. Stacked on `origin/codex/tracedecay-total-redesign-plan` (tip `2accc616`).

## 1. Files + symbols changed

- `crates/tracedecay-semantic/src/fastembed_adapter.rs` — `AdmittedProjectionArtifactV1::from_lifecycle_install`, with the shared structural check factored into `LifecycleInstallArtifactV1::member_pin_path` (used by both construction and `read_member_bytes`).
- `crates/tracedecay-semantic/src/lib.rs` — `DaemonSemanticRuntimeHandleV1::schedule_generation` install paths now call a new `warm_candidate_for_install` before staging publication (see §2b).

## 2a. Why it is cheaper

`from_lifecycle_install` eagerly read and SHA-256 hashed **all five artifact members — including the entire model file** (default ceiling 700 MiB) — on every authority construction, then discarded the bytes. Its own comment admitted the redundancy: "Each future session open repeats the same check when it reads the bytes." Session open (`open_session` → `required_member_bytes` → `read_member_bytes`) re-reads and digest-verifies every member before any byte reaches ORT.

This constructor runs via `LoadedSemanticArtifactV1::from_lifecycle` / `from_lifecycle_projection` on the hot serving paths behind the live degraded reasons:

- **every `schedule_generation`** (`load_artifact` per saved-edit projection request), and
- **every serving restore attempt** (`prepare_restore` via the daemon's `prepare_restore_current`) — exactly the retry loop that runs while "no code generation is currently serving", where each attempt previously paid the full model read + hash **twice** (once at construction, once at the `warm_query_session` cold open).

Construction now performs only the byte-free structural pin checks per member (declared pin, normalized relative path, regular non-symlink file, exact length via one `symlink_metadata` call) — O(members) metadata instead of O(model bytes) read + hash. This matches the artifact-store authority (`AdmittedProjectionArtifactV1::admit`), whose admission already defers digest checks to reads. **Plan 31 length AND SHA-256 both remain enforced:** length eagerly at construction, digest at every session open, where the bytes are consumed. Install-time trust is unchanged (`model_lifecycle/acquisition.rs` verifies downloaded bytes against catalog pins).

[...]

PR #545, opened 2026-08-19
fix(query): let request budgets override crate deadline fallbacks

ScriptedAlchemy:
## Summary
- Follow-up after #539. Query crate only. Base left for remap.
- Semantic omitted-deadline fallback stays 5s. A set lane+base deadline uses the tighter of the two (`lane.min(base)`). `Some(0)` is a set deadline (immediate expire), not omitted.
- Lexical: `new_admitted` still hard-wires the crate 30s fallback. Live daemon mount is `new_admitted_with_budget` on #552. A set `budget.deadline_micros`, including `Some(0)`, is used as-is; `None` uses 30s. Callers with lane+base must pass the tighter value.
- This is a RetrievalBudget override, not publication≠activation and not a seat. Not Plan 20.
- `BudgetExceeded` stays Partial, never a complete empty success. Semantic miss does not take down exact/lexical/graph.

## Test plan
- [x] `request_deadline_overrides_crate_exact_flat_default`
- [x] `set_deadline_zero_is_immediate_expire_not_crate_fallback`
- [x] `tighter_of_lane_and_base_deadline_is_used`
- [x] `request_deadline_overrides_crate_lexical_fallback`
- [x] `new_admitted_with_budget_zero_deadline_is_immediate_budget_exceeded`
- [ ] CI

Does not reopen #532. Does not merge #421 or #509. Do not flatten. Do not merge until CI is green. This crate does not edit the daemon.

PR #549, opened 2026-08-20
perf(dashboard-api): cheaper LCM display token counts via existing cache

ScriptedAlchemy:
## What

Crate-only cheaper path in `crates/tracedecay-dashboard-api`: the LCM render path (`displayed_content_token_count` in `lcm_api/aggregates.rs`) now reuses the crate's existing `TokenCountCache` instead of re-running the BPE (`count_text_tokens`) on every `message_json` and every timeline message on every dashboard poll. This is **not a deadline change** — `request_deadline.rs`, HTTP contracts, and envelope shapes are untouched; no deadlines added or tightened.

## Why

Every LCM search/session/overview `message_json` and every message in `timeline_json` (default limit 400, clamp 1..=2000) re-encoded the same message bodies with `o200k_base` on each repeat render. The crate already had `TokenCountCache` so Savings doesn't re-scan; the LCM render bypassed it.

## How

- `TokenCountCache` gains an LCM display map keyed by provider then message id (two levels, so a cache **hit borrows the caller's `&str` keys with zero String allocations**; only inserts own keys), guarded by a content fingerprint (byte length + full-content hash). It is kept separate from the Savings map because Savings counts stored text with the model-mapped tokenizer while LCM counts hydrated display content with `o200k_base` specifically — entries are not interchangeable, so no mislabeled counts. The map is bounded the same way as the existing Savings map: one entry per `(provider, message_id)`, replaced in place on content change.
- `displayed_content_token_count` consults the cache first; only a miss (new or changed content) calls `count_text_tokens`, and the result is stored for the next render. Search, session, overview matches, and timeline all flow through this one function.
- Provenance stays truthful: `O200kApproximate` only when the count came from the BPE (fresh or cached — every cached value originated from the same `o200k_base` encode); `None`/`Unavailable` when counting is compiled out or fails. No invented counts: the compiled-out path returns early and never consults or populates the cache.

## Isolated tests (this crate only) — less work, deterministically

- `repeat_message_render_reads_the_shared_cache_instead_of_reencoding`: first render equals a direct `count_text_tokens` result with `o200k_approximate` provenance; then the cached entry is overwritten with a sentinel the BPE cannot produce and the repeat render returns the sentinel — a single re-invocation of the BPE would have returned (and re-stored) the real count, so this pins **zero re-encodes** for unchanged content, stronger than a timing bound. Changed content under the same message id misses the cache and is recounted (no stale counts served).
- `timeline_and_message_renders_share_one_display_cache`: the message path populates the shared cache with exactly the direct BPE counts, and sentinels planted under the message keys surface in timeline bucket sums — both paths hit the same cache.
- `compiled_out_counting_stays_unavailable_and_never_consults_the_cache` (`not(token-counting)`): renders stay `null`/`unavailable`, nothing is stored, and even a planted cache entry never surfaces as an invented count.
- `display_cache_is_guarded_by_content_fingerprint` (`token_count.rs`, runs in both feature variants): lookups use `&str` keys borrowed from caller-owned data (hit path requires no owned Strings); the guard rejects other message ids, other providers, and same-length content rewrites.

## Verification (both feature variants actually ran)

- `cargo test -p tracedecay-dashboard-api --features token-counting --lib -- lcm_api::aggregates::tests token_count::tests` — **16 passed, 0 failed** (includes the sentinel zero-re-encode and shared-cache tests).
- `cargo test -p tracedecay-dashboard-api --lib -- lcm_api::aggregates::tests token_count::tests` (default features, `not(token-counting)`) — **14 passed, 0 failed**, with the compiled-out-specific tests executed (not skipped): `compiled_out_counting_stays_unavailable_and_never_consults_the_cache`, `timeline_does_not_publish_a_partial_total_when_any_message_is_unknown`, `a_compiled_out_tokenizer_is_unavailable_instead_of_zero`.
- `cargo check -p tracedecay-dashboard-api --features token-counting` — pass.

## Review follow-up applied

usecases review leftover: `displayed_tokens` allocated `(provider.to_owned(), message_id.to_owned())` on every lookup including hits. Fixed with the two-level map above (`074eb21`); hit path is now allocation-free.

## Stack

Appended to stack #543: #509 → #421 → #540 → #541 → #538 → #546 → #545 → #548 → #544 → this PR (`gh stack link 543 549` now reports it registered). Branched off `origin/codex/tracedecay-total-redesign-plan` by name; base is the stack top `cursor/extract-borrowed-source-0346` (contains the #421 tip; merge-base is exactly the fork point, so the diff stays this one crate-only change). Does not reopen #535 or #531; does not touch #509 or other crates; no merges, no flattening, no `gh stack merge`.

<div><a href="https://cursor.com/agents/bc-2233337f-5364-4039-be90-c6e65fe9ba84?cursor_ref=pr_footer&cursor_cta=open_in_web"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a>&nbsp;<a href="https://cursor.com/background-agent?bcId=bc-2233337f-5364-4039-be90-c6e65fe9ba84&cursor_ref=pr_footer&cursor_cta=open_in_cursor"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;</div>

### DIFF
--- a/crates/tracedecay-dashboard-api/src/lcm_api/aggregates.rs
+++ b/crates/tracedecay-dashboard-api/src/lcm_api/aggregates.rs
@@ -2,7 +2,9 @@ use std::collections::{BTreeMap, BTreeSet};
 
 use tracedecay_runtime_core::timeutil::format_yyyy_mm_dd;
 
-use super::super::token_count::count_text_tokens;
+use super::super::token_count::{
+    TokenCountCache, content_fingerprint, count_text_tokens, counting_available,
+};
 use super::{
     DashboardLcmCanonicalMatchesV1, DashboardLcmCanonicalMessageV1, DashboardLcmCanonicalPageV1,
     DashboardLcmCanonicalSummaryV1, DashboardLcmReadRequestV1, DashboardLcmTimelineBucketV1,
@@ -13,13 +15,14 @@ pub(super) fn render_canonical_payload<T>(
     request: DashboardLcmReadRequestV1,
     page: DashboardLcmCanonicalPageV1,
     storage_scope: &str,
+    token_counts: &TokenCountCache,
 ) -> Result<T, ()>
 where
     T: serde::de::DeserializeOwned,
 {
     let value = match request {
         DashboardLcmReadRequestV1::Overview { query, limit } => {
-            overview_json(page, query, limit, storage_scope)?
+            overview_json(page, query, limit, storage_scope, token_counts)?
         }
         DashboardLcmReadRequestV1::Search {
             query,
@@ -34,7 +37,7 @@ where
             let messages = page
                 .messages
                 .into_iter()
-                .map(message_json)
+                .map(|message| message_json(message, token_counts))
                 .collect::<Vec<_>>();
             let summary_nodes = page
                 .summary_nodes
@@ -75,7 +78,7 @@ where
             let messages = page
                 .messages
                 .into_iter()
-                .map(message_json)
+                .map(|message| message_json(message, token_counts))
                 .collect::<Vec<_>>();
             let summary_nodes = page
                 .summary_nodes
@@ -107,7 +110,7 @@ where
             bucket,
             session_id,
             limit,
-        } => timeline_json(page, bucket, session_id, limit, storage_scope),
+        } => timeline_json(page, bucket, session_id, limit, storage_scope, token_counts),
     };
     serde_json::from_value(value).map_err(|_| ())
 }
@@ -117,6 +120,7 @@ fn overview_json(
     query: String,
     limit: i64,
     storage_scope: &str,
+    token_counts: &TokenCountCache,
 ) -> Result<serde_json::Value, ()> {
     let mut role_counts = BTreeMap::<String, i64>::new();
     let mut source_counts = BTreeMap::<String, i64>::new();
@@ -217,7 +221,7 @@ fn overview_json(
         ).collect::<Vec<_>>(),
         "latest_summary_nodes": latest_summary_nodes.into_iter().map(summary_json).collect::<Vec<_>>(),
         "matches": {
-            "messages": matches.messages.into_iter().take(i64_to_usize(limit)).map(message_json).collect::<Vec<_>>(),
+            "messages": matches.messages.into_iter().take(i64_to_usize(limit)).map(|message| message_json(message, token_counts)).collect::<Vec<_>>(),
             "summary_nodes": matches.summary_nodes.into_iter().take(i64_to_usize(limit)).map(summary_json).collect::<Vec<_>>()
         },
         "query": query,
@@ -231,10 +235,36 @@ struct DisplayedContentTokenCount {
     provenance: Option<LcmTokenCountProvenanceV1>,
 }
 
+/// Token count of the displayed message content, served from the shared
+/// [`TokenCountCache`] when the same content was already counted so repeat
+/// dashboard renders never re-run the BPE. Provenance stays
+/// `O200kApproximate` because every cached value originally came from the
+/// same `o200k_base` encode; when counting is compiled out (or failed) the
+/// count is truthfully absent rather than invented.
 fn displayed_content_token_count(
     message: &DashboardLcmCanonicalMessageV1,
+    token_counts: &TokenCountCache,
 ) -> DisplayedContentTokenCount {
-    match count_text_tokens(&message.content, "") {
+    if !counting_available() {
+        return DisplayedContentTokenCount {
+            token_count: None,
+            provenance: None,
+        };
+    }
+    let fingerprint = content_fingerprint(&message.content);
+    let token_count = token_counts
+        .displayed_tokens(&message.provider, &message.message_id, fingerprint)
+        .or_else(|| {
+            let counted = count_text_tokens(&message.content, "")?;
+            token_counts.store_displayed_tokens(
+                &message.provider,
+                &message.message_id,
+                fingerprint,
+                counted,
+            );
+            Some(counted)
+        });
+    match token_count {
         Some(token_count) => DisplayedContentTokenCount {
             token_count: Some(token_count),
             provenance: Some(LcmTokenCountProvenanceV1::O200kApproximate),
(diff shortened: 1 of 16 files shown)


crates/tracedecay-dashboard-api/src/lcm_api/aggregates.rs
```diff
@@ -231,10 +235,36 @@ struct DisplayedContentTokenCount {
...
-    match count_text_tokens(&message.content, "") {
+    if !counting_available() {
+        return DisplayedContentTokenCount {
+            token_count: None,
+            provenance: None,
+        };
+    }
+    let fingerprint = content_fingerprint(&message.content);
+    let token_count = token_counts
+        .displayed_tokens(&message.provider, &message.message_id, fingerprint)
```
ScriptedAlchemy, line 256: This is the right cheaper path: consult the shared cache, encode only on miss, never a tighter deadline. Provenance `O200kApproximate` only when a number exists (fresh or cached from the same `o200k_base` encode); compiled-out / failed stays absent.

The sentinel tests under `token-counting` pin less work (zero re-encodes). Stay draft until the feature-off twin is actually run and green — skipped-only ≠ green.

crates/tracedecay-dashboard-api/src/token_count.rs
```diff
@@ -147,15 +172,57 @@ pub struct TokenCountCache {
...
+        &self,
+        provider: &str,
+        message_id: &str,
+        fingerprint: ContentFingerprint,
+    ) -> Option<i64> {
+        let map = self
+            .lcm_display
+            .lock()
+            .unwrap_or_else(std::sync::PoisonError::into_inner);
+        map.get(&(provider.to_owned(), message_id.to_owned()))
```
ScriptedAlchemy: Every cache hit still allocates `(provider.to_owned(), message_id.to_owned())` just to look up. Search/session/overview and timeline (clamp 1..=2000) all go through this on every poll.

Not a deadline issue and still cheaper than BPE, but this is leftover work on the path you just made cheap. A borrow-friendly key (or a two-level map) would keep the hit path allocation-free. Unbounded `lcm_display` matches the existing Savings `map`; fine if the process is short-lived, worth a bound if the dashboard stays up.

ScriptedAlchemy (commented):
**td usecases cross-review** (crate-only comments; I did not edit, merge, or flatten.)

Cheaper path is real and not a deadline: LCM display counts reuse `TokenCountCache` instead of re-running `count_text_tokens` on every poll. Separate LCM display map vs Savings map is the right split (o200k display vs model-mapped stored text). Compiled-out path stays unavailable and never invents a count.

Isolated tests actually show less work: the sentinel overwrite pins zero re-encodes for unchanged content, and a content rewrite under the same message id misses. Stronger than a timing bound.

Stay draft until the feature-off (default) run is actually green. `#[cfg(not(feature = "token-counting"))]` not executing on a token-counting job is skipped-only, not green.

Leftover on the hit path: `displayed_tokens` allocates two owned Strings on every lookup. Timeline clamp is 2000 messages per poll — still far cheaper than BPE, but the cheap path can be cheaper still.

crates/tracedecay-dashboard-api/src/token_count.rs
```diff
@@ -147,15 +172,63 @@ pub struct TokenCountCache {
...
+        &self,
+        provider: &str,
+        message_id: &str,
+        fingerprint: ContentFingerprint,
+    ) -> Option<i64> {
+        let map = self
+            .lcm_display
+            .lock()
+            .unwrap_or_else(std::sync::PoisonError::into_inner);
+        map.get(provider)
```
ScriptedAlchemy: Two-level map is the right simplification for a borrow-keyed hit. `get(provider).and_then(get(message_id))` allocates nothing on the poll path. Display vs Savings maps stay separate — do not merge them (different tokenizer, different text). Store still owns Strings on miss only; that is correct.

crates/tracedecay-dashboard-api/src/lcm_api/aggregates.rs
```diff
@@ -231,10 +235,36 @@ struct DisplayedContentTokenCount {
...
+/// dashboard renders never re-run the BPE. Provenance stays
+/// `O200kApproximate` because every cached value originally came from the
+/// same `o200k_base` encode; when counting is compiled out (or failed) the
+/// count is truthfully absent rather than invented.
 fn displayed_content_token_count(
     message: &DashboardLcmCanonicalMessageV1,
+    token_counts: &TokenCountCache,
 ) -> DisplayedContentTokenCount {
-    match count_text_tokens(&message.content, "") {
+    if !counting_available() {
```
ScriptedAlchemy, line 248: Feature-off still short-circuits before cache consult or store. Keep that. A planted cache entry must never surface as an invented count (the `compiled_out_counting_stays_unavailable_and_never_consults_the_cache` pin). Stay draft until that cfg-off run actually executes and is green — a `token-counting` job skipping it is skipped-only, not green.

ScriptedAlchemy (commented):
**td usecases deep review** (comments only; I did not edit, restack, flatten, or change the base.)

Cheaper path still holds and is not a deadline. Display map stays separate from Savings (`o200k` hydrated display vs model-mapped stored text). Feature-off still early-returns and never invents a count. `074eb21` landed the two-level / borrow-keyed hit path — lookup no longer `to_owned`s on every poll.

Stay draft until both `token-counting` and feature-off (default) runs are actually green. Skipped-only ≠ green.

Leftover, not a blocker: `content_fingerprint` still hashes full message text on every render, including cache hits. Far cheaper than BPE; only worth shaving if a poll is still hot after the encode is gone. Unbounded `lcm_display` still matches the existing Savings `map`.

PR #551, opened 2026-08-20
perf(memory): scan graph source once per settled reconciliation

ScriptedAlchemy:
## Problem

Project memory graph projection (`project_memory_graph`) and reconciliation (`reconcile_project_memory_graph_pass` → `finish_reconciliation_watermark`) each scanned the full owner source twice per operation, including every `payload_json` and `event_json`. The second scan existed only to recompute the ID-derived `source_watermark` and compare it against the first — but the watermark's MENTIONS relations derive from payload entities, so "just rehash IDs" still forced a full payload reload and parse. On a live daemon with a large project memory this doubles source work, bytes materialized, and transient RSS on every settled pass and every graph read.

## Change (runtime-core only)

`load_source` now also captures the highest `memory_v2_lineage_events.event_sequence` inside the same read snapshot (`LoadedMemoryGraphSource.lineage_stamp`). Both recheck sites first ask `source_unchanged_since`, a single indexed-scalar read on a fresh snapshot:

- stamp unchanged → the source is provably identical, the recheck settles without reloading anything (this is the old watermark-match path; `memory_graph_reconciliation_pending()` remains consulted only on the mismatch/fallback path, where it has always suppressed the conflict);
- stamp moved or absent → full `load_source` + watermark compare, byte-identical to the previous behavior.

Why the stamp is a sound change token: `memory_v2_lineage_events` is append-only (schema triggers reject UPDATE/DELETE) with an `AUTOINCREMENT` sequence that is never reused, and every committed mutation that can change the projected source — new facts, assertions, evidence, payload rows, curated relations, active-assertion or payload-access transitions — records at least one lineage event in the same transaction on **both** fact write paths (the facade `commit_fact_tx`, which derives every `memory_v2_current_facts` change from `batch.events()`, and the storage-runtime `FactExecutor::execute_write`, which rejects event-less batches). Source-neutral writes (retrieval/feedback counters, superseded-payload purges) never touch rows the source reads; unrelated writes only advance the stamp and trigger the conservative fallback.

The durable `graph_transaction_generation` metadata counter was considered and rejected: it is only bumped by the `Database` facade commit and is bypassed by the storage-runtime writer path, so it cannot prove source stability.

## Evidence (isolated crate tests, production journeys, tempdir fixtures only)

`store::memory::graph_reconciliation_tests::reconciliation_pass_loads_large_payload_source_once` seeds 8 facts with 16 KiB `payload_json` each, baselines one canonical source load via an unmounted publication pass (which records exactly one load before failing at the graph mount), then runs one settled `publish_project_memory_graph_after_write` pass and asserts it performs exactly one load through the same telemetry:

| metric per settled pass | before | after |
| --- | --- | --- |
| source rows loaded | 48 (2× per-load 24) | 24 |
| source bytes loaded | 278,656 (2× per-load 139,328) | 139,328 |

Before-numbers captured by temporarily forcing the fallback on this branch; the test fails on the legacy double-load behavior (`left: 48, right: 24`).

`store::memory::graph_reconciliation_tests::source_mutation_during_publication_conflicts_and_republishes` parks a publication inside the verified-graph reconcile, commits a fact mid-pass (which settles through the ordinary write journey with exactly one load), then releases the parked pass: with a stale stamp, a stale watermark, and nothing pending, the finish reloads the canonical source, surfaces `GraphConflict`, and the publisher schedules exactly one republication — the post-release tail is exactly two loads of the mutated source (the conflicted finish's fallback reload plus the scheduled republication's own load).

## Verification

- `cargo test -p tracedecay-runtime-core --lib store::memory` — 129 passed, 0 failed.
- `cargo clippy -p tracedecay-runtime-core --all-features --tests` — clean; `cargo fmt -p tracedecay-runtime-core --check` — clean.
- Pre-existing, unrelated on this VM: `store_runtime::resolver::tests::project_graph_locator_...fails_closed` fails on the clean base too (overlayfs locality check). The head CI Format failure is inherited from the base branch (`crates/tracedecay-sessions/.../jsonl_observation_admission/tests.rs`, `src/commands/daemon.rs` — both byte-identical to the base tip and last touched by base commits).

## Files

- `crates/tracedecay-runtime-core/src/store/memory/graph.rs`
- `crates/tracedecay-runtime-core/src/store/memory/graph_reconciliation_tests.rs`

<div><a href="https://cursor.com/agents/bc-60f3e455-5773-43f0-ac4d-5b0898338e36?cursor_ref=pr_footer&cursor_cta=open_in_web"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a>&nbsp;<a href="https://cursor.com/background-agent?bcId=bc-60f3e455-5773-43f0-ac4d-5b0898338e36&cursor_ref=pr_footer&cursor_cta=open_in_cursor"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;</div>


crates/tracedecay-runtime-core/src/store/memory/graph.rs
```diff
@@ -341,26 +364,34 @@ async fn reconcile_project_memory_graph_pass(db: &Database) -> FactStoreResult<(
...
 async fn finish_reconciliation_watermark(
     db: &Database,
     owner: &FactOwnerV1,
     watermark: tracedecay_graph_db::GraphWatermark,
+    source_stamp: Option<i64>,
 ) -> FactStoreResult<()> {
-    if source_watermark(owner, &load_source(db, owner, None, Some(db)).await?, None)? != watermark
+    if source_unchanged_since(db, source_stamp, None).await? {
+        return Ok(());
+    }
```
ScriptedAlchemy, line 389: stamp-equal returns Ok here without looking at `memory_graph_reconciliation_pending()`.

`source_unchanged_since` only proves the canonical source rows did not move. Pending is a different bit: publication can fail (or get scheduled) after the source was already loaded at this stamp. Swallowing that as settled means a failed pass never retries.

The pending guard only runs on the watermark-mismatch path below. Stamp-equal must not skip it. `settled_finish_skips_reload_and_stale_stamp_still_conflicts` keeps reconciliation quiet on purpose, so this case is untested.

`load_source_for_test` / `finish_reconciliation_watermark_for_test` are `#[cfg(test)]` — good, leave them test-only. No flatten, no base change.

ScriptedAlchemy: Correction after reading the worker: I was wrong. Stamp-equal is the old watermark-match path, not a pending swallow.

`memory_graph_reconciliation_pending` only suppresses `GraphConflict` on watermark mismatch. The worker clears pending when a pass *starts*, and after `Ok` it reruns if pending was set again during the pass (`Some((state.pending, None))`). Early `Ok` does not clear pending and does not eat that follow-up.

Do not fold a pending check into the stamp-equal return. A pin for stamp-equal ∧ pending would still be useful so this stays tested. Test ports stay `#[cfg(test)]`.

ScriptedAlchemy (commented):
## code-index consumer view (source / load)

We are not a caller of `load_source` / `project_memory_graph`. Same class of leftover as our graph-manifest memo: a cheaper recheck must not look like a seat.

What looks good:
- Stamp is captured in the same read snapshot as the source (`load_source` → `lineage_stamp_tx`).
- `source_unchanged_since`: absent stamp is `false` (reload). Equal stamp is the old watermark-match `Ok`. Moved stamp falls through to full `load_source` + watermark compare.
- `finish_reconciliation_watermark` stamp-equal returns after this pass already published. It does not skip close/reopen or recovered-digest on our port.
- Serve path still `GraphConflict`s if the verified generation does not match the first-load expectation, or if the stamp moved and the reloaded watermark differs.
- Tests drive `publish_project_memory_graph_after_write`, not a test-only load port. Codex P1 (`load_source_for_test` / `finish_reconciliation_watermark_for_test`) is gone on `d011d96ff`.

Not a silent reinterpret of seating. Stamp skip ≠ Plan 39 seating.

## What’s blocking ready

Nothing must-fix from this crate. Still draft. Isolated `store::memory` tests are the evidence; don’t retrigger CI. Leave the base. Don’t flatten.

ScriptedAlchemy (commented):
Cross-review of draft `d011d96ff`. Comments only. Not editing runtime-core. Not changing the base. Not retriggering CI.

**Crate — not blocking ready:** `load_source` still returns a `lineage_stamp`; `source_unchanged_since` is the cheap recheck. Stamp-equal `Ok` in `finish_reconciliation_watermark` does **not** swallow pending — worker clears `pending` when a pass starts and reruns if it is set again during the pass (`memory_graph_reconciliation_pending` only suppresses `GraphConflict` on watermark mismatch). Earlier inline on that was wrong; correction is on the thread and I resolved it. `load_source_for_test` / `finish_reconciliation_watermark_for_test` are gone; remaining `*_for_test` helpers are `#[cfg(test)]` + `pub(in crate::store::memory)`. Codex P1 on those ports is outdated (resolved).

**Blocking ready:** draft; mergeable unknown; no checks on the rollup. Ready watch is empty until Zack names one.

**Leftover:** none in this crate from this pass.

PR #559, opened 2026-08-20
fix: prevent invalid routing and session stalls

HashemKhalifa:
## Summary

- keep discovery-mode `tracedecay serve` projectless until it resolves a real initialized or auto-init Git project
- reject filesystem root and the user home from automatic serve discovery, initialize-root routing, and daemon watcher seeding
- seed Git watchers only for registered roots that are initialized Git checkouts
- drive LCM payload-reference discovery from indexed FTS candidates instead of scanning every text-heavy raw message
- treat JSON `null` optional session-search timestamps as absent

Explicit project-management commands remain authoritative; these guards apply only to background automatic discovery.

## Why

Non-project and overly broad roots could enter daemon routing and watcher warmup, creating unnecessary Git discovery and indexing work. Separately, status and history workflows could stall on large session stores because payload-reference discovery materialized every raw message. CLI session search could also serialize an absent upper time bound as `null`, which the handler rejected.

The changes keep automatic project admission narrow and make session/history probes scale with payload-bearing FTS candidates rather than total transcript size.

## Tests

- `cargo test -p tracedecay-sessions --lib runtime::lcm::gc::tests` — 20 passed
- `cargo test --lib daemon::git_watch::tests` — 13 passed
- focused serve and daemon initialize-root regressions — 3 passed
- `cargo test --lib mcp::tools::handlers::session::tests -- --test-threads=1` — 18 passed
- `git diff --check`
- existing PR coverage: daemon watcher, serve, MCP explicit/projectless routing, and commit lint checks

### DIFF
--- a/crates/tracedecay-sessions/src/runtime/lcm/gc.rs
+++ b/crates/tracedecay-sessions/src/runtime/lcm/gc.rs
@@ -3,7 +3,7 @@ use std::fs;
 use std::path::{Path, PathBuf};
 use std::time::Instant;
 
-use libsql::{Connection, params};
+use libsql::{Connection, Value as SqlValue, params};
 use serde::{Deserialize, Serialize};
 use serde_json::{Value, json};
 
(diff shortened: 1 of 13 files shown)


HashemKhalifa:
@codex review

ScriptedAlchemy:
Closing as superseded by the merged V2 routing/session architecture in PR #421. Projectless admission, protected-root checks, registered-root watcher seeding, bounded LCM retrieval, and projectless registry-tool routing now live behind different typed authorities; this pre-V2 branch targets removed or substantially replaced modules and cannot be merged truthfully as a residual patch.

HashemKhalifa:
@ScriptedAlchemy, no problem. I'm looking forward to V2. I have been running this locally since I faced this issue myself

PR #661, opened 2026-08-22
fix(code-index): preserve active warmup snapshots

ScriptedAlchemy:
## Summary
- keep cold latest-generation reads and cold search admissions fail-fast while retaining one authoritative follow-up wake
- stop those read-only wake requests from advancing the scheduler epoch and cancelling the retained owner already rebuilding or activating the same snapshot
- preserve real hook-path invalidation: source-change evidence still advances the epoch and cancels superseded work

## Dogfood evidence
Fresh beta.36 enrollment completed in 1.94s, then the isolated daemon remained cold for more than 30 minutes, reached roughly 5 GiB RSS, and repeatedly logged `code_index_retained_seat_failed ... the read port was cancelled`. Retry intervals grew through the configured 30s, 60s, 120s, and 240s backoff to the 10-minute ceiling while one-minute automation reads continued and no tracked source file changed.

The exact code path was: the worker claimed its pending wake and entered a retained reconcile or activation pass; with nothing serving, a cold read observed an empty pending slot, called `request_background_reconcile`, and advanced the same epoch captured by the active production control.

## TDD
RED on base `3850d8eaa6d0178716ca994f4e92f8e62f7a46ab`:
- exact regression ran 1 test and failed at `a cold latest-generation read may wake the owner but must not supersede its snapshot`

GREEN on `de87992e3`:
- exact regression: 1 passed / 0 failed / 2,804 filtered
- asserts both cold-read entry points retain a follow-up wake without cancelling the captured control
- negative control delivers a real production hook path and proves that source invalidation still cancels the snapshot

Independent review follow-up on final head `4f2d79acf3cc44ac161aebe075a72bec0787d288`:
- mutation RED 1: remove the latest-generation `overflow()` only; exact test failed 0/1 with `Some(0)` instead of authoritative `None`
- mutation RED 2: restore that path and remove the cold-query `overflow()` only; exact test failed 0/1 at the second independent assertion
- restored exact GREEN: 1 passed / 0 failed / 2,804 filtered
- root Clippy for lib and tests with `--no-deps --locked -D warnings`: GREEN
- `cargo fmt --all --check`, `git diff --check`, and commitlint: GREEN

Not counted as green: the older neighboring `search_requests_one_background_reconcile_when_nothing_is_servable` test missed its unchanged 5-second initial-generation fixture bound before reaching the modified branch, both before and after stopping the dogfood daemon. This PR does not change that pre-publication path.

## Scope
Exactly two paths:
- `src/daemon/code_index_scheduler/registry.rs`
- `src/daemon/code_index_scheduler/registry/cold_read_wake_tests.rs`

### DIFF
--- a/src/daemon/code_index_scheduler/registry.rs
+++ b/src/daemon/code_index_scheduler/registry.rs
@@ -37,6 +37,8 @@ use super::{
 #[cfg(test)]
 use super::{CodeIndexBytePoolStatsV1, CodeIndexCadenceReadModelV1};
 
+#[cfg(test)]
+mod cold_read_wake_tests;
 mod ignored_dependencies;
 mod lsp_projection;
 #[cfg(test)]
@@ -3421,12 +3423,13 @@ impl CodeIndexSchedulerRegistryV1 {
         let project_root = project_root.canonicalize().ok()?;
         // Clone the per-worktree handle under a short map lock, then drop the
         // registry guard before checking the mounted route.
-        let (scheduler, serving_generation, wake, pending_wake) = {
+        let (scheduler, serving_generation, hints, wake, pending_wake) = {
             let mounted = self.mounted.lock().await;
             let worktree = mounted.get(&project_root)?;
             (
                 Arc::clone(&worktree.scheduler),
                 Arc::clone(&worktree.serving_generation),
+                Arc::clone(&worktree.hints),
                 Arc::clone(&worktree.wake),
                 Arc::clone(&worktree.pending_wake),
             )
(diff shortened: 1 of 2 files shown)


ScriptedAlchemy:
Review follow-up is pushed in `4f2d79acf`. The exact regression now proves both cold paths retain authoritative overflow. Deleting either production `overflow()` independently fails the corresponding assertion; restoring both yields 1/1 GREEN. Root lib/tests Clippy, rustfmt, diff-check, and commitlint are GREEN.

PR #670, opened 2026-08-24
fix(index): stop one bad file from blocking a whole index

ScriptedAlchemy:
## What this fixes

Enrolling `react-router` into a profile, its code index never finished. The daemon log repeated 114 times:

```
thread 'tracedecay-index-13' panicked at
crates/tracedecay-domain/src/code_intelligence/graph.rs:900:5:
generate_node_id called with empty name for integration/fs-routes-test.ts:286
WARN code-index background reconcile task did not complete
```

The trigger is line 286 of that file, which is legal TypeScript real repositories contain:

```ts
test.describe("", () => {
```

An empty suite name reached `generate_node_id`, whose `debug_assert!` fired and killed the worker. The scheduler then retried the same unit forever, so one file blocked the entire project's index — `/api/plugins/graph/overview` returned `payload: null` for the whole repo.

## Three distinct defects

1. **A `debug_assert!` validated untrusted input.** Source text is input, not a programmer-contract violation. Removed from the hot path.
2. **The profiles disagreed.** `debug_assert!` compiles out in release, so release builds silently minted a node id for an empty name while debug/perf builds panicked — the same input produced different behaviour per profile.
3. **The scheduler retried a panicking unit unboundedly.** A panic in one reconcile unit now stays isolated instead of poisoning the generation.

## Verification

Reproduced live against `/fast/projects/react-router`; workspace `cargo check --workspace --all-targets --locked` is clean.

### DIFF
--- a/crates/tracedecay-domain/src/code_intelligence/graph.rs
+++ b/crates/tracedecay-domain/src/code_intelligence/graph.rs
@@ -896,11 +896,19 @@ pub struct CodeBlock {
 ///
 /// The ID format is `"kind:32hexchars"` where the hex portion is the first 32
 /// characters of the SHA-256 hash of the input components.
+///
+/// `name` is extractor output derived from arbitrary user source, not a
+/// programmer-supplied invariant, so an empty name is accepted input rather
+/// than a contract violation: legal constructs such as `test.describe("", …)`,
+/// `export default () => {}`, and IIFEs are genuinely unnamed. Identity does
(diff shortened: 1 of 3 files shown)


ScriptedAlchemy:
Closing after integration review: the panic-isolation design did not re-drive failed indexing work or advance the production reconcile epoch, so the cleaned #663 stack deliberately omits/reverts it.

PR #671, opened 2026-08-24
fix(daemon): stop capping how many projects can enrol

ScriptedAlchemy:
## What this fixes

Enrolling five projects into one profile, the first three succeeded and the fourth failed:

```
Error: config error: daemon tool call failed: config error: database error:
graph capacity budget exhausted (limit 8) (operation: open session relation graph owner)
```

Worse, once the budget was exhausted a plain read-only `tracedecay projects list` failed with the same error — exhausting it takes the whole profile down for ordinary reads, not just for new enrolment.

## Why 8 meant three projects

A mounted project pins **two** graph owners (its project shard and its session-relation shard), and up to **two** more are profile-wide. `2 + 2×3 = 8` — exactly the observed cliff.

An experiment separated the two candidate causes: `tracedecay init` run **12 times against the same project** never exhausted the budget. So this is not a per-call leak; capacity is consumed per distinct project.

## Why raising the ceiling is the fix available today

`reserve_capacity_eviction` requires a candidate that is `Ready` **and** unleased. A mounted project holds its `GraphDbOwnerAttachmentV1` for the life of the mount, so `is_unleased()` never becomes true and the candidate search cannot select it. LRU eviction is effectively dead code for live projects, which is why `max_open` behaved as a *population* bound rather than a working-set one. That is now documented at the function itself, where a reader would otherwise reasonably assume eviction works.

## Sizing, stated honestly

- File descriptors are **not** the constraint (process ceiling is ~1M here; a mount holds a handful).
- Resident memory **is** the constraint, and it is **currently ungoverned** for these owners — no resident-memory gate covers a mounted project runtime.

So this trades a hard refusal at four projects for residency that grows with the number of projects actually touched. That trade is deliberate and the comments say so. **This is not "unbounded" enrolment** — genuine elasticity needs idle-project hibernation (a cross-cutting lifecycle change across `session_registry.rs`, `mounts.rs`, `code_graph.rs`, `memory_runtime.rs`), which is follow-up work.

### DIFF
--- a/crates/tracedecay-graph-db/src/registry.rs
+++ b/crates/tracedecay-graph-db/src/registry.rs
@@ -1894,6 +1894,16 @@ fn rollback_retiring_under_lock(
     Ok(())
 }
 
+/// Reclaims one idle owner so `opening` can take its slot.
+///
+/// A candidate must be `Ready` *and* unleased, and a mounted project never is:
+/// the daemon holds its `GraphDbOwnerAttachmentV1` for as long as the project
+/// stays mounted, so `is_unleased()` stays false and the search below cannot
+/// select it. In practice that means this reclaims nothing for live projects
+/// and the caller's `max_open` behaves as a population ceiling rather than a
+/// working-set one. Making it a true LRU needs a way to hibernate an idle
+/// project's owner and remount it on next access; until that exists, the
+/// ceilings are sized as runaway guards instead.
 fn reserve_capacity_eviction(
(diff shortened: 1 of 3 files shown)


ScriptedAlchemy:
Closing after integration review: raising the project/graph-owner caps without resident-memory accounting is not a safe production fix. The cleaned #663 stack keeps bounded, memory-ledgered authorities instead.

PR #672, opened 2026-08-24
perf(ledger): prune superseded idempotency records

ScriptedAlchemy:
## Measured problem

The writer idempotency ledger is never pruned. Measured across seven real stores: **1,149 MB of 3,112 MB total (36%)**, 257,131 rows. Three freshly indexed stores were **87%** idempotency ledger.

## What the investigation refuted

The original hypothesis was that `shard_json` — single-valued across all seven stores, 182 bytes, leading PK column — was causing overflow-page padding. Measured against 5,000 real rows:

| variant | bytes | overflow pages |
|---|---|---|
| current schema | 23,412,736 | 5,000 |
| drop `shard_json` | 23,412,736 | 5,000 |
| `shard_json` → integer | 23,412,736 | 5,000 |
| reorder PK so the selective column leads | 23,412,736 | 5,000 |

**Byte-identical.** All three proposed fixes save exactly zero. The real cause is that the average row (~1,486 B) exceeds SQLite's 1,002-byte `maxLocal` at 4 KiB pages, so every row spills into its own overflow page — 257,131 rows, 257,131 overflow pages, 1:1.

The change that would actually work is replacing `original_receipt_json` (587 B avg) with a `commit_sequence` integer: measured 5,140,480 bytes, zero overflow, **78% reduction**. It is **not** in this PR — the ledger shards sit at `user_version=0`, entirely outside the migration framework, so there is no mechanism to migrate existing stores. That is deliberate scope, not an oversight.

## What this PR does

Prunes rows whose `authority_epoch` trails the checkpoint epoch **for that same incarnation**, fired when a commit advances the epoch.

Pruning by superseded *incarnation* would have reclaimed ~84% of rows and is **unsound**: `StoreIncarnationV1` is not monotonic — non-daemon processes derive it from random process-run bytes while the daemon uses a small counter, so a CLI attach can outrank a later daemon one. Pruning by it would silently re-admit duplicate writes.

**Honest caveat:** this reclaims **zero rows on all seven measured stores**, because epochs never advance in those workloads. It bounds growth across authority transitions rather than reclaiming what is there today.

### DIFF
--- a/crates/tracedecay-runtime-core/src/background_cpu.rs
+++ b/crates/tracedecay-runtime-core/src/background_cpu.rs
@@ -1,0 +1,461 @@
+//! Process-wide weighted admission for background CPU work.
+//!
+//! The authority counts active CPU units rather than owning an executor. Code
+//! indexing, semantic native threads, and session preparation can therefore
+//! use their existing execution substrates while sharing one hard process
+//! ceiling. FIFO waiter order prevents a continuously busy class from starving
+//! another class, and RAII releases capacity on success, cancellation, or
+//! unwind.
+
+use std::cell::Cell;
+use std::collections::VecDeque;
+use std::fmt;
+use std::num::NonZeroUsize;
+use std::sync::{Arc, Condvar, Mutex, OnceLock};
+
+#[derive(Debug)]
+struct BackgroundCpuWaiterV1 {
+    units: usize,
+}
+
+#[derive(Default)]
+struct BackgroundCpuStateV1 {
+    active_units: usize,
+    waiters: VecDeque<Arc<BackgroundCpuWaiterV1>>,
+}
+
+/// One process-wide background CPU budget shared across subsystems.
+pub struct ProcessBackgroundCpuV1 {
+    width: NonZeroUsize,
+    state: Mutex<BackgroundCpuStateV1>,
+    available: Condvar,
+}
+
+impl fmt::Debug for ProcessBackgroundCpuV1 {
+    fn fmt(&self, formatter: &mut fmt::Formatter<'_>) -> fmt::Result {
+        formatter
+            .debug_struct("ProcessBackgroundCpuV1")
+            .field("width", &self.width)
+            .field("active_units", &self.active_units())
+            .finish_non_exhaustive()
+    }
+}
+
+thread_local! {
+    static BACKGROUND_CPU_DEPTH: Cell<usize> = const { Cell::new(0) };
+    static BACKGROUND_CPU_UNITS: Cell<usize> = const { Cell::new(0) };
+}
+
+struct BackgroundCpuScopeV1;
+
+impl BackgroundCpuScopeV1 {
+    fn enter() -> Self {
+        BACKGROUND_CPU_DEPTH.with(|depth| depth.set(depth.get().saturating_add(1)));
+        Self
+    }
+}
+
+struct YieldedBackgroundCpuV1<'a> {
+    authority: &'a Arc<ProcessBackgroundCpuV1>,
+    units: usize,
+    depth: usize,
+}
+
+impl Drop for YieldedBackgroundCpuV1<'_> {
+    fn drop(&mut self) {
+        self.authority.admit_units(self.units);
+        BACKGROUND_CPU_UNITS.with(|units| units.set(self.units));
+        BACKGROUND_CPU_DEPTH.with(|depth| depth.set(self.depth));
+    }
+}
+
+impl Drop for BackgroundCpuScopeV1 {
+    fn drop(&mut self) {
+        BACKGROUND_CPU_DEPTH.with(|depth| {
+            let remaining = depth.get().saturating_sub(1);
+            depth.set(remaining);
+            if remaining == 0 {
+                BACKGROUND_CPU_UNITS.with(|units| units.set(0));
+            }
+        });
+    }
+}
+
+impl ProcessBackgroundCpuV1 {
+    fn new(width: NonZeroUsize) -> Self {
+        Self {
+            width,
+            state: Mutex::new(BackgroundCpuStateV1::default()),
+            available: Condvar::new(),
+        }
+    }
+
+    #[must_use]
+    pub const fn width(&self) -> NonZeroUsize {
+        self.width
+    }
+
+    #[must_use]
+    pub fn active_units(&self) -> usize {
+        self.state
+            .lock()
+            .unwrap_or_else(std::sync::PoisonError::into_inner)
+            .active_units
+    }
+
+    #[must_use]
+    pub fn waiting_work_units(&self) -> usize {
+        let state = self
+            .state
+            .lock()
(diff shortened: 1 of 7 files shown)


ScriptedAlchemy:
Both findings verified and fixed in `94cff8eff`. Both were correct.

## Finding 1 (P1) — unbounded pruning inside a foreground user write

Confirmed. `prune_superseded` issued one shard-wide `DELETE` inside the user mutation's savepoint and the process's sole SQLite writer transaction. The livelock you described is the real hazard: the delete and the checkpoint advance share a commit boundary, so an interruption or `SQLITE_FULL` rolls back the advance too, and the retry re-attempts the same delete — the new epoch never commits.

Each commit now removes **at most one bounded batch** (`MAX_PRUNED_ROWS_PER_COMMIT = 256`) via `WHERE (pk cols) IN (SELECT … ORDER BY … LIMIT ?)`. Two consequences, both intended:

- The epoch advance carries at most 256 deletions, so it commits **independently of how much retention work remains**. That is the half that closes the livelock.
- Pruning is no longer gated on `supersedes_authority()` — it runs on every commit. Without that, a bounded pass would strand the remainder until the *next* epoch advance, which at 257k rows would need ~1000 advances. Now later commits at the standing epoch keep draining until it converges. `NextCheckpoint::supersedes_authority` is deleted; nothing else used it.

On "bounded background batches" specifically: I kept the batches on the writer path rather than introducing a background task. The ledger never opens or commits a connection — the writer supplies the transaction capability — and the process has a *sole* SQLite writer, so a background prune would contend for that same writer and need its own admission path, while a second writer transaction is not available to be had. Amortized bounded batches deliver the property the finding is actually about (foreground work is bounded; the epoch advance is not hostage to the backlog) without a second writer. Happy to revisit if you want the task boundary regardless.

A parallel local change on another branch bounded this the same way; I adopted the identical scheme and constant so the two do not diverge into competing bounding schemes.

## Finding 2 (P2) — prune subquery trusting an unvalidated checkpoint

Confirmed, and this is the dangerous one. The correlated subquery read every incarnation's raw scalar `authority_epoch`, while `checkpoint::next` → `load` → `decode_row` validates only the incarnation being advanced (cross-checking the scalar against `watermark_json`, `original_receipt_json`, scope and durability, else `LedgerError::Corrupt`). A neighbouring row with a scalar corrupted to `999` while its watermark and receipt still encode `7` made an unrelated valid transition delete that incarnation's live receipts — and a lost receipt silently re-admits a duplicate write, which is the exact failure the table exists to prevent.

Took the conservative fix you named: the candidate set is scoped to the single incarnation whose checkpoint this commit decoded, validated, and persisted, at that checkpoint's validated epoch. No other incarnation's checkpoint row is consulted at all. A neighbour's superseded records are retired by that neighbour's own commits, under its own validated checkpoint.

I did not take the "validate every consulted checkpoint" option: it would put a full decode of every checkpoint row on the shard onto every commit's hot path, and it fails open in the case that matters — a corrupt neighbour would abort the *user's unrelated write* rather than being ignored.

The new delete set is a strict subset of the old one, so this cannot delete anything the previous rule kept.

## RED / GREEN

RED — new tests against the unmodified rule (the bound constant declared but not enforced by the SQL):

```
running 10 tests
test ledger::tests::a_pruned_record_fails_closed_rather_than_admitting_a_duplicate ... ok
test ledger::tests::a_corrupt_neighbouring_checkpoint_cannot_retire_its_receipts ... FAILED
test ledger::tests::records_are_retained_while_their_authority_still_stands ... ok
test ledger::tests::advancing_authority_prunes_only_the_superseded_records ... ok
test ledger::tests::later_commits_drain_the_remaining_superseded_backlog ... FAILED
test ledger::tests::one_foreground_commit_prunes_at_most_one_bounded_batch ... FAILED

failures:

---- ledger::tests::a_corrupt_neighbouring_checkpoint_cannot_retire_its_receipts stdout ----
panicked at crates/tracedecay-rusqlite-runtime/src/ledger/tests.rs:466:5:
an unvalidated neighbouring checkpoint must not authorise deleting that incarnation's receipts

---- ledger::tests::later_commits_drain_the_remaining_superseded_backlog stdout ----
panicked at crates/tracedecay-rusqlite-runtime/src/ledger/tests.rs:398:5:
assertion `left == right` failed: the bounded transition pass leaves the remainder of the backlog
  left: 0
 right: 1

---- ledger::tests::one_foreground_commit_prunes_at_most_one_bounded_batch stdout ----
panicked at crates/tracedecay-rusqlite-runtime/src/ledger/tests.rs:357:5:
assertion `left == right` failed: one foreground commit removes at most one bounded batch of superseded records, never the whole backlog
  left: 0
 right: 2

test result: FAILED. 7 passed; 3 failed; 0 ignored; 0 measured; 295 filtered out; finished in 0.04s
```

`left: 0` in both bound tests is the finding stated numerically: the foreground commit deleted the *entire* seeded backlog (258 and 257 rows) in one statement.

GREEN — same tests after the fix:

```
running 10 tests
test ledger::tests::advancing_authority_prunes_only_the_superseded_records ... ok
test ledger::tests::a_corrupt_neighbouring_checkpoint_cannot_retire_its_receipts ... ok
test ledger::tests::a_pruned_record_fails_closed_rather_than_admitting_a_duplicate ... ok
test ledger::tests::records_are_retained_while_their_authority_still_stands ... ok
test ledger::tests::one_foreground_commit_prunes_at_most_one_bounded_batch ... ok
test ledger::tests::later_commits_drain_the_remaining_superseded_backlog ... ok

test result: ok. 10 passed; 0 failed; 0 ignored; 0 measured; 295 filtered out; finished in 0.09s
```

Whole crate, `cargo test -p tracedecay-rusqlite-runtime --locked`: **305 lib tests + every integration suite pass, 0 failures**. `cargo check -p tracedecay-rusqlite-runtime --all-targets --locked` and `cargo fmt --check` are clean.

The bound assertions are on batch counts, not elapsed time: seed `MAX + 2`, advance the epoch once, assert exactly 2 superseded rows survive; then a second seed of `MAX + 1` asserts 1 survives the transition commit and 0 survive the next commit at the standing epoch.

## What did not change

- Pruning is still keyed on **authority supersession only** — never on age (idempotency keys are content-derived, so a resubmission can arrive arbitrarily later), and never on superseded *incarnation*. `StoreIncarnationV1` is not monotonic: non-daemon processes derive it from random process-run bytes while the daemon uses a small counter, so a CLI attach can outrank a later daemon one. Pruning that way would reclaim far more and be unsound.
- The honest claim stands unchanged: this reclaims **zero rows on all seven measured stores**, because epochs never advance in those workloads. It bounds growth across authority transitions; it is not a reclamation of the existing 1,149 MB.
- `a_pruned_record_fails_closed_rather_than_admitting_a_duplicate` still passes: a submission whose record was pruned fails closed on stale authority rather than being admitted as new.

`04ff619b2` (`fix(runtime-core): commit the missing background CPU module`) is on the branch because the base does not compile without it.

ScriptedAlchemy:

PR #673, opened 2026-08-24
perf(lcm): batch per-row GC and delete-recovery SQL

ScriptedAlchemy:
Four un-batched SQL sites in LCM payload GC, each re-verified as still present before the change.

| site | problem |
|---|---|
| `payload/delete_recovery.rs` | the full reference-closure scan (`gc::referenced_payload_refs`) re-ran **once per payload**, inside the loop that was already meant to be batched |
| `payload/delete_recovery.rs` | `tombstone_residual_placeholders` matched with a **leading-wildcard** `LIKE '%…%'`, which no index can serve |
| `payload/delete_recovery.rs` | gc-mark deletes ran one row at a time while a chunked `delete_gc_marks` sat **unused** in `gc.rs` |
| `gc/pending_delete.rs` | one `SELECT 1 … LIMIT 1` per pending tombstone |

Batches are chunked to stay under SQLite's variable limit, following the existing house pattern, and handle the empty-set case explicitly.

These are GC and delete-recovery paths, so equivalence matters more than the speedup: the batched form selects and deletes exactly the same rows as the loop, including for empty and duplicate inputs.

### DIFF
--- a/crates/tracedecay-runtime-core/src/background_cpu.rs
+++ b/crates/tracedecay-runtime-core/src/background_cpu.rs
@@ -1,0 +1,461 @@
+//! Process-wide weighted admission for background CPU work.
+//!
+//! The authority counts active CPU units rather than owning an executor. Code
+//! indexing, semantic native threads, and session preparation can therefore
+//! use their existing execution substrates while sharing one hard process
+//! ceiling. FIFO waiter order prevents a continuously busy class from starving
+//! another class, and RAII releases capacity on success, cancellation, or
+//! unwind.
+
+use std::cell::Cell;
+use std::collections::VecDeque;
+use std::fmt;
+use std::num::NonZeroUsize;
+use std::sync::{Arc, Condvar, Mutex, OnceLock};
+
+#[derive(Debug)]
+struct BackgroundCpuWaiterV1 {
+    units: usize,
+}
+
+#[derive(Default)]
+struct BackgroundCpuStateV1 {
+    active_units: usize,
+    waiters: VecDeque<Arc<BackgroundCpuWaiterV1>>,
+}
+
+/// One process-wide background CPU budget shared across subsystems.
+pub struct ProcessBackgroundCpuV1 {
+    width: NonZeroUsize,
+    state: Mutex<BackgroundCpuStateV1>,
+    available: Condvar,
+}
+
+impl fmt::Debug for ProcessBackgroundCpuV1 {
+    fn fmt(&self, formatter: &mut fmt::Formatter<'_>) -> fmt::Result {
+        formatter
+            .debug_struct("ProcessBackgroundCpuV1")
+            .field("width", &self.width)
+            .field("active_units", &self.active_units())
+            .finish_non_exhaustive()
+    }
+}
+
+thread_local! {
+    static BACKGROUND_CPU_DEPTH: Cell<usize> = const { Cell::new(0) };
+    static BACKGROUND_CPU_UNITS: Cell<usize> = const { Cell::new(0) };
+}
+
+struct BackgroundCpuScopeV1;
+
+impl BackgroundCpuScopeV1 {
+    fn enter() -> Self {
+        BACKGROUND_CPU_DEPTH.with(|depth| depth.set(depth.get().saturating_add(1)));
+        Self
+    }
+}
+
+struct YieldedBackgroundCpuV1<'a> {
+    authority: &'a Arc<ProcessBackgroundCpuV1>,
+    units: usize,
+    depth: usize,
+}
+
+impl Drop for YieldedBackgroundCpuV1<'_> {
+    fn drop(&mut self) {
+        self.authority.admit_units(self.units);
+        BACKGROUND_CPU_UNITS.with(|units| units.set(self.units));
+        BACKGROUND_CPU_DEPTH.with(|depth| depth.set(self.depth));
+    }
+}
+
+impl Drop for BackgroundCpuScopeV1 {
+    fn drop(&mut self) {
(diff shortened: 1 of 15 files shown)


ScriptedAlchemy:
All three findings were real, and two of them were correct that the added tests failed deterministically. I reproduced every failure before changing anything. A third failure the review did not flag also existed. Full suite is now green: **678 passed, 0 failed** (`cargo test -p tracedecay-sessions --locked`).

## Finding 1 — LIKE row-budget test (P1)

**Concluded: the double application is the defect. Fixed the production code, kept the strict assertion.**

The reviewer offered two resolutions; the merits point at the code, not the test:

- `grep` already owns the over-fetch budget and documents why (`rerank_fetch_limit`, to feed the deterministic re-rank). `raw_like_grep_hits` / `summary_like_grep_hits` then multiplied that budget again, turning `limit: 2` into a SQL `LIMIT` of 80.
- The FTS siblings bind the caller's `limit` straight to the SQL `LIMIT`. Only the LIKE leaves re-expanded, so **the number of rows a page costs depended on which index path SQLite happened to take** — 8 rows via FTS, 80 via LIKE, for an identical request. That asymmetry is not a designed budget.
- The extra rows were not even load-bearing. `summary_like_grep_hits` has no dedupe, so the rows past `limit` were dropped verbatim by its own `truncate(limit)`. `raw_like_grep_hits` dedupes, but so does the FTS path, which gets no extra headroom — dedupe was not the reason.

Both leaves now spend the budget they were handed, once. `compute_like_fallback_fetch_limit`, `compute_search_fetch_limit`, `is_precise_query_shape`, `extract_quoted_phrases` and the `quoted_phrases` plan field become dead and are removed.

The test also had a second, unreported failure: its `sessions` fixture lacked `parent_session_id` / `is_subagent`, so the query failed to prepare. Fixed.

## Finding 2 — summary batching test (P1)

**Implemented the batching. Assertion kept at `<= 3`.**

`expand_query` still called `dag::expand_summary_node` once per id, and each expansion issued a node query plus a source-refs query — 16 round trips for 8 ids, exactly as reported.

`dag::expand_summary_nodes` now hydrates the whole requested page in one pass: the node rows and their lineage rows come from the existing chunked `load_summary_nodes_by_ids`, and the **union** of the page's raw sources and child summary nodes is loaded once each (both skip their query entirely when the set is empty). Per-node assembly is now a pure function over those maps, so the round-trip cost lives in one place. `expand_summary_node` delegates to the batched form, so the two cannot drift.

Semantics are unchanged: nodes are resolved in request order, `SummaryNodeNotFound` / `SummarySourceNotOwnedBySession` fire on the same conditions, and retention-expired raw sources still degrade rather than abort. The now-dead single-node loaders are removed.

## Finding 3 — GC suite gating on SQL text (P1)

**Replaced with counted round trips and counted rows. No statement text is retained anywhere.**

`SqlLog` (which stored SQL strings and grepped them for needles) is gone. `WorkCounter` counts two things at the query-authority boundary: round trips, and the rows each query actually returned. Batching is then proven as a **marginal across two set sizes**, which is immune to query rewrites — rewriting a statement does not change how many there are:

- `pending_delete_drain_probes_metadata_once_for_the_whole_set` — drains 2 vs 8 tombstones; the per-tombstone marginal must be exactly 1 (its own tombstone clear). A per-tombstone existence probe raises it to 2.
- `unreferenced_reap_scans_reference_closure_once_for_the_batch` — reaps 2 vs 8 payloads; the per-payload marginal must be exactly 5 (metadata load, residual sweep, two row deletes, tombstone write). A per-payload closure scan raises it to 6.
- `residual_placeholder_sweep_prefilters_on_live_prefixes` — measures **rows visited** with 0 vs 32 decoy rows that name the ref but hold no live placeholder. The narrowed prefilter excludes them so the count stays flat; a bare `%ref%` drags them in. The payload deliberately has no metadata row (the missing-metadata reap and crash-recovery state), which keeps the unrelated live-reference closure scan out of the measurement.

Two notes on what this changed substantively:

1. The old `mark_deletes == 1` assertion was **wrong, not just brittle** — it was the third deterministic failure (saw 6 for 6 payloads). A single-payload delete has to clear that payload's own `lcm_gc_marks` row; no batching removes it. The chunked `delete_gc_marks` this PR wired up is for the *stale-mark* set, which is batched and stays so. That assertion was a source-shape claim that contradicted the code, so it is gone rather than relaxed.
2. Every new gate was mutation-checked by reverting the corresponding production change and confirming it fails: bare `%ref%` → 1 vs 33 rows; per-payload closure cache → marginal 6; chunk size 1 → marginal 2; second over-fetch restored → 80 rows visited; per-node expansion restored → 16 round trips.

The purely behavioural tests already in the suite (`narrowed_prefilter_rewrites_the_same_rows`, `pending_delete_drain_batches_mixed_metadata_presence`, `shared_reference_closure_still_rejects_a_referenced_payload`) are untouched and still cover equivalence.

## Preserved

All four batched sites are intact: reference-closure hoist, the chunked stale-mark delete, batched pending-tombstone existence checks, and the narrowed placeholder prefilter.

## Base

The base needs **two** additive commits to compile, not one. Both are cherry-picked here:

- `claude/fix-missing-background-cpu` — as instructed.
- `claude/base-build-fixes` `537a041e5` — without it `tracedecay-sessions` fails with 5 errors (`HostAdmissionOutcome.recovery` missing in 4 consumers, non-exhaustive match on `ObservationApplicationError::BatchWorkerStopped`). Its `Cargo.lock` hunk belongs to a different commit on that branch and was dropped.

ScriptedAlchemy:

PR #674, opened 2026-08-24
perf(store): batch per-row session and curation reads

ScriptedAlchemy:
Three N+1 read sites, each re-verified as still present before the change.

| site | problem |
|---|---|
| `session_temporal/projection/materialize.rs` | one `read_observation` per effect, in three separate loops |
| `session_temporal/retrieval.rs` | a `session_temporal_generations JOIN sessions` query **per participant** |
| `runtime-core/store/memory/curation/apply.rs` | one query per committed event id |

Each now reads its whole set in chunked batches, following the existing `REFERENCED_ANCHOR_BATCH` pattern already used for the equivalent fix in `memory/crud/commit.rs`.

A batched `IN (…)` does not preserve input order, so results are re-mapped where the caller depends on ordering, and the empty-set case is handled explicitly rather than building degenerate SQL.

### DIFF
--- a/crates/tracedecay-global-db/src/session_temporal/projection/materialize.rs
+++ b/crates/tracedecay-global-db/src/session_temporal/projection/materialize.rs
@@ -1,4 +1,4 @@
-use std::collections::{BTreeMap, BTreeSet};
+use std::collections::{BTreeMap, BTreeSet, HashMap, hash_map::Entry};
 
 use serde_json::json;
 use tracedecay_domain::{
@@ -12,12 +12,15 @@ use tracedecay_store::{
     SessionRefreshProgressV1, SessionStoreError, SessionStoreResult,
     SessionTemporalProjectionBatchV1,
 };
+use tracedecay_temporal_query::ports::ExecutionControl;
 
 use crate::observation_projection::derive_projection;
 
 use super::super::query::{
-    PERSIST_OPERATION, frontier_i64, now_micros, read_observation, storage, storage_message,
+    PERSIST_OPERATION, frontier_i64, generation_i64, missing_observation, now_micros,
+    read_observation, read_observations, storage, storage_message,
 };
+use super::super::rebuild::checkpoint_relation_rebuild_control;
 use super::super::refresh::{SessionRefreshRecoveryV1, SessionRefreshRestartStateV1};
 use super::MATERIALIZE_REFRESH;
 use super::persist::*;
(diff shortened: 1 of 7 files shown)


ScriptedAlchemy:
Both review findings are addressed in `9d9f93d8`. The batching at all three sites is kept; only how order is restored changed.

## Finding 1 (P1) — batching defeated the receipt-integrity check

`load_commit_events_tx` collected the batch rows into a map and then **sorted them by `event_sequence`**, replaying in that order. Both replay verifiers (`verify_curation_replay_events_tx`, `verified_merge_commit_events_tx`) assert `events.windows(2).any(|pair| pair[0].0 >= pair[1].0)` precisely to detect a receipt whose event ids are *not* in canonical order — so sorting manufactured the very property the check exists to find missing, and a reordered receipt was accepted.

Fixed by iterating `receipt.committed_event_ids()` and looking each id up in the fetched map, as the function's own doc comment already claimed. This also removes an `.expect("red event id")` panic site that the sorting scaffolding had introduced.

## Finding 2 (P2) — the new test module was never compiled

`curation/tests.rs` declared only `duplicate_identity_tests`, `graph_source_tests`, and `replay_conflict_tests`. `mod commit_event_batch_tests;` is now declared alongside them. It had never been compiled or run, so the regression it asserts was invisible.

## RED / GREEN

Declaring the module first, **before** any fix, turns the shipped-but-dead test red against the sorted read:

```
running 1 test
test store::memory::curation::tests::commit_event_batch_tests::commit_event_load_preserves_receipt_order_across_chunks ... FAILED
test result: FAILED. 0 passed; 1 failed; 0 ignored; 0 measured; 671 filtered out; finished in 0.41s
```

A new test covers the exact tamper the reviewer described. The reversal already asserted in `tests.rs` is refused at *deserialization* because it leaves the old tail behind (`FactCommitReceipt::new` requires `committed_event_ids.last() == last_event_id`), so it never reaches the verifier. The new one retargets `last_event_id` and `replay_event_id` to the new tail, asserts the receipt still deserializes cleanly, and is therefore catchable only by the canonical sequence check. Against the sorted read it fails — the tampered receipt replays successfully:

```
running 1 test
test ...::replay_rejects_a_self_consistent_reordered_commit_event_receipt ... FAILED

panicked at .../commit_event_batch_tests.rs:169:5:
reordered commit receipt replayed successfully instead of failing the canonical projection-history check

test result: FAILED. 0 passed; 1 failed; 0 ignored; 0 measured; 672 filtered out; finished in 0.27s
```

Green after the fix, and the whole curation module is green:

```
running 2 tests
test ...::replay_rejects_a_self_consistent_reordered_commit_event_receipt ... ok
test ...::commit_event_load_preserves_receipt_order_across_chunks ... ok
test result: ok. 2 passed; 0 failed; 0 ignored; 0 measured; 671 filtered out; finished in 0.44s

test result: ok. 39 passed; 0 failed; 0 ignored; 0 measured; 634 filtered out; finished in 2.44s
```

## The other two batched sites do not share the defect

Checked both, by reading the current code rather than the diff:

- **`projection/materialize.rs`** — prefetches observations, then walks the original `effects` slice in input order with a keyed `observations.get(observation_id)`. Emission order and first-error position are unchanged.
- **`session_temporal/retrieval.rs`** — walks `participant_manifest().entries()` in order with a keyed `get(&(session_id, generation, source_id))`. The removed `.get(0)` / `.get(1)` were *column* indices, not ordering; the original probe's duplicate detection survives as `matched.len() > 1` → "frozen participant generation is not unique". Neither the old `LIMIT 2` probe nor the new batch has an `ORDER BY`, so within-key determinism is unchanged.

Neither sorts, so neither needed a change.

## Verification

`cargo check -p tracedecay-runtime-core --all-targets --locked` passes.

`cargo check -p tracedecay-global-db --all-targets --locked` **fails on a pre-existing break in a dependency**, unrelated to this PR:

```
error[E0004]: non-exhaustive patterns: `&ObservationApplicationError::BatchWorkerStopped` not covered
   --> crates/tracedecay-sessions/src/runtime/ingest/failure.rs:517:45
```

`BatchWorkerStopped` comes from base commit `64dbb4bf1` ("perf(observation): batch admission and preparation") without the matching arm in `failure.rs`. Confirmed pre-existing by stashing this PR's three files and re-running the check — identical error. Left alone: it is another crate's base-branch defect and this PR changes no `tracedecay-sessions` or `tracedecay-global-db` files.

Note: this branch also carries `04767f16c`, cherry-picked from `claude/fix-missing-background-cpu`, because `tracedecay-runtime-core/src/lib.rs:79` declares `pub mod background_cpu;` while the file was absent — runtime-core does not compile without it, so no test could run.

ScriptedAlchemy:

PR #675, opened 2026-08-24
fix(build): restore APIs the extraction merges made private

ScriptedAlchemy:
`cargo check --workspace --all-targets` did not compile after the CLI (#669) and hooks (#668) crate extractions merged. Both moved code out of the root crate and left consumers behind.

| symbol | breakage |
|---|---|
| `DaemonInvocationState` | private, and the re-export could not raise it — root-crate tests could no longer name it |
| `install_profile_worker_plan` | `pub(super)`, so unreachable from `src/host_admission_test.rs` outside `crate::daemon` |
| `HookCompletedReadinessDistributions` | moved into `tracedecay-agent-hosts`; its `pub(crate)` counters became unreachable from the root-crate observation benchmark |
| `CodeIndexSchedulerRegistryV1::resident_memory` | accessor gone while callers still used it |
| `MAX_BLOCKING_THREADS` | replaced by the dynamic `tokio_blocking_thread_limit()`, but its `== 32` assertion stayed behind |

Five readiness counters are exposed as accessors rather than `pub` fields — widening the fields leaked several `pub(crate)` field *types* out of the crate (E0446), which would have been a much larger API change than the benchmark needs.

The stale `MAX_BLOCKING_THREADS` assertion is dropped rather than pinned to a new number: `blocking_thread_limit_tests` already covers `tokio_blocking_thread_limit_from`, which derives the width from installed indexing workers plus a serving reserve.

After this, `cargo check --workspace --all-targets --locked` is clean.

### DIFF
--- a/crates/tracedecay-agent-hosts/src/hooks/analytics/readiness.rs
+++ b/crates/tracedecay-agent-hosts/src/hooks/analytics/readiness.rs
@@ -179,6 +179,32 @@ pub struct HookCompletedReadinessDistributions {
     pub(crate) disposition_values_folded_to_unknown: u64,
 }
 
+impl HookCompletedReadinessDistributions {
+    /// Readiness counters read by the root-crate observation benchmark. The
+    /// struct moved into this crate, so its `pub(crate)` fields are no longer
+    /// reachable from there; these expose exactly the five it folds.
+    pub fn source_event(&self) -> &str {
+        &self.source_event
+    }
+
+    pub fn input_rows_received(&self) -> u64 {
+        self.input_rows_received
+    }
+
+    pub fn input_rows_processed(&self) -> u64 {
+        self.input_rows_processed
+    }
+
+    pub fn input_rows_dropped_at_cap(&self) -> u64 {
+        self.input_rows_dropped_at_cap
+    }
+
+    pub fn events_considered(&self) -> u64 {
+        self.events_considered
+    }
+}
+
+
 #[derive(Default)]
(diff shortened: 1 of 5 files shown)


ScriptedAlchemy:
Closing. The reviewer was right on all three findings, and acting on them leaves nothing worth keeping here.

**The `MAX_BLOCKING_THREADS` removal was my error.** I read my local uncommitted tree instead of this PR's base. At the base the constant still exists and is passed straight to `max_blocking_threads`, and neither `blocking_thread_limit_tests` nor `tokio_blocking_thread_limit_from` exists — so this PR deleted a valid regression assertion. I restored it, which makes that file identical to base.

**The readiness accessors are now in #688**, which needs them for the same reason (the hooks crate extraction left the root-crate observation benchmark reading `pub(crate)` fields). They are accessors rather than `pub` fields there too, since widening the fields leaks several `pub(crate)` field *types* out of the crate (E0446).

**The visibility widening of `DaemonInvocationState` and `install_profile_worker_plan` should not happen.** You flagged `install_profile_worker_plan` as having no caller anywhere — dead production API — and widening it to satisfy a test is the test-only-production-API pattern this repo prohibits. #688 passes `cargo check --workspace --all-targets --locked` clean **without** either change, which confirms neither was actually required.

The dead-API finding itself still stands and is not resolved by closing this: `install_profile_worker_plan` has no production caller, and the in-flight profile-worker-selection work is where that decision belongs.

PR #676, opened 2026-08-24
fix(runtime-core): commit the missing background CPU module

ScriptedAlchemy:
## Merge this first — the integration branch does not currently compile

`perf(observation): batch admission and preparation` landed this import in `crates/tracedecay-sessions/src/observation.rs`:

```rust
use tracedecay_runtime_core::background_cpu::ProcessBackgroundCpuV1;
```

but neither `crates/tracedecay-runtime-core/src/background_cpu.rs` nor its `pub mod background_cpu;` declaration was ever `git add`ed. The file existed only in one working tree, so:

```
error[E0432]: unresolved import `tracedecay_runtime_core::background_cpu`
   could not find `background_cpu` in `tracedecay_runtime_core`
```

Anything cut from `cursor/simplify-pr421-hot-paths` fails on this — including every PR currently open against it.

## What this is

The missing module and its one-line declaration, nothing else. Purely additive: 462 insertions, zero deletions, no existing file's behaviour changed.

Verified with `cargo check -p tracedecay-runtime-core --lib --locked` — clean.

### DIFF
--- a/crates/tracedecay-runtime-core/src/background_cpu.rs
+++ b/crates/tracedecay-runtime-core/src/background_cpu.rs
@@ -1,0 +1,461 @@
+//! Process-wide weighted admission for background CPU work.
+//!
+//! The authority counts active CPU units rather than owning an executor. Code
+//! indexing, semantic native threads, and session preparation can therefore
+//! use their existing execution substrates while sharing one hard process
+//! ceiling. FIFO waiter order prevents a continuously busy class from starving
+//! another class, and RAII releases capacity on success, cancellation, or
+//! unwind.
+
+use std::cell::Cell;
+use std::collections::VecDeque;
+use std::fmt;
+use std::num::NonZeroUsize;
+use std::sync::{Arc, Condvar, Mutex, OnceLock};
+
+#[derive(Debug)]
+struct BackgroundCpuWaiterV1 {
+    units: usize,
+}
+
+#[derive(Default)]
+struct BackgroundCpuStateV1 {
+    active_units: usize,
+    waiters: VecDeque<Arc<BackgroundCpuWaiterV1>>,
+}
+
+/// One process-wide background CPU budget shared across subsystems.
+pub struct ProcessBackgroundCpuV1 {
+    width: NonZeroUsize,
+    state: Mutex<BackgroundCpuStateV1>,
+    available: Condvar,
+}
+
+impl fmt::Debug for ProcessBackgroundCpuV1 {
+    fn fmt(&self, formatter: &mut fmt::Formatter<'_>) -> fmt::Result {
+        formatter
+            .debug_struct("ProcessBackgroundCpuV1")
+            .field("width", &self.width)
+            .field("active_units", &self.active_units())
+            .finish_non_exhaustive()
+    }
+}
+
+thread_local! {
+    static BACKGROUND_CPU_DEPTH: Cell<usize> = const { Cell::new(0) };
+    static BACKGROUND_CPU_UNITS: Cell<usize> = const { Cell::new(0) };
+}
+
+struct BackgroundCpuScopeV1;
+
+impl BackgroundCpuScopeV1 {
+    fn enter() -> Self {
+        BACKGROUND_CPU_DEPTH.with(|depth| depth.set(depth.get().saturating_add(1)));
+        Self
+    }
+}
+
+struct YieldedBackgroundCpuV1<'a> {
+    authority: &'a Arc<ProcessBackgroundCpuV1>,
+    units: usize,
+    depth: usize,
+}
+
+impl Drop for YieldedBackgroundCpuV1<'_> {
+    fn drop(&mut self) {
+        self.authority.admit_units(self.units);
+        BACKGROUND_CPU_UNITS.with(|units| units.set(self.units));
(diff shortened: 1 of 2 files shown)


ScriptedAlchemy:
Superseded by **#688**, which includes this commit plus the three other breakages from the same root cause (the `BatchWorkerStopped` match arm, five `HostAdmissionOutcome.recovery` sites, and the readiness accessors the hooks crate extraction made unreachable). #688 passes `cargo check --workspace --all-targets --locked`; this branch alone does not.

ScriptedAlchemy:
## Re-verified both reviewer findings against the current branch head — both are already fixed

I was asked to treat finding 1 as a production outage and fix it first. Before changing anything I re-ran the verification, and the good news is that **both findings were resolved upstream** while this branch was being advanced (36 commits, primarily `96564b7e9 perf(runtime): coordinate background worker resources`). The reviewer's findings were accurate against `04ff619b2`; they are stale against `6325658f4`.

### Finding 1 (P1, "CPU authority never installed in a real daemon") — FIXED upstream

At `04ff619b2` the finding was exactly right. `git grep install_process_background_cpu` returned only `host_admission_batch_test.rs`, so `HostAdmissionFacade::application` converted `None` into `Unavailable`/`background_cpu_unavailable` and every real host observation capture was rejected before preparation.

At `6325658f4` there is a real production installer:

- `crates/tracedecay-code-index/src/parallelism.rs:420` — `install_worker_plan` installs the authority at `requested.effective_workers`
- reached in production from `src/daemon/bootstrap.rs:140` / `:533` → `install_profile_worker_plan` → `src/daemon/invocation_state.rs:96`, and from `src/daemon/code_index_scheduler.rs:2846`

I also checked the follow-on question the finding implies — whether non-daemon processes can reach capture without the authority. They cannot:

- `src/serve.rs:152` — `serve` is "a database-free proxy to the managed daemon" (`proxy_stdio_to_daemon`); `src/serve.rs:69` disables direct project database access
- `crates/tracedecay-agent-hosts/src/hooks/claude.rs:371` — the hook handlers submit to the daemon (`daemon_hook_action`)
- `src/daemon/service/invocation/registrars.rs:697` states the design explicitly: profile worker-plan admission is the sole installer and "project registration must never invent or replace process CPU authority"

So the daemon is the sole capture host and the sole installer. **I did not re-add a CLI-side install.** I had one working (composition root, `install_process_cpu_authorities`, RED/GREEN proven), and discarded it: installing at the daemon CPU width would now collide with the worker plan's width and make `install_process_background_cpu` return `ConflictingWidth`, failing daemon startup. Rebasing before pushing is what caught that.

### Finding 2 (P2, "indexing bypasses the shared CPU budget") — FIXED upstream

The documented ceiling is now real rather than aspirational:

- `crates/tracedecay-code-index/src/parallelism.rs:546` — `with_background_cpu_permits` routes indexing work units through `authority.with_permits`
- `:572` — `install` wraps the pool in `authority.with_yielded_permits`, which is the nesting discipline that keeps a parent Rayon worker from holding a unit while waiting on children
- semantic work routes through the same `parallelism::install`, and its session width derives from the shared budget (`EmbeddingSessionLimitingReasonV1::SharedCodeIndexCpuBudget`)

With indexing, semantic, and session preparation all participating, the `background_cpu.rs` module doc claim is now accurate, so **no documentation correction was needed**. Had it still been overstated I would have corrected it rather than attempting the routing in this PR.

### What I actually changed: one regression test

Verifying finding 1 turned up a real gap: `background_cpu_unavailable` appears exactly once in the tree, at the production rejection site `crates/tracedecay-usecases/src/host_admission.rs:768`, with **no test coverage anywhere**. The contract that just cost this branch a P1 was unpinned — a regression dropping the install would resurface as silent ingest loss rather than a red test.

`crates/tracedecay-usecases/tests/host_capture_background_cpu_admission.rs` pins it in both directions against a real registered store (`HostAdmissionTestRuntimeV1::profile`), asserting on outcomes only, never elapsed time:

1. uninstalled → `capture_observations` fails closed with `status: Unavailable`, `reason_code: Some("background_cpu_unavailable")`
2. install the authority
3. installed → the same capture on the same facade is admitted, every frame `Persisted`/`AcceptedForReplay`, and `active_units() == 0` afterwards

It is its own integration binary because the authority is a process-wide `OnceLock` and the uninstalled half is only observable in a process no other test has initialized.

**Falsification** — asserting the opposite (that capture succeeds uninstalled) reproduces the exact outage verbatim:

```
thread 'capture_is_rejected_without_the_authority_and_admitted_with_it' panicked at
crates/tracedecay-usecases/tests/host_capture_background_cpu_admission.rs:160:10:
FALSIFICATION: claim capture succeeds with NO authority installed:
HostAdmissionOutcome { status: Unavailable, retryable: false,
                       reason_code: Some("background_cpu_unavailable"), recovery: None }
```

**Passing:**

```
running 1 test
test capture_is_rejected_without_the_authority_and_admitted_with_it ... ok

test result: ok. 1 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 1.62s
```

No concurrency ceiling was raised, and no production code changed. The width in the test is 2 and is documented as proving admission, not throughput. Per-crate builds only (`-p tracedecay-usecases`); `cargo clippy -p tracedecay-usecases --test host_capture_background_cpu_admission` reports zero findings in the new file.

PR #513, opened 2026-08-19
chore(release): release 0.0.74

ScriptedAlchemy:
:robot: I have created a release *beep* *boop*
---

### Bug Fixes

* **daemon:** abort watchers before joins ([fa93535](https://github.com/ScriptedAlchemy/tracedecay/commit/fa935352e75c9cdb9b22ea3f513ac247c002eaae))
* **daemon:** bound maintenance shutdown ([caba6eb](https://github.com/ScriptedAlchemy/tracedecay/commit/caba6eba7ae2a41bcd9c8827230a39db101331b0))
* **db:** restrict consolidation authority ([a556506](https://github.com/ScriptedAlchemy/tracedecay/commit/a556506d2eaf460fc62fb6f74603c17dc043e5fc))
* **migrate:** authorize staged databases ([850414b](https://github.com/ScriptedAlchemy/tracedecay/commit/850414bb1a91a3ee3ea89470a14de455332a9054))
* **storage:** fail closed on unreadable inventory ([6e92ed4](https://github.com/ScriptedAlchemy/tracedecay/commit/6e92ed4f3d5501700a666b3b103119413b70d6d1))
* **storage:** harden project store authority and resolution ([3050018](https://github.com/ScriptedAlchemy/tracedecay/commit/30500186caa9375d3daa145aee0f88803e51792f))
* **storage:** harden recovery validation ([71b19fc](https://github.com/ScriptedAlchemy/tracedecay/commit/71b19fc1f9ed03e8a5be7507d7d241b3dffd18aa))
* **storage:** honor aliases with repository markers ([91944bc](https://github.com/ScriptedAlchemy/tracedecay/commit/91944bc6ce0828e9556411de45be135249a2a677))
* **storage:** isolate branch recovery markers ([82876ea](https://github.com/ScriptedAlchemy/tracedecay/commit/82876ea591988eb0e3d242f1e3d727a18991d81d))
* **storage:** preserve exact-root precedence ([df96e9b](https://github.com/ScriptedAlchemy/tracedecay/commit/df96e9b5c2ba3aba162ab47c91777567d17ae181))
* **storage:** reject empty recovery stores ([043a047](https://github.com/ScriptedAlchemy/tracedecay/commit/043a047e03b87cdeb0bf77241a852ec2bf0f80d7))
* **storage:** trust exact worktree registry aliases ([a79f70c](https://github.com/ScriptedAlchemy/tracedecay/commit/a79f70cc3bb4b307ff17444a3adafaae8b52f670))
[...]

PR #677, opened 2026-08-24
perf(store): batch graph replay dependency reads

ScriptedAlchemy:
## What

`read_replays_by_sequences` (the batched paging path used by `replay_page`)
already fetches its whole replay set through one `IN (...)` query, but each
row was then decoded with `decode_row`, which called `read_dependencies`
**once per replay** to fetch that replay's `direct_dependency_generations`.
For a page of N replays that's N extra single-row queries against
`graph_publication_replay_dependencies_v1`.

This is the one site left over from #513/`perf(store): batch graph replay
reads and bound Busy retries` (14f728799) — that change batched the main
replay-row query and the blob reads, but left the dependency read as a
per-row call inside `decode_row`. Its siblings (batching the main replay
rows, the payload/blob reads, and the metadata page query) already landed;
this was the one still-open N+1 site from that audit.

## Fix

- Added `read_dependencies_batch(transaction, sequences, retired)`, which
  chunks `sequences` into `GRAPH_REPLAY_DEPENDENCY_BATCH`-sized slices (500,
  matching `REFERENCED_ANCHOR_BATCH` in
  `tracedecay-runtime-core/src/store/memory/crud/commit.rs`) and issues one
  `IN (...)` query per chunk instead of one query per replay.
- Split `decode_row` into a thin wrapper plus `decode_row_with_dependencies`,
  which takes already-fetched dependencies instead of querying for them.
- `read_replays_by_sequences` now calls `read_dependencies_batch` once for
  the whole sequence set, then decodes each row from the resulting map.
- `decode_row` itself (used by `one_replay`/`read_conflicts`, which are
  single-row or near-zero-cardinality conflict-detection paths, not "read a
  whole replay set" paths) is unchanged in behavior — it still calls
  `read_dependencies` per call, exactly as before.

## Why the batched form returns exactly the same data

`read_dependencies` (unchanged) selects one owner's dependency rows ordered
by `ordinal ASC`, checks the row count against a `LIMIT MAX+1` guard, checks
ordinal contiguity (`0..len`), and decodes each row into a
`GraphDependencyGenerationIdentityV1`.

`read_dependencies_batch` reproduces this per-owner exactly, just batched
across many owners in one query:

```sql
SELECT owner, ordinal, shard_id, namespace, projection, generation
FROM (
    SELECT owner_replay_sequence AS owner, ordinal, shard_id, namespace,
           projection, generation,
           ROW_NUMBER() OVER (
               PARTITION BY owner_replay_sequence ORDER BY ordinal ASC
           ) AS dependency_rank
    FROM graph_publication_replay_dependencies_v1
    WHERE owner_replay_sequence IN (...)
)
WHERE dependency_rank <= ?  -- MAX_GRAPH_REPLAY_DIRECT_DEPENDENCIES_V1 + 1
ORDER BY owner ASC, ordinal ASC
```

- The `ROW_NUMBER() OVER (PARTITION BY owner ...)` window reproduces
  `read_dependencies`'s per-owner `LIMIT {MAX+1}` *inside* the single
  batched query — each owner still contributes at most `MAX+1` rows, so a
  corrupt owner with a runaway dependency count is still caught (and still
  bounded in how many of its rows get read) rather than only detected after
  buffering everything.
- Rows are grouped by `owner` in Rust as they stream in, and for each
  owner the code checks `ordinal == entry.len()` (the next expected
  ordinal) before pushing — byte-for-byte the same contiguity check
  `read_dependencies` does with `rows.iter().enumerate()`.
- The same `entry.len() > MAX_GRAPH_REPLAY_DIRECT_DEPENDENCIES_V1` Corrupt
  error is raised per-owner if the count guard is somehow exceeded, with
  the identical error message.
- For an owner with **zero** dependency rows, `read_dependencies` returns
  `Vec::new()`; the batched map simply has no entry for that owner, and
  `read_replays_by_sequences` treats a missing map entry as `Vec::new()`
  via `.remove(&sequence).unwrap_or_default()` — same result.

One deliberate, called-out difference: the per-owner `LIMIT+1` bound in
`read_dependencies` also protects memory in a byzantine-corruption scenario
by never reading past `MAX+1` rows for *that one query*. The batched query
still bounds each owner to `MAX+1` rows via the window filter, but if many
owners in one chunk were all corrupt with runaway dependency counts, the
single query would buffer all of them (`chunk_size * (MAX+1)` rows in the
worst case) before the per-owner check fires, rather than failing on the
very first corrupt owner's query. This only matters for already-corrupt
data; correctness of the returned data and normal-case memory/query
behavior are identical.

## Order

`read_replays_by_sequences`'s main query is `ORDER BY sequence ASC`, and
rows are decoded via `rows.into_iter().map(...)`, preserving that order —
unchanged. Only the *dependency* lookup went through a batched `IN (...)`
(which doesn't preserve input order), so dependencies are looked up by
keying the returned map on each row's own `sequence` (`dependencies_by_owner
.remove(&sequence)`), never by positional/array order. The one caller that
depends on ordering — `exact.rs`'s `replay_page`, which takes
`records.last()` to build the next-page cursor — sees no change: the
records vector's order is still exactly the main query's `ORDER BY sequence
ASC`, only the per-row dependency fetch moved from N queries to O(N /
GRAPH_REPLAY_DEPENDENCY_BATCH) queries.

## Empty set

`read_dependencies_batch` returns an empty map immediately for an empty
`sequences` slice (mirroring the existing empty-set guard already in
`read_replays_by_sequences`), before ever building a chunk or a query —
no `IN ()` is ever constructed.

## Bound parameter limit

`sequences.chunks(GRAPH_REPLAY_DEPENDENCY_BATCH)` (500, matching
`REFERENCED_ANCHOR_BATCH`) keeps every batched query's parameter count
under SQLite's `SQLITE_MAX_VARIABLE_NUMBER` regardless of how large a
replay page or sequence set is.

## RED/GREEN evidence (query counts and returned rows, not timing)

New test:
`crates/tracedecay-rusqlite-runtime/src/repository/graph_publication/support.rs`
→ `dependency_batch_tests::read_dependencies_batch_matches_looped_reads_in_one_query`.

It appends 3 owner replays, each depending on the same 2 verified
dependency targets, wraps a real `ExactSqlTransaction` in a `QueryCounter`
that counts every query whose SQL text touches
`graph_publication_replay_dependencies_v1`, and asserts:

- `read_dependencies_batch` for all 3 owners issues **exactly 1** query.
- Calling `read_dependencies` once per owner (the pre-fix shape) issues
  **exactly 3** queries (one per replay — reproduced deliberately, in the
  same test, as the shape this change removes).
- The two produce byte-for-byte identical per-owner dependency data
  (content and ordinal order).

**GREEN (current code):**

```
test repository::graph_publication::exact::support::dependency_batch_tests::read_dependencies_batch_matches_looped_reads_in_one_query ... ok
test result: ok. 1 passed; 0 failed; 0 ignored; 0 measured; 299 filtered out; finished in 0.03s
```

**RED (temporarily reverted `read_dependencies_batch` to loop over
`read_dependencies` per sequence — the pre-fix N+1 shape — to prove the
test actually catches the regression; reverted back before committing):**

```
thread '...read_dependencies_batch_matches_looped_reads_in_one_query' panicked at .../support.rs:1520:9:
assertion `left == right` failed: batched dependency read must issue exactly one query for the whole replay set, regardless of how many replays it covers
  left: 3
 right: 1
test result: FAILED. 0 passed; 1 failed; 0 ignored; 0 measured; 299 filtered out; finished in 0.03s
```

Full `graph_publication` suite after restoring the fix (22/22, including
the new test):

```
test result: ok. 22 passed; 0 failed; 0 ignored; 0 measured; 278 filtered out; finished in 0.22s
```

## Verification

- `cargo check -p tracedecay-rusqlite-runtime --all-targets --locked` — clean.
- `cargo fmt -p tracedecay-rusqlite-runtime -- --check` — clean.
- `cargo test -p tracedecay-rusqlite-runtime --locked --lib graph_publication` — 22/22 passing.

## SCOPE DEVIATION

None. The change is limited to `read_dependencies_batch`,
`decode_row_with_dependencies`, and `read_replays_by_sequences` in
`support.rs`, plus the new test module. `decode_row`, `decode_tombstone_row`,
`read_conflicts`, and `read_tombstone_conflicts` — the other call sites of
`read_dependencies`/`decode_row`/`decode_tombstone_row` — are single-key or
near-zero-cardinality conflict-detection paths, not "read a whole replay
set" paths, and were intentionally left untouched (they were not named by
the audit and are not the N+1 site this task targets).


ScriptedAlchemy:
Addressed both P1 review findings in `115a14b9f`.

**Finding 1 — chunk size exceeded the exact-SQL row cap.** Confirmed the reviewer's numbers directly: `exact_sql/mod.rs:29` sets `MAX_QUERY_ROWS = 10_000`, enforced at line 927 (`ExactSqlError::QueryLimitExceeded`), and `query()` in `support.rs` maps that to `GraphPublicationStoreErrorV1::Infrastructure`. `MAX_GRAPH_REPLAY_DIRECT_DEPENDENCIES_V1 = 256` (`tracedecay-store/src/runtime/graph_publication.rs:11`), and the batched query's `ROW_NUMBER() ... <= MAX + 1` sentinel caps each owner at 257 rows — so `REFERENCED_ANCHOR_BATCH`-style 500-owner chunking bounds *parameters*, not the *row* budget that actually gates this query. `GRAPH_REPLAY_DEPENDENCY_BATCH` is now derived from the row budget: `10_000 / 257 = 38` (38\*257=9,766 ≤ 10,000; 39\*257=10,023 > 10,000), with the arithmetic written into the constant's doc comment.

**Finding 2 — test gated on SQL text.** `QueryCounter` no longer filters on `statement.sql.contains(needle)`; it now counts every `ExactQueryAuthority::exact_query` invocation directly. Each counter is still built fresh immediately before its one isolated operation, so no text filter is needed to keep unrelated statements out — this is the same house pattern noted for #673.

**RED/GREEN, both findings, real output:**

Finding 1 — new test `read_dependencies_batch_stays_under_the_row_cap_for_a_full_dependency_page` reproduces the reviewer's exact worked example (40 owners × 256 dependencies = 10,240 rows in one chunk under the old size):

RED (constant temporarily reverted to 500):
```
thread '...read_dependencies_batch_stays_under_the_row_cap_for_a_full_dependency_page' panicked at support.rs:1734:17:
read_dependencies_batch must stay under the exact-SQL row cap for a full 40-owner page with 256 dependencies each (10240 rows): Infrastructure
test result: FAILED. 1 passed; 1 failed; 0 ignored; 0 measured; 299 filtered out; finished in 0.37s
```

GREEN (restored to 38):
```
test ...read_dependencies_batch_matches_looped_reads_in_one_query ... ok
test ...read_dependencies_batch_stays_under_the_row_cap_for_a_full_dependency_page ... ok
test result: ok. 2 passed; 0 failed; 0 ignored; 0 measured; 299 filtered out; finished in 0.37s
```

Finding 2 — verified the rewritten (non-string-matching) `QueryCounter` is actually falsifiable by temporarily reverting the existing test's batched call to the pre-fix per-replay loop (while keeping the `== 1` assertion), then restoring:

RED (batched call temporarily swapped for a loop):
```
thread '...read_dependencies_batch_matches_looped_reads_in_one_query' panicked at support.rs:1563:9:
assertion `left == right` failed: batched dependency read must issue exactly one query for the whole replay set, regardless of how many replays it covers
  left: 3
 right: 1
test result: FAILED. 0 passed; 1 failed; 0 ignored; 0 measured; 300 filtered out; finished in 0.02s
```

GREEN (restored):
```
test ...read_dependencies_batch_matches_looped_reads_in_one_query ... ok
test ...read_dependencies_batch_stays_under_the_row_cap_for_a_full_dependency_page ... ok
test result: ok. 2 passed; 0 failed; 0 ignored; 0 measured; 299 filtered out; finished in 0.37s
```

`cargo check -p tracedecay-rusqlite-runtime --all-targets --locked` and `cargo fmt -p tracedecay-rusqlite-runtime -- --check` are clean.

Preserved everything the PR got right: batched `read_dependencies_batch`, the `ROW_NUMBER() OVER (PARTITION BY owner ...)` reproduction of the per-owner `LIMIT MAX+1` corruption guard, and the `ORDER BY sequence ASC` on the main query.

ScriptedAlchemy:
Integrated directly into PR #421 after #663 as 08adf019b and 07a0347ee. The follow-up caps owners at 38 so every dependency query remains below Exact SQL’s 10,000-row bound.

PR #678, opened 2026-08-24
docs(memory): describe the current retrieval and trust model

ScriptedAlchemy:
## What changed

Rewrites the bodies of `docs/RETRIEVAL-QUALITY-EVAL.md` and `docs/TRUST-DECAY-SEMANTICS.md`, which described a single-crate `src/memory/{retrieval,store}.rs` layout that was fully rewritten into `crates/tracedecay-runtime-core/src/store/memory/{candidates.rs,scoring.rs,search.rs}` (async, `project_memory_*`-prefixed, `memory_v2_*` tables). The temporary "Status: historical" banners added in an earlier pass are removed now that the bodies themselves are accurate.

Verified directly against the current source before writing (no invented `file:line` anchors — locations are described by file/function/table name in prose):

- **Risk A is resolved**, not merely stale. `crates/tracedecay-runtime-core/src/memory/entities.rs` now strips inflected leading verbs (`is_non_entity_leading_word` matches `Prefers`/`Uses`/etc., not just exact base forms) and extracts single capitalized proper nouns (`Postgres`, `Tokio`, `Kubernetes`, `Database`) that used to require a ≥2-word phrase. Both are covered by dedicated unit tests in `entities.rs`, which the doc now cites by name.
- **Two previously-undocumented behaviors** are now described: FTS5 prefix-wildcarding for query tokens ≥4 chars (`project_memory_fts_query`), and a retrieval-count `usage_boost` term in `project_memory_combined_score` (capped at 1.5x via `ln(1+n)`).
- The trust constants (`HELPFUL_DELTA = 0.05`, `UNHELPFUL_DELTA = -0.10`, `DEFAULT_TRUST = 0.5`, `DEFAULT_MIN_TRUST = 0.3`) and the 365-day half-life / 0.10-floor decay formula are confirmed unchanged, now sourced to `memory/trust.rs` and `store/memory/scoring.rs::project_memory_temporal_decay`.
- Trust is now event-sourced: every `trust_score` mutation is a `FactLineageEventKindV1::TrustChanged` event replayed into `memory_v2_current_facts`, not an in-place `UPDATE`. There is no longer a separate "manual trust-delta bump" MCP action — an absolute trust set is just one field of the general fact-update patch.

## What I found that pushes back on the brief

- The task brief described this as two purely documentation-only fixes on top of an otherwise-intact pipeline. In practice large parts of both documents needed a full rewrite: the FTS score normalization changed from a `1/(1+|bm25|)` formula to a max-relative normalization plus a coverage multiplier, the recall gate's shape changed slightly, and a new graph-assist candidate-expansion channel was added that has no analog in the old pipeline at all. None of this was called out in the task description, so I documented it as new/changed behavior rather than silently reusing the old framing.
- I could not re-run the original fixture-based empirical measurement (the exact score tables in the old `RETRIEVAL-QUALITY-EVAL.md` §2) because the scoring formula changed enough that the old numbers would not replay verbatim, and re-running it would require rebuilding the CLI from this exact worktree — out of scope for a docs-only change. I kept the qualitative risk conclusions that are still true per the source (no stemming, trust-as-hard-gate, decorative holographic signal, probe/reason ignoring relevance) but explicitly flagged the old numeric table as retired rather than re-publishing stale numbers under a new banner.
- The old TRUST-DECAY-SEMANTICS doc described a "manual MCP trust bump" tool with its own `trust_delta` argument, distinct from the edit path. That tool no longer exists as a separate concept — the current `ProjectMemoryFactUpdatePatchV1` just carries an optional absolute `trust` field alongside content/tags/etc. I documented the current (merged) shape rather than preserving the old distinction.

## Scope

Docs only — no `.rs`/`.ts`/workflow files touched (verified via `git diff --stat`).

### DIFF
--- a/docs/RETRIEVAL-QUALITY-EVAL.md
+++ b/docs/RETRIEVAL-QUALITY-EVAL.md
@@ -1,269 +1,308 @@
 # Holographic Memory — Retrieval & Entity Quality Evaluation
 
-Task `t_3d566784` ("Evaluate retrieval and entity quality"). This document maps the
-current recall pipeline, then reports **empirically measured** quality risks gathered by
-driving the real `tracedecay` binary (`target/debug/tracedecay tool fact_store`) against a
-throwaway fixture project, and proposes concrete follow-ups.
+This document maps the current project-memory recall pipeline (candidate
+generation, scoring, fusion) and the entity extractor that feeds it, then
+records which of the previously-identified quality risks still hold against
+the current implementation.
 
-All evidence below is reproducible: every query/probe/reason result was produced by the
-actual shipped binary against a freshly-initialized temp project (temp `HOME`, temp project
-dir — the shared repo `.tracedecay/` was never touched). Per-signal scores (`fts_score`,
-`jaccard_score`, `holographic_score`, `trust_score`) come straight from the `why` field the
-search handler emits.
-
----
-
-> **Status: historical — code layer superseded.** Every `retrieval.rs:NNN` /
-> `entities.rs:NNN` / `encoding.rs:NNN` / `migrations.rs:NNN` / `trust.rs:NNN`
-> anchor below was written against a single-crate `src/memory/` layout that no
-> longer exists. The retrieval pipeline was rewritten and split into
-> `crates/tracedecay-runtime-core/src/store/memory/{candidates.rs,scoring.rs,search.rs}`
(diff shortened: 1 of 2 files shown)


ScriptedAlchemy:
Merged directly into PR #421 after #663 as c0f414565.

PR #679, opened 2026-08-24
test(writer): pin the batch-to-frame coalescing ratio

ScriptedAlchemy:
## Current-state finding: the 14:1 ratio was already fixed, twice

I was sent to cut writer transaction amplification from a measured 26,382
`submit_authorized` calls settling as 26,380 durable transactions for 1,878
persisted frames. **I did not reimplement that fix, because it already
exists — in two independent places.** Both landed after the binary those
numbers were measured on.

1. **Upstream, in the stores.** `persist_observations` submits one writer
   request per scan batch instead of one per frame
   (`crates/tracedecay-sessions/src/runtime/jsonl_observation_admission.rs`,
   `MAX_CAPTURE_WINDOW = 256`). This is already regression-covered at two
   layers, both asserting on transaction counts rather than time:
   - `tracedecay-global-db`: `persist_observations_opens_one_writer_transaction_for_the_batch`
   - `tracedecay-usecases`: `persist_observations_opens_one_writer_transaction_for_the_batch`,
     paired with `n_persist_observation_calls_open_n_writer_transactions` as
     its negative control.

2. **In the writer worker.** A peer's local commit `perf(writer): coalesce
   compatible queued writes` adds a dwell window so the queue accumulates
   compatible arrivals before the worker drains it. Without it, `build_batches`
   had nothing to coalesce: a sequential submitter awaits each settle before
   submitting again, so the queue was empty at every wake and the average
   batch was exactly 1.0 — which is precisely the measured 26,382:26,380.
   **That commit is not in this PR and I did not cherry-pick it.**

So the honest answer to "is problem 1 already fixed?" is **yes**. The
remaining ratio is set by the batch budget, not by the frame count.

### What this PR actually adds

One gap survived both fixes: **`build_batches` itself has no assertion in
this crate.** It is the pure function that groups queued requests into
transactions, and every existing assertion sits *above* it — constraining how
many requests get submitted, not how they get grouped. The stores can stay
correct and the dwell window can keep filling the queue while this function
silently regresses to one batch per request, and nothing currently fails.

Three tests, all asserting on counts:

- `one_scan_batch_of_compatible_writes_opens_one_transaction` — 256 compatible
  foreground requests under a 512-operation budget must produce exactly 1
  batch. Ratio `0.0039` batches/frame.
- `batch_count_is_bounded_by_the_budget_not_by_the_request_count` — 256
  requests under a 64-operation budget must produce exactly 4 batches, and no
  batch may exceed its budget.
- `an_isolated_commit_still_gets_its_own_transaction` — durability guard, see
  below.

**No timing assertions.** The profile this work came from swings ~12% run to
run on identical input, so an elapsed-time assertion would prove nothing. The
failure messages quote the ratio so a regression reports the number directly.

## Durability argument

**This PR widens no transaction window and changes no production code.** It is
test-only against `crates/tracedecay-rusqlite-runtime/src/writer/worker/mod.rs`.
Durability semantics are therefore unchanged by construction — there is no
at-least-once coverage to lose, no commit order to perturb, and no write lock
held for one additional instruction.

On the coalescing that *does* exist, the two properties that matter are already
structural rather than incidental, and the third test pins one of them:

- **The write lock covers only the commit.** `process_batch` takes
  `BEGIN IMMEDIATE` and then does all per-request work inside savepoints on an
  already-open transaction. No parse and no I/O wait happens between the lock
  being taken and the commit — the dwell window that accumulates requests runs
  *before* the transaction begins, not inside it.
- **A request that asked to commit alone still does.** `build_batches` splits
  the batch on `requires_isolated_commit()`. Widening a window must never
  quietly absorb such a request, so
  `an_isolated_commit_still_gets_its_own_transaction` asserts the split
  survives.
- **Batch failure is not silently shared.** Already handled and already
  covered: `settle_authority_denied` gives the members that did *not* lose
  authority a retryable `Faulted` rather than blaming them for a peer's
  revocation.

## Problem 2: the `begin_immediate` tail is not lock contention — diagnosis, no fix

I was asked to stop and report rather than guess if this needed a design
change. It does. Reporting.

The symptom: `rusqlite.begin_immediate` has a **mean of 10.65 ms against a p95
of 82 µs** — ~130x — across 4,900 calls. That shape says nearly every call is
instant and a handful block enormously. The natural reading is that
`BEGIN IMMEDIATE` is fighting for SQLite's write lock.

**That reading is wrong, and the metric name is what makes it look right.**

**Evidence chain:**

1. `rusqlite.begin_immediate` is emitted from exactly one place —
   `crates/tracedecay-rusqlite-runtime/src/exact_sql/mod.rs:427`. It is **not**
   the writer's transaction begin; the writer emits `rusqlite.writer.begin`
   from `writer/transaction.rs`. This is why its 4,900 calls never reconciled
   with `process_batch`'s 26,380 — they are different paths.

2. The measured span is not a SQLite statement. `begin_transaction`
   (`exact_sql/mod.rs:466`) does `try_send(WriterCommand::BeginTransaction)`
   onto the writer actor's exact-SQL channel and then **blocks on
   `response.recv()`**. The span therefore includes the full wait for the
   writer worker thread to get around to the command.

3. The worker deliberately deprioritises that command.
   `select_auxiliary_work` returns `None` whenever product writes are queued
   and `prefer_auxiliary` is false. In the main loop, `prefer_auxiliary` is
   reset to `true` only *after* `queue.drain_fair()` has been fully processed
   — every batch from `build_batches`, with `run_scheduled_checkpoint`
   interleaved between them.

4. So an exact-SQL `BeginTransaction` arriving mid-drain waits for the entire
   remaining drain plus its interleaved checkpoints. When the worker is idle it
   is served immediately — that is the 82 µs p95.

5. It compounds: `run_writer_command` executes `run_transaction(...)`
   **inline on the worker thread** (`exact_sql/command.rs:186-228`), driven by
   the caller over a `sync_channel(1)`. The writer actor is occupied for the
   whole lifetime of an exact-SQL transaction, blocked waiting on the caller's
   next statement. Product writes and exact-SQL transactions mutually exclude
   each other on one actor.

**Consequences for how this should be read:**

- `rusqlite.begin_immediate` is **actor-queue scheduling latency**, not lock
  acquisition. Its mean is a fairness property of the writer worker. Treating
  it as I/O or as SQLite contention will send the next person down the wrong
  path, which is roughly what its name invites.
- **Widening the writer's transaction window makes this metric worse, not
  better** — a longer drain is a longer wait for the exact-SQL command behind
  it. That is a direct argument against "fix the tail by batching harder".
- The peer's dwell window does **not** worsen it: `dwell_for_batch` returns
  immediately on any non-`Write` wake, so an arriving exact-SQL command ends
  the dwell rather than waiting it out. Good design, worth preserving.

**Why I did not fix it.** Every available option is a design change, not a
local edit: give exact-SQL its own write connection (changes the single-writer
invariant), make the drain yield to auxiliary work between batches (changes
fairness and starvation guarantees), or split the metric into queue-wait vs
actual `BEGIN` (honest, but a contract change to the profile everyone is
reading). Picking one of those unbriefed is exactly the guessing I was told to
avoid.

**Cheapest useful next step**, if someone wants it: split the
`rusqlite.begin_immediate` span into `rusqlite.exact_sql.queue_wait` and
`rusqlite.exact_sql.begin`. That alone converts "24.3% of measured total,
cause unknown" into two numbers with obvious owners, and costs no behaviour
change. I have not done it here because it changes a metric others are
currently comparing against.

## Carried commit

This branch carries `fix(runtime-core): commit the missing background CPU
module` (`origin/claude/fix-missing-background-cpu`, PR #676). The base branch
does not compile without it — `crates/tracedecay-sessions/src/observation.rs`
imports `tracedecay_runtime_core::background_cpu`, whose module file was never
`git add`ed as part of `perf(observation): batch admission and preparation`. It
is purely additive. If #676 merges first this will fold away cleanly.

## Merge-order note

`crates/tracedecay-rusqlite-runtime/src/writer/worker/mod.rs` is also heavily
modified by the peer's unpushed `perf(writer): coalesce compatible queued
writes` (+803 lines in that file). **Expect a textual conflict there.** It is
additive on both sides — my tests are a new `batch_coalescing_tests` module at
the end of the file and touch nothing the dwell work touches. Resolving by
keeping both modules is correct. Landing the peer's commit first is the lower
friction order.

## Verification

Per-crate only; no workspace build or workspace clippy.

- `cargo check -p tracedecay-rusqlite-runtime --all-targets --locked` — clean
- `cargo check -p tracedecay-rusqlite-runtime --all-targets --locked --features hotpath` — clean
- `cargo clippy -p tracedecay-rusqlite-runtime --all-targets --locked` — no warnings
- `cargo test -p tracedecay-rusqlite-runtime --lib writer::` — 33 passed, 0 failed

hotpath remains off by default.

### DIFF
--- a/crates/tracedecay-runtime-core/src/background_cpu.rs
+++ b/crates/tracedecay-runtime-core/src/background_cpu.rs
@@ -1,0 +1,461 @@
+//! Process-wide weighted admission for background CPU work.
+//!
+//! The authority counts active CPU units rather than owning an executor. Code
+//! indexing, semantic native threads, and session preparation can therefore
+//! use their existing execution substrates while sharing one hard process
+//! ceiling. FIFO waiter order prevents a continuously busy class from starving
+//! another class, and RAII releases capacity on success, cancellation, or
+//! unwind.
+
+use std::cell::Cell;
+use std::collections::VecDeque;
+use std::fmt;
+use std::num::NonZeroUsize;
+use std::sync::{Arc, Condvar, Mutex, OnceLock};
+
+#[derive(Debug)]
+struct BackgroundCpuWaiterV1 {
+    units: usize,
+}
+
+#[derive(Default)]
+struct BackgroundCpuStateV1 {
+    active_units: usize,
+    waiters: VecDeque<Arc<BackgroundCpuWaiterV1>>,
+}
+
+/// One process-wide background CPU budget shared across subsystems.
+pub struct ProcessBackgroundCpuV1 {
+    width: NonZeroUsize,
+    state: Mutex<BackgroundCpuStateV1>,
+    available: Condvar,
+}
+
+impl fmt::Debug for ProcessBackgroundCpuV1 {
+    fn fmt(&self, formatter: &mut fmt::Formatter<'_>) -> fmt::Result {
+        formatter
+            .debug_struct("ProcessBackgroundCpuV1")
+            .field("width", &self.width)
+            .field("active_units", &self.active_units())
+            .finish_non_exhaustive()
+    }
+}
+
+thread_local! {
+    static BACKGROUND_CPU_DEPTH: Cell<usize> = const { Cell::new(0) };
+    static BACKGROUND_CPU_UNITS: Cell<usize> = const { Cell::new(0) };
+}
+
+struct BackgroundCpuScopeV1;
+
+impl BackgroundCpuScopeV1 {
+    fn enter() -> Self {
+        BACKGROUND_CPU_DEPTH.with(|depth| depth.set(depth.get().saturating_add(1)));
+        Self
+    }
+}
+
+struct YieldedBackgroundCpuV1<'a> {
+    authority: &'a Arc<ProcessBackgroundCpuV1>,
+    units: usize,
+    depth: usize,
+}
+
+impl Drop for YieldedBackgroundCpuV1<'_> {
+    fn drop(&mut self) {
+        self.authority.admit_units(self.units);
+        BACKGROUND_CPU_UNITS.with(|units| units.set(self.units));
+        BACKGROUND_CPU_DEPTH.with(|depth| depth.set(self.depth));
+    }
+}
+
+impl Drop for BackgroundCpuScopeV1 {
+    fn drop(&mut self) {
+        BACKGROUND_CPU_DEPTH.with(|depth| {
+            let remaining = depth.get().saturating_sub(1);
+            depth.set(remaining);
+            if remaining == 0 {
+                BACKGROUND_CPU_UNITS.with(|units| units.set(0));
+            }
+        });
+    }
+}
+
+impl ProcessBackgroundCpuV1 {
+    fn new(width: NonZeroUsize) -> Self {
+        Self {
+            width,
+            state: Mutex::new(BackgroundCpuStateV1::default()),
+            available: Condvar::new(),
+        }
+    }
+
+    #[must_use]
+    pub const fn width(&self) -> NonZeroUsize {
+        self.width
+    }
+
+    #[must_use]
+    pub fn active_units(&self) -> usize {
+        self.state
+            .lock()
+            .unwrap_or_else(std::sync::PoisonError::into_inner)
+            .active_units
+    }
+
+    #[must_use]
+    pub fn waiting_work_units(&self) -> usize {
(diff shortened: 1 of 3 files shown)


ScriptedAlchemy:

PR #681, opened 2026-08-24
fix(build): land the batching commit's unpushed consumers

ScriptedAlchemy:
## Merge this first — it supersedes #676

`cursor/simplify-pr421-hot-paths` does not compile. `perf(observation): batch admission and preparation` landed changes whose consumers and one whole file were never `git add`ed, so the work exists only in one local working tree. Every branch cut from the pushed head fails **before its own code is built** — that is currently every open PR and every agent working against this branch.

Four separate breakages, all the same root cause:

| Missing | Referenced by |
|---|---|
| `crates/tracedecay-runtime-core/src/background_cpu.rs` (whole file + `mod` line) | `observation.rs` import |
| `ObservationApplicationError::BatchWorkerStopped` arm | `runtime/ingest/failure.rs:517` |
| `HostAdmissionOutcome.recovery` field | `kiro.rs`, `cline_like.rs`, `ingest/tests.rs` ×2 |
| `CodeIndexWorkerPlanInstallErrorV1`, `worker_reservation_bytes`, `installed_worker_status`, `with_background_cpu_permit` | `src/daemon/code_index_scheduler.rs` |

The last one is a ~694-added/170-removed-line delta in `crates/tracedecay-code-index/src/parallelism.rs` — the in-flight worker-planning work. It is carried here as its original commit rather than reconstructed.

## One judgement call, flagged for the author

`BatchWorkerStopped` needed a classification and none was written. It is classified **retryable / `Unavailable`** with reason code `observation_batch_worker_stopped`: a worker that went away before reaching a verdict decided nothing about the payload, so re-running the same input can succeed. Classifying it `permanent` would silently drop observations on a transient worker fault. If the author intended different semantics, this is the line to change.

## Verification

`cargo check -p tracedecay-sessions --all-targets` clean. `Cargo.lock` is updated because the carried commit adds dependencies.

Two agents independently hit these and reported rather than repairing, which is how the scope got established rather than guessed.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -6143,6 +6143,7 @@ dependencies = [
  "tracedecay-code-extraction",
  "tracedecay-domain",
  "tracedecay-graph-db",
+ "tracedecay-runtime-core",
  "tracedecay-rusqlite-runtime",
  "tracedecay-store",
  "tree-sitter",
--- a/crates/tracedecay-code-index/Cargo.toml
+++ b/crates/tracedecay-code-index/Cargo.toml
@@ -16,6 +16,7 @@ hotpath = [
     "tracedecay-application/hotpath",
     "tracedecay-code-extraction/hotpath",
     "tracedecay-graph-db/hotpath",
+    "tracedecay-runtime-core/hotpath",
 ]
 hotpath-alloc = [
     "hotpath",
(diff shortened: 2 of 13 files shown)


ScriptedAlchemy:
Closing in favour of **#688**, which is the same fix minus one commit that cannot land on this base.

#681 also carried `crates/tracedecay-code-index/src/parallelism.rs`. That commit changes `parallelism::install` to return a `Result` and removes `install_daemon_worker_ceiling`, which breaks `tracedecay-semantic` (`embedding_parallelism.rs`) and `tracedecay-cli` (`main.rs:207`) on this base — it needs its own unpushed follow-up to be coherent, and supplying `available_memory_bytes` to the new plan API is a design call belonging to its author, not something a build fix should invent.

#688 covers `background_cpu.rs`, the `BatchWorkerStopped` arm, all five `HostAdmissionOutcome.recovery` sites, and the readiness accessors, and passes `cargo check --workspace --all-targets --locked` clean.

PR #682, opened 2026-08-24
perf(codex): decide rollout scope before the frame is decoded

ScriptedAlchemy:
Of 1,905 frames a hotpath-instrumented ingest run skipped, **1,886 (99%) were skipped as `out_of_scope`**. Each was read, decoded into a `serde_json::Value` tree, and walked by the structural validator *before* anything consulted its scope — the scope test lives inside the normalizer that `parse_normalized_observation_record_v1` calls last. This moves that decision ahead of the decode.

The set of admitted frames, and the coverage reason written for every skipped one, is unchanged. That is the claim to review; the speedup is not.

## Where the 1,886 come from

`ObservationCoverageReason::OutOfScope` reaches `record_frame_skipped` from exactly one place: the normalizer closure in `crates/tracedecay-sessions/src/runtime/codex/observation.rs`. The other `admit_jsonl_observations` callers (vibe, kimi, cursor) never construct it, and `skipped_reason()` only produces `BlankFrame`/`OversizedFrame`. So the whole 99% is Codex rollout admission: a scope is handed every rollout under a profile or project sweep, and owns almost none of them.

## Equivalence argument

**Claim.** For a frame `F` observed under context state with cwd `C`, if
1. `!scope_matcher.accepts(C)`, and
2. `probe_codex_frame(F) == Inert`,

then rejecting `F` as `OutOfScope` before decoding is *identical* — same disposition, same coverage reason — to what the existing path produces after decoding.

**Proof.** The existing path, for each frame, is:

```
observe_context_record(&native, ..)      // may move the cwd
if !scope_matcher.accepts(state.cwd) -> OutOfScope
```

reached only after `parse_normalized_observation_record_v1` clears, in order: `validate_record_frame` (empty / over `MAX_OBSERVATION_RECORD_BYTES` / range-length mismatch), `serde_json::from_slice::<Value>`, `is_object`, and `validate_structure` (depth ≤ 96, values ≤ 50,000). Failing any of those returns before `non_durable_reason` is ever set, so the frame is covered as `MalformedFrame` — a *different* reason. An early check therefore has to prove both halves.

*Half one — the cwd cannot move.* `observe_context_record` mutates `self.cwd` only via `session_meta_from_record` or `turn_context_from_record`, and both return `None` unless the record's **top-level `type`** is the string `"session_meta"` / `"turn_context"` (`crates/tracedecay-sessions/src/runtime/codex/meta.rs:67,148`). Nothing else in either function is reached first. `Inert` is returned only when the probe has deserialized that exact field and found neither value, so `observe_context_record` returns `false` and the post-state cwd is `C`. The late check then evaluates `!accepts(C)`, which is premise 1. Same verdict.

*Half two — no earlier gate would have fired.* The probe replicates `validate_record_frame` literally (emptiness, byte cap, `range.end() - range.start()` against `record.len()`; Codex admits only in the `FileBytes` domain, which is the domain that check applies to), runs the same `serde_json` tokenizer over the same bytes, requires the top level to be a map, calls `Deserializer::end()` so trailing content is rejected the way `from_slice` rejects it, and charges the same depth/value budget `validate_structure` charges. Any failure yields `Undecided`, never `Inert`.

**Why the `type` field is read, not scanned.** `Value::as_str()` returns the *decoded* string, so `{"type":"session_meta"}` **is** a `session_meta` record. A raw-substring scan for `session_meta` would call that frame inert and wrongly skip it. The probe deserializes the field through serde, so it decodes escapes identically. Both escaped encodings are in the test corpus and both come back `Undecided`.

**Where the probe is deliberately imprecise, and in which direction.** A `serde_json::Value` map collapses duplicate keys; a streaming walk sees both. So the probe's value count can *exceed* the validator's, pushing a frame over the budget that the validator would have accepted. That turns `Inert` into `Undecided` — it costs a decode and changes no verdict. It can never turn `Undecided` into `Inert`. Duplicate `type` keys resolve last-wins in both, matching the map.

`Undecided` is always safe: the frame takes the existing path byte-for-byte.

## Falsifiability

Both tests assert on counters, never on elapsed time.

- `inert_is_claimed_only_where_the_authoritative_pipeline_agrees` is differential: for every fixture the probe calls `Inert`, it runs the **real** `parse_normalized_observation_record_v1` and the **real** context classifiers over the same bytes and requires the parser to have reached the normalizer and the classifiers to disclaim the cwd. It does not restate the probe's rules.
- `out_of_scope_frames_are_refused_before_they_are_decoded` runs a real profile admission whose cwd is owned by a registered root. The `session_meta` line *can* move the cwd, so it must still be decoded; the message line cannot, so it must not be. The test pins `frames_rejected_before_decode == 1` and asserts every coverage advance still carries `OutOfScope`.
- `in_scope_frames_still_admit_in_full` is the counterpart: nothing in scope may be refused early.

RED was demonstrated two ways — neutering the fast path (the counter goes to 0, and `frames_decoded`/`frames_skipped`/`frames_persisted` stay put, so it is the parse assertion that fires), and making the probe unsound by always answering `Inert` (caught in three places, including "probe claimed inert for a frame that can move the session cwd"). Verbatim output is in the agent report.

## Counter, not a shadow metric

`frames_rejected_before_decode` rides the existing `JsonlObservationAdmissionProgress` → `CodexJsonlAdmissionProgress` production data flow and is emitted as `sessions.jsonl.frames.skipped.before_decode` through the existing hotpath authority. `record_frame_skipped`'s own doc comment already asked for this split: it notes that the aggregate cannot tell a blank line's decode from an out-of-scope frame's full parse. No new metrics authority, no test-only port. hotpath stays off by default.

## Verification

- `cargo check -p tracedecay-sessions --all-targets --locked` — no new errors (see caveat)
- `cargo check -p tracedecay-sessions --all-targets --locked --features hotpath` — clean
- `cargo clippy -p tracedecay-sessions --all-targets --locked` — clean
- `cargo test -p tracedecay-sessions --locked --lib` — 659 passed, 0 failed

## Two things a reviewer should know

**This branch carries a cherry-pick.** `origin/claude/fix-missing-background-cpu` (#676) is included because the base does not compile without it. When #676 merges, the duplicate resolves cleanly.

**The base has three further pre-existing compile errors**, unrelated to this change and untouched by it, reported rather than repaired:
- `runtime/ingest/failure.rs:517` — match on `ObservationApplicationError` does not cover `BatchWorkerStopped` (fires for lib and lib test)
- `runtime/kiro.rs:1331`, `runtime/cline_like.rs:1413`, `runtime/ingest/tests.rs:233,253` — `HostAdmissionOutcome` initializers missing the new `recovery` field

They look like the tail of `64dbb4bf1 perf(observation): batch admission and preparation`. Confirmed pre-existing by stashing this work and re-running: identical error set. They were patched in the working tree only, to let this crate build for verification, and reverted before committing — this branch contains none of them. After reverting, `cargo check` returns exactly the base's error set, so this change adds zero new errors.

### DIFF
--- a/crates/tracedecay-runtime-core/src/background_cpu.rs
+++ b/crates/tracedecay-runtime-core/src/background_cpu.rs
@@ -1,0 +1,461 @@
+//! Process-wide weighted admission for background CPU work.
+//!
+//! The authority counts active CPU units rather than owning an executor. Code
+//! indexing, semantic native threads, and session preparation can therefore
+//! use their existing execution substrates while sharing one hard process
+//! ceiling. FIFO waiter order prevents a continuously busy class from starving
+//! another class, and RAII releases capacity on success, cancellation, or
+//! unwind.
+
+use std::cell::Cell;
+use std::collections::VecDeque;
+use std::fmt;
+use std::num::NonZeroUsize;
+use std::sync::{Arc, Condvar, Mutex, OnceLock};
+
+#[derive(Debug)]
+struct BackgroundCpuWaiterV1 {
+    units: usize,
+}
+
+#[derive(Default)]
+struct BackgroundCpuStateV1 {
+    active_units: usize,
+    waiters: VecDeque<Arc<BackgroundCpuWaiterV1>>,
+}
+
+/// One process-wide background CPU budget shared across subsystems.
+pub struct ProcessBackgroundCpuV1 {
+    width: NonZeroUsize,
+    state: Mutex<BackgroundCpuStateV1>,
+    available: Condvar,
+}
+
+impl fmt::Debug for ProcessBackgroundCpuV1 {
+    fn fmt(&self, formatter: &mut fmt::Formatter<'_>) -> fmt::Result {
+        formatter
+            .debug_struct("ProcessBackgroundCpuV1")
+            .field("width", &self.width)
+            .field("active_units", &self.active_units())
+            .finish_non_exhaustive()
+    }
+}
+
+thread_local! {
+    static BACKGROUND_CPU_DEPTH: Cell<usize> = const { Cell::new(0) };
+    static BACKGROUND_CPU_UNITS: Cell<usize> = const { Cell::new(0) };
+}
+
+struct BackgroundCpuScopeV1;
+
+impl BackgroundCpuScopeV1 {
+    fn enter() -> Self {
+        BACKGROUND_CPU_DEPTH.with(|depth| depth.set(depth.get().saturating_add(1)));
+        Self
+    }
+}
+
+struct YieldedBackgroundCpuV1<'a> {
+    authority: &'a Arc<ProcessBackgroundCpuV1>,
+    units: usize,
+    depth: usize,
+}
+
+impl Drop for YieldedBackgroundCpuV1<'_> {
+    fn drop(&mut self) {
+        self.authority.admit_units(self.units);
+        BACKGROUND_CPU_UNITS.with(|units| units.set(self.units));
+        BACKGROUND_CPU_DEPTH.with(|depth| depth.set(self.depth));
+    }
+}
+
+impl Drop for BackgroundCpuScopeV1 {
+    fn drop(&mut self) {
+        BACKGROUND_CPU_DEPTH.with(|depth| {
+            let remaining = depth.get().saturating_sub(1);
+            depth.set(remaining);
+            if remaining == 0 {
+                BACKGROUND_CPU_UNITS.with(|units| units.set(0));
+            }
+        });
+    }
+}
+
+impl ProcessBackgroundCpuV1 {
+    fn new(width: NonZeroUsize) -> Self {
+        Self {
+            width,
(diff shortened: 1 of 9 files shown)


ScriptedAlchemy:
The three pre-existing base compile errors you hit and correctly declined to repair — the `BatchWorkerStopped` match arm in `runtime/ingest/failure.rs:517` and the missing `HostAdmissionOutcome.recovery` field in `kiro.rs`, `cline_like.rs` and `ingest/tests.rs` ×2 — are now fixed in **#681**, along with two more of the same class (`background_cpu.rs` and the four missing `parallelism.rs` items that `code_index_scheduler.rs` references).

All five are one root cause: `perf(observation): batch admission and preparation` landed changes whose consumers, and one whole file, were never `git add`ed.

**Merge #681 before this PR** and CI here should go green. This PR carries the #676 cherry-pick, which #681 supersedes; git will drop it as empty.

ScriptedAlchemy:

PR #683, opened 2026-08-24
perf(ledger): store commit sequence, not the receipt

ScriptedAlchemy:
## What

`td_runtime_writer_idempotency_v1` replaces its 587-byte `original_receipt_json`
column with a `commit_sequence` integer, and gains a transactional in-place
upgrade so existing stores move to the narrow shape.

## Why the receipt is reconstructible

This is the load-bearing claim, so it is argued twice — structurally and
empirically.

**Structurally.** `StoreCommitReceiptV1` is `#[serde(deny_unknown_fields)]` over
exactly seven fields. Six were already columns of this same row:

| receipt field | column |
| --- | --- |
| `operation_id` | `operation_id` |
| `idempotency.key` | `idempotency_key` |
| `idempotency.command_digest` | `request_digest` |
| `shard_id` | `shard_json` |
| `incarnation` | `incarnation` |
| `authority_epoch` | `authority_epoch` |
| `committed_at` | `committed_at_micros` |
| `commit_sequence` | **nothing** |

`decode_row` already *required* every one of those equalities and answered
`LedgerError::Corrupt` otherwise, so the encoded receipt was never authority for
anything except `commit_sequence`. The reconstruction is that same check run
backwards. Four of the six — shard, incarnation, authority epoch, idempotency
key — are the primary key the row was matched on, so they are the caller's own
arguments by construction.

**Empirically.** Every real row on this machine was rebuilt byte-for-byte from
the six columns plus the sequence:

```
proj_099ef1e4af1c6508/sessions.db: rows=24558 reconstruct_mismatches=0 unexpected_key_shape=0
proj_7b2798f1d4448b5d/sessions.db: rows=40003 reconstruct_mismatches=0 unexpected_key_shape=0
proj_a5b3d7e3ebe14ca7/sessions.db: rows=94313 reconstruct_mismatches=0 unexpected_key_shape=0
.tracedecay/user-sessions.db:      rows=65634 reconstruct_mismatches=0 unexpected_key_shape=0

TOTAL rows scanned:            224508
byte-exact reconstruct FAILED: 0
receipts with unexpected keys: 0
RESULT: commit_sequence is the ONLY field not already a column.
```

## Why the row was spilling

`maxLocal` for an index b-tree at 4 KiB pages is `((4096-12)*64/255)-23 = 1002`
bytes. Measured payloads, and the projection after the change:

```
BEFORE (payload incl. record header)          AFTER (projected)
proj_099ef1e4af1c6508  avg=1486 max=1493      min=895 avg=899 max=902  over_maxlocal=0
proj_7b2798f1d4448b5d  avg=1478 max=1493      min=895 avg=895 max=902  over_maxlocal=0
proj_a5b3d7e3ebe14ca7  avg=1485 max=1493      min=895 avg=898 max=902  over_maxlocal=0
user-sessions          avg=1372 max=1382      min=821 avg=823 max=828  over_maxlocal=0
```

Every one of the 224,508 rows was over `maxLocal` before and none is after. The
`dbstat` page census showed overflow pages exactly 1:1 with rows in all four
stores (24,558 / 40,003 / 94,313 / 65,634), each holding ~1,005 bytes in a
4,096-byte page — 688 MB of pure slack.

Note the remaining headroom is ~100 bytes, and the largest remaining column is
`transaction_scope_json` at 456 bytes, which re-encodes the binding and
durability this row already carries. Normalising it (and `shard_json`) is the
obvious follow-up, but it is a separate change and this one has to land first.

## Why the migration is safe

The ledger shards sit at `user_version = 0`, outside the schema framework, and
share their file with ~207 unrelated tables. Claiming that file-global pragma
for the ledger would let a stamp disagree with the shape it describes, so
**the shape is the version**: the presence of `original_receipt_json` is the old
one, read from `pragma_table_info`, which is derived from the same schema record
the queries compile against and therefore cannot drift from what is there.

- **Interrupted migration.** The rewrite runs entirely inside the writer's
  existing transaction, and SQLite DDL is transactional. A crash rolls back to
  the old table *including its column set*, so the next open re-detects the old
  shape and retries. There is no state outside the transaction that could
  disagree with it, so there is no half-migrated state to detect.
- **Concurrent access.** One writer actor owns one shard, and SQLite serializes
  writers across processes. A racing process either waits for the exclusive lock
  and observes the new shape, or is refused `SQLITE_BUSY` having changed nothing.
- **Older binary afterwards.** Its `SELECT` names `original_receipt_json`, which
  no longer exists, so the statement fails to *prepare*. The direction matters:
  it is refused, never told "no such row", so it cannot mistake an unreadable
  shape for an absent record and admit a duplicate. Pinned by
  `an_older_binary_is_refused_by_a_migrated_store_rather_than_misreading_it`.
  This is forward-only, consistent with the framework's own stance that a store
  at an unexpected shape "is refused at open".
- **Unmigratable row.** A commit sequence that is not a positive integer aborts
  the migration with the same `Corrupt` class that reading that row already
  raised, leaving the store byte-identical at the old shape. `typeof` is checked
  explicitly because SQLite orders text above every integer, so a bare `> 0`
  would have let `"not-a-number"` through. Dropping such a row instead would
  silently shrink the set of submissions the ledger recognises — the one outcome
  that could admit a duplicate write.
- **Row count is verified** after the copy; a short copy fails closed.
- No view, trigger, or index depends on the table (checked against a real
  store), so the `DROP`/`RENAME` swap has no dependents.

## Measured, on a copy of a real 94,366-row store

The exact statements the migration runs, applied to a copy of a live store:

```
--- before ---
  rows=94366
  ledger_bytes=442458112 overflow_pages=94366
  db_bytes=1604157440
--- migrating ---
  unmigratable_rows=0
--- after ---
  rows=94366
  ledger_bytes=96649216 overflow_pages=0
  db_bytes=1239306240

--- receipt equivalence over the whole real store ---
  expected receipts: 94366
  rebuilt receipts:  94366
  RESULT: every receipt round-trips byte-for-byte through the migration.
```

Ledger 442,458,112 → 96,649,216 bytes (**78.2%**, 345,808,896 saved), overflow
pages 94,366 → 0, whole store down 22.7%, every receipt unchanged.

## RED / GREEN

RED, with the migration neutered to an early `Ok(())` (the "row-width fix with
no migration mechanism" state):

```
running 6 tests
test ledger::migrate::tests::an_unmigratable_receipt_fails_closed_without_dropping_the_row ... FAILED
test ledger::migrate::tests::an_interrupted_migration_leaves_the_legacy_store_untouched_and_retries ... FAILED
test ledger::migrate::tests::a_migrated_duplicate_still_replays_instead_of_committing_again ... FAILED
test ledger::migrate::tests::migrating_an_already_narrow_store_is_a_no_op ... ok
test ledger::migrate::tests::migration_preserves_the_exact_receipt_a_legacy_store_recorded ... FAILED
test ledger::migrate::tests::the_narrow_shape_removes_the_per_row_overflow_page ... ok

---- a_migrated_duplicate_still_replays_instead_of_committing_again stdout ----
called `Result::unwrap()` on an `Err` value: Sqlite(SqlInputError { ...
  msg: "no such column: commit_sequence", ... })

---- migration_preserves_the_exact_receipt_a_legacy_store_recorded stdout ----
assertion failed: !has_column(&connection, "original_receipt_json")

---- an_interrupted_migration_leaves_the_legacy_store_untouched_and_retries stdout ----
assertion failed: !has_column(&transaction, "original_receipt_json")

---- an_unmigratable_receipt_fails_closed_without_dropping_the_row stdout ----
assertion failed: matches!(initialize_schema(&transaction), Err(LedgerError::Corrupt { .. }))

test result: FAILED. 2 passed; 4 failed; 0 ignored; 0 measured; 299 filtered out
```

That third failure is itself the shape-mismatch evidence: a mismatch is a hard
SQL refusal, never a silent miss.

GREEN:

```
running 12 tests
test repository::fact::tests::fact_executor_does_not_claim_replay_without_writer_ledger ... ok
test ledger::migrate::tests::migrating_an_already_narrow_store_is_a_no_op ... ok
test ledger::migrate::tests::an_unmigratable_receipt_fails_closed_without_dropping_the_row ... ok
test ledger::migrate::tests::an_older_binary_is_refused_by_a_migrated_store_rather_than_misreading_it ... ok
test ledger::migrate::tests::migration_preserves_the_exact_receipt_a_legacy_store_recorded ... ok
test ledger::migrate::tests::a_migrated_duplicate_still_replays_instead_of_committing_again ... ok
test ledger::migrate::tests::an_interrupted_migration_leaves_the_legacy_store_untouched_and_retries ... ok
rows=500 legacy_bytes=2314240 narrow_bytes=466944 saved=1847296 reduction=79%
test ledger::migrate::tests::the_narrow_shape_removes_the_per_row_overflow_page ... ok

test result: ok. 12 passed; 0 failed; 0 ignored; 0 measured; 294 filtered out; finished in 0.12s
```

The migration tests start from `LEGACY_IDEMPOTENCY_DDL`, the pre-change table
definition reproduced verbatim, with receipts encoded by the same serializer
production used — a real pre-migration shape, not something the new code could
have written.

Duplicate-admission behaviour is covered directly:
`a_migrated_duplicate_still_replays_instead_of_committing_again` seeds a legacy
row, migrates, then asserts a same-key/same-digest retry still `Replay`s the
original receipt, a same-key/different-digest submission still `Conflict`s, and
that neither added a row.

Full crate suite: 440 tests across 20 binaries, all passing. No assertion
anywhere is on elapsed time.

## Checks

```
cargo check -p tracedecay-rusqlite-runtime --all-targets --locked   # clean
cargo clippy -p tracedecay-rusqlite-runtime --all-targets --locked  # no warnings
cargo test -p tracedecay-rusqlite-runtime --locked                  # all pass
```

Per-crate only, as requested.

## Relationship to the pruning rule

`perf(ledger): prune superseded idempotency records` is not on this base branch.
The two are complementary and near-disjoint — prune deletes unreachable rows,
this narrows the rows that remain. On merge the only textual overlaps are two
adjacent lines in `commit.rs` (prune inserts a `prune_superseded` call directly
above the `idempotency::insert` call whose `receipt_json` argument this removes)
and separate test functions in `tests.rs`; the migration tests were deliberately
placed in `migrate.rs` to keep `tests.rs` conflict-free. Nothing in this change
touches `prune.rs`'s `DELETE`, which selects on `shard_json` and
`authority_epoch` only.

## Scope deviation

- `tests.rs::malformed_canonical_json_fails_closed` and
  `tests/runtime_actor/faults.rs` both corrupted `original_receipt_json` to
  assert fail-closed behaviour. That column no longer exists, so both were
  retargeted to `transaction_scope_json`, which is still an encoded column, to
  preserve each test's intent.
- `commit.rs` no longer encodes the receipt to JSON, so its now-unused
  `encode_json` import was dropped.
- Not done, deliberately: `transaction_scope_json` and `shard_json`
  normalisation (the further ~20%), and any change to
  `td_runtime_writer_checkpoint_v1`, which holds one row per shard/incarnation
  and is not a size problem.

### DIFF
--- a/crates/tracedecay-rusqlite-runtime/src/ledger.rs
+++ b/crates/tracedecay-rusqlite-runtime/src/ledger.rs
@@ -8,6 +8,7 @@ mod commit;
 mod error;
 mod idempotency;
 mod inbox;
+mod migrate;
 mod outbox;
 mod schema;
 mod sqlite;
--- a/crates/tracedecay-rusqlite-runtime/src/ledger/commit.rs
+++ b/crates/tracedecay-rusqlite-runtime/src/ledger/commit.rs
@@ -6,7 +6,7 @@ use tracedecay_store::{
 
 use super::{
     LedgerDisposition, LedgerError, checkpoint, idempotency, inbox, outbox,
-    sqlite::{LedgerTransaction, Submission, encode_json},
+    sqlite::{LedgerTransaction, Submission},
 };
 
 enum RuntimeBookkeeping<'a> {
@@ -73,9 +73,8 @@ fn record_with_bookkeeping(
         commit_sequence: checkpoint.watermark.commit_sequence,
         committed_at: metadata.admitted_at,
     };
-    let receipt_json = encode_json(&receipt, "original_receipt_json")?;
     checkpoint::persist(transaction, &submission, &checkpoint, &receipt)?;
-    idempotency::insert(transaction, &submission, &receipt, &receipt_json)?;
+    idempotency::insert(transaction, &submission, &receipt)?;
     match bookkeeping {
         RuntimeBookkeeping::None => {}
         RuntimeBookkeeping::Outbox(entry) => {
--- a/crates/tracedecay-rusqlite-runtime/src/ledger/idempotency.rs
+++ b/crates/tracedecay-rusqlite-runtime/src/ledger/idempotency.rs
@@ -1,7 +1,9 @@
 use rusqlite::{Row, params};
+use tracedecay_domain::UtcMicros;
 use tracedecay_store::{
-    CommandDigestV1, DurabilityClassV1, IdempotencyIdentityV1, RuntimeTransactionScopeV1,
-    StoreCommitReceiptV1, StoreIdempotencyKeyV1, StoreOperationIdV1, StoreRuntimeBindingV1,
+    CommandDigestV1, CommitSequenceV1, DurabilityClassV1, IdempotencyIdentityV1,
+    RuntimeTransactionScopeV1, StoreCommitReceiptV1, StoreIdempotencyKeyV1, StoreOperationIdV1,
+    StoreRuntimeBindingV1,
 };
 
 use super::{
@@ -11,7 +13,7 @@ use super::{
 
 const IDEMPOTENCY_TABLE: &str = "td_runtime_writer_idempotency_v1";
 const SELECT_IDEMPOTENCY: &str = r#"
-SELECT request_digest, original_receipt_json, transaction_scope_json,
+SELECT request_digest, commit_sequence, transaction_scope_json,
        operation_id, durability_json, committed_at_micros
 FROM td_runtime_writer_idempotency_v1
 WHERE shard_json = ?1 AND incarnation = ?2 AND authority_epoch = ?3
@@ -20,7 +22,7 @@ WHERE shard_json = ?1 AND incarnation = ?2 AND authority_epoch = ?3
 const INSERT_IDEMPOTENCY: &str = r#"
 INSERT OR IGNORE INTO td_runtime_writer_idempotency_v1 (
     shard_json, incarnation, authority_epoch, idempotency_key, request_digest,
-    original_receipt_json, transaction_scope_json, operation_id, durability_json,
+    commit_sequence, transaction_scope_json, operation_id, durability_json,
     committed_at_micros
 ) VALUES (?1, ?2, ?3, ?4, ?5, ?6, ?7, ?8, ?9, ?10)
 "#;
@@ -83,8 +85,8 @@ pub(super) fn insert(
     transaction: &impl LedgerTransaction,
     submission: &Submission<'_>,
     receipt: &StoreCommitReceiptV1,
-    receipt_json: &str,
 ) -> Result<(), LedgerError> {
+    let commit_sequence = sqlite_u64(receipt.commit_sequence.0, "commit sequence")?;
     let changed = transaction.execute(
         INSERT_IDEMPOTENCY,
         params![
@@ -93,7 +95,7 @@ pub(super) fn insert(
             submission.authority_epoch_sql,
             submission.metadata.idempotency.key.as_str(),
             submission.metadata.idempotency.command_digest.as_str(),
-            receipt_json,
+            commit_sequence,
             &submission.transaction_scope_json,
             submission.metadata.operation_id.as_str(),
             &submission.durability_json,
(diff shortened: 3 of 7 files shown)


ScriptedAlchemy:
Integrated directly into PR #421 as f55a1d62e, resolving the ledger conflict by preserving the already-landed bounded 256-row superseded-record pruning while adopting the narrower commit-sequence representation.

ScriptedAlchemy:
Post-merge review found two blockers in the integrated migration: it copies the full legacy table on the first foreground writer transaction, defeating bounded pruning, and it drops original_receipt_json after validating only commit_sequence rather than all legacy receipt invariants. The integration was reverted from PR #421 in the follow-up revert commit; this PR remains closed pending a bounded/background migration with full corruption validation.

PR #684, opened 2026-08-24
fix(cli): honest dashboard reuse reporting + init -p/--path

ScriptedAlchemy:
## Summary

Fixes two CLI defects on `tracedecay-cli`.

### Defect 1 — `tracedecay dashboard --port <N>` (root cause + fix)

**Root cause, established by reading the code, not assumed:** the
`tracedecay_dashboard` MCP tool runs *at most one dashboard per daemon
process* (`src/mcp/tools/handlers/dashboard.rs`, `DASHBOARD_MANAGER`, a
process-wide `OnceLock<Mutex<Option<RunningDashboard>>>`). This reuse is
**documented and intentional** ("Idempotent: returns the existing URL if
already running for this process"), not a bug. The CLI's daemon is a single
long-lived process shared across every `tracedecay dashboard` invocation for
the profile, so a second `dashboard --port 7400 --path mold` call finds the
first dashboard (react-router, port 7399) still up and reuses it.

**What was genuinely wrong:** the CLI ignored the daemon's `status` field
and always printed as if the request had been honored — `"Serving project
{requested_path}"` was printed even when the reachable server was actually
serving a *different* project, and the daemon never even reported which
project the reused instance was actually serving, so there was no way for
the CLI to detect or report the mismatch. The requested `--port` was
silently discarded with no warning and exit code 0.

**Fix:**
- `src/mcp/tools/handlers/dashboard.rs`: `RunningDashboard` now records
  `host`, `port`, and the canonicalized `project_root` it is actually
  serving. The `start` reuse path (`already_running`/`stopping`) now returns
  those actual values plus a `matches_request` bool (computed by comparing
  the running instance's host/port/project against what was just requested).
- `crates/tracedecay-cli/src/main.rs`: `tracedecay dashboard` now branches
  on the daemon's response:
  - `started` → unchanged success path.
  - already running **and** it happens to match the request exactly → a
    truthful no-op ("dashboard already listening on ...").
  - already running (or shutting down) **and it does not match** → the CLI
    now **errors** (`process::exit(1)`) with a message naming the actual
    project/URL being served and stating plainly that the requested
    host/port/project were **not** started, with a working next step
    (`tracedecay tool tracedecay_dashboard --args '{"action":"stop"}'`).

I did not turn the process-wide single-dashboard-per-daemon design into a
multi-dashboard registry (keyed by project/port) — the code already has a
dedicated single-instance shutdown/task-completion lifecycle
(`shutdown_dashboard_until`, `DASHBOARD_MANAGER`) that a multi-instance
registry would need to redesign, which is well past "a small change" and
risks colliding with concurrent hot-path work already in flight on this
branch. Erroring loudly and truthfully on a genuine conflict, while keeping
matching requests a real no-op, satisfies "never told a falsehood" without
that redesign.

`--port 0` ("pick a free port") is untouched for the fresh-start path; it
was already handled correctly by `bind_dashboard`, and continues to be.

### Defect 2 — `tracedecay init -p <path>`

`Init`'s `path` field was a *bare positional* (`path: Option<String>`, no
`#[arg(...)]`), unlike `dashboard`/`gitignore`/`bench`, which declare
`#[arg(short, long)] path: Option<String>` and so accept `-p`/`--path`. Top
level `--help` documents `--path` as one of the general ways to target
another project, so this was a real, if small, inconsistency.

Resolution: added `-p`/`--path` as an explicit flag (`path_flag`) alongside
the existing positional `path`, `conflicts_with` the positional so passing
both is a parse error rather than a silent pick. `tracedecay init /repo`
keeps working exactly as documented in `INIT_AFTER_HELP`.

## Real defect vs. intended behavior

| Symptom | Verdict |
|---|---|
| Second `dashboard` call reuses the first daemon-wide instance instead of binding a new port | **Intended** (single dashboard per daemon process, documented) |
| CLI prints "Serving project X" for a request that was never honored | **Defect** — fixed |
| Explicit `--port` silently discarded with no warning/error | **Defect** — fixed (now a loud, truthful error when it actually conflicts) |
| `tracedecay init -p <path>` rejected | **Defect** — fixed |

## Tests (RED/GREEN)

- `crates/tracedecay-cli/src/dashboard_start_tests.rs` (new): fabricates the
  exact `tracedecay_dashboard` tool response from the reported repro
  (react-router already running on 7399, `mold --port 7400` requested) and
  asserts the CLI's pure response-interpretation function
  (`interpret_dashboard_start_response`) returns a `Conflict` naming the
  *actual* project/port and saying the requested ones were not started —
  never a silent `Started`. Also covers the exactly-matching no-op case, the
  `stopping` status, and the missing-URL-is-an-error case.
- `crates/tracedecay-cli/src/cli/parse_tests.rs`
  (`init_accepts_short_and_long_path_flag_like_dashboard_does`): asserts
  `-p`/`--path` and the positional PATH all parse, and that supplying both
  together is a parse-time `ArgumentConflict`, not a silent pick.

GREEN, pasted verbatim (`cargo test -p tracedecay-cli --locked` filtered to
the new tests, both against the fixed code):

```
running 5 tests
test dashboard_start_tests::a_response_missing_url_is_an_error_not_a_silent_success ... ok
test dashboard_start_tests::already_running_response_for_a_different_project_is_reported_as_a_conflict_not_a_falsehood ... ok
test dashboard_start_tests::already_running_response_matching_the_request_is_a_truthful_no_op ... ok
test dashboard_start_tests::started_response_is_reported_as_started ... ok
test dashboard_start_tests::stopping_status_mismatched_with_the_request_is_also_a_conflict ... ok
test result: ok. 5 passed; 0 failed; 0 ignored; 0 measured; 294 filtered out

test cli::parse_tests::init_accepts_short_and_long_path_flag_like_dashboard_does ... ok
test result: ok. 1 passed; 0 failed; 0 ignored; 0 measured; 298 filtered out
```

RED (the exact scenario reconstructed as evidence, not a hypothetical):
`already_running_response_for_a_different_project_is_reported_as_a_conflict_not_a_falsehood`
fabricates the daemon's real pre-fix response shape — `status:
"already_running"`, `url: "http://127.0.0.1:7399/"` (react-router, already
running) — for a request targeting mold on port 7400. Before this fix the
CLI had no `matches_request`/`project_root` fields to check and no
`Conflict` outcome at all: it unconditionally printed
`"tracedecay dashboard listening on {url}"` (the real, reachable URL) and
`"Serving project {requested_path}"` (false — react-router, not mold, was
being served) and returned `Ok(())` — exit 0. The test's assertions
(`message.contains("react-router")`, `.contains("7399")`,
`.contains("7400") && .contains("mold")`,
`.contains("NOT started") || .contains("were not")`) are only satisfiable
by the fixed `Conflict` path; the pre-fix code path had no `message` to
even inspect.

Full `cargo test -p tracedecay-cli --locked --bin tracedecay` on the fixed
tree: **291 passed** (including the 6 new tests above), **8 failed** — all
in `agent_cmd::tests::{codex,opencode,kimi}_*`, `tool_command::tests::
successful_tool_result_exits_zero`, and `startup_tests::
agent_install_health_check_is_selective`. None touch dashboard or init;
they reference `host_component_registration.rs` filesystem/receipt
plumbing for Codex/OpenCode/Kimi lifecycle, a subsystem this PR never
touches. Re-running the exact same command against a stash of every
defect-fix file (keeping only the three build-unblocker commits) reproduces
the identical 291/8 split with the identical 8 test names — proving these
failures pre-exist on this branch and are unaffected by this change.

## Scope deviation

This branch (`cursor/simplify-pr421-hot-paths`) does not compile at the
commit I branched from — confirmed by reproducing every failure below on a
clean `git stash` of my diff, i.e. none of it is caused by my change:

1. `tracedecay-sessions`: unresolved `tracedecay_runtime_core::background_cpu`
   import. Already fixed, unmodified, on sibling branch
   `claude/perf-graph-publication-batching` (commit `1d591191e`) — cherry-picked.
2. `tracedecay-sessions`: non-exhaustive match on
   `ObservationApplicationError::BatchWorkerStopped`. Already fixed,
   unmodified, on sibling branch `local/verify-stack` (commit `bfba3221e`) —
   cherry-picked.
3. Root `tracedecay` crate: two `HostAdmissionOutcome { .. }` struct
   literals (`src/host_admission.rs`,
   `src/host_admission/integration_test_support.rs`) were missing the
   `recovery` field added to the struct by an in-flight refactor. No existing
   fix found on any branch, so I added `recovery: None` at both sites myself
   (the same value every other constructor in the codebase uses for a
   non-recovery outcome).

None of these three touch dashboard or init behavior; they exist solely to
make this branch — and this PR's own diff — actually compile so its tests
can run. I made no other changes outside the two assigned CLI defects and
these three minimal, additive build-unblockers.

### DIFF
--- a/crates/tracedecay-cli/src/cli.rs
+++ b/crates/tracedecay-cli/src/cli.rs
@@ -191,6 +191,13 @@ pub enum Commands {
     Init {
         /// Project path (default: current directory)
         path: Option<String>,
+        /// Project path, as an explicit flag. Equivalent to the positional
+        /// PATH argument above; accepted for consistency with `-p`/`--path`
+        /// on other project-scoped commands (e.g. `dashboard`, `gitignore`,
+        /// `bench`). Conflicts with the positional PATH — pass one or the
+        /// other, not both.
+        #[arg(short = 'p', long = "path", value_name = "PATH", conflicts_with = "path")]
+        path_flag: Option<String>,
         /// Folders to skip during indexing (can be repeated)
         #[arg(long = "skip-folder", num_args = 1..)]
         skip_folders: Vec<String>,
--- a/crates/tracedecay-cli/src/cli/parse_tests.rs
+++ b/crates/tracedecay-cli/src/cli/parse_tests.rs
@@ -716,6 +716,62 @@ fn branch_autotrack_enable_parses_poll_secs_and_path() {
 }
 
 #[test]
+fn init_accepts_short_and_long_path_flag_like_dashboard_does() {
+    // `-p, --path` is documented (see TOP_LEVEL_AFTER_HELP) and already works
+    // on `dashboard`, `gitignore`, and `bench`; `init` previously only took
+    // PATH positionally and rejected `-p`/`--path` outright.
+    let short = Cli::try_parse_from(["tracedecay", "init", "-p", "/tmp/project"])
+        .expect("init -p PATH should parse");
+    assert!(matches!(
+        short.command,
+        Some(Commands::Init {
+            path: None,
+            path_flag,
+            ..
+        }) if path_flag.as_deref() == Some("/tmp/project")
+    ));
+
+    let long = Cli::try_parse_from(["tracedecay", "init", "--path", "/tmp/project"])
+        .expect("init --path PATH should parse");
+    assert!(matches!(
+        long.command,
+        Some(Commands::Init {
+            path: None,
+            path_flag,
+            ..
+        }) if path_flag.as_deref() == Some("/tmp/project")
+    ));
+
+    // The positional form keeps working unchanged.
+    let positional = Cli::try_parse_from(["tracedecay", "init", "/tmp/project"])
+        .expect("init PATH should still parse positionally");
+    assert!(matches!(
+        positional.command,
+        Some(Commands::Init {
+            path,
+            path_flag: None,
+            ..
+        }) if path.as_deref() == Some("/tmp/project")
+    ));
+
+    // Supplying both the positional PATH and `-p`/`--path` is refused rather
+    // than silently picking one — clap's `conflicts_with` rejects it.
+    // `Cli` does not derive `Debug`, so match directly instead of
+    // `.expect_err(...)` (which requires the `Ok` type to be `Debug`).
+    let conflict = match Cli::try_parse_from([
+        "tracedecay",
+        "init",
+        "/tmp/project",
+        "--path",
+        "/tmp/other",
+    ]) {
+        Ok(_) => panic!("init PATH and --path together should be rejected as a conflict"),
+        Err(error) => error,
+    };
+    assert_eq!(conflict.kind(), ErrorKind::ArgumentConflict);
+}
+
+#[test]
 fn init_and_sync_parse_runtime_skip_and_include_folders() {
     let init = Cli::try_parse_from([
         "tracedecay",
@@ -732,6 +788,7 @@ fn init_and_sync_parse_runtime_skip_and_include_folders() {
         init.command,
         Some(Commands::Init {
             path,
+            path_flag: None,
             skip_folders,
             include_folders,
             adopt_project: None,
--- a/crates/tracedecay-cli/src/dashboard_start_tests.rs
+++ b/crates/tracedecay-cli/src/dashboard_start_tests.rs
@@ -1,0 +1,158 @@
+//! RED/GREEN coverage for `tracedecay dashboard`'s handling of the daemon's
+//! process-wide dashboard singleton (see `src/mcp/tools/handlers/dashboard.rs`).
+//!
+//! These fabricate the `tracedecay_dashboard` tool response directly instead
+//! of spinning up a real daemon, so the assertions are on the reported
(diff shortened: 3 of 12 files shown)


ScriptedAlchemy:
The pre-existing base breaks you had to work around — the `background_cpu` import, the non-exhaustive `ObservationApplicationError` match, and the two `HostAdmissionOutcome` literals missing `recovery` — are all fixed in **#681**, together with a fourth of the same class (four `parallelism.rs` items that `src/daemon/code_index_scheduler.rs` references but which were never `git add`ed).

All of them trace to one root cause: `perf(observation): batch admission and preparation` landed changes whose consumers, and one whole file, were never staged.

**Merge #681 first**, then drop the carried base fixes from this branch so the diff is only the dashboard and init work. Your `recovery: None` choice matches what #681 does, so there should be no semantic conflict.

ScriptedAlchemy:
The CLI init -p/--path fix was integrated directly into PR #421 as 461b16cf8. The dashboard singleton patch was not copied because #421 already carries the stronger per-canonical-project manager and truthful actual-port reuse reporting.

PR #685, opened 2026-08-24
perf(semantic): reserve per-session bytes, not process budget

ScriptedAlchemy:
## The infer-vs-wait split: 100% infer, 0% wait — by construction, not by measurement

**`semantic.session_pool.wait` can never fire for embedding.** Its only
`measure_block!` sits inside `SessionPool::acquire_blocking`, and every call
site of `acquire_blocking` is inside `session_pool.rs`'s own `#[cfg(test)]`
module. The projection path acquires through `RuntimeChunkVectorEncoderV1::
ensure_sessions` → `SemanticRuntimeService::acquire` → the **non-blocking**
`SessionPool::acquire`. No embedding caller ever waits.

So the split could not have been read off the existing labels. Pool pressure
reaches this path only through `ensure_sessions`' `Err(_) => break` arm, which
silently narrows the width and, until this PR, incremented nothing.

Everything below follows from that.

## What the width actually is

The premise I was given was `96 / 4 = 24 → capped to 16 → ~64 cores busy`.
That is wrong, and the host disagrees.

`embedding_session_width_for` divides `indexing_worker_target(total_cores)`,
not `total_cores`. Measured live on this 96-core host via the hotpath MCP
`gauges` surface:

```
code_index_worker_count = 8   (last_value, 1340 logs)
```

`indexing_worker_target(96)` = 8 (the parser-memory ceiling), so
`embedding_session_width(4, …)` = `8 / 4` = **2 sessions**, not 16 — a
derived ceiling of ~8 cores, not ~64. The crate's own test already asserts
this (`embedding_session_width_for(96, 4, 64) == 2`).

## The bug: one session reserves the entire process budget

`VerifiedEmbeddingArtifactV1` set `resident_byte_ceiling = resources.
max_resident_bytes` (2 GiB, the *process* semantic budget) at both production
construction sites, and both `resident_bytes_reservation` impls returned that
ceiling as the *per-session* reservation.

In `acquire_verified` the pool admits while
`resident_bytes + reserved <= memory_ceiling_bytes`:

| session | resident before | reserved | projected | admitted |
|---|---|---|---|---|
| 1 | 0 | 2 GiB | 2 GiB | yes |
| 2 | 2 GiB | 2 GiB | 4 GiB | **MemoryCeilingExceeded** |

`ensure_sessions(2)` therefore returned 1, hit `Err(_) => break`, and
embedding ran at **1 session × 4 intra-op threads ≈ 4 of 96 cores** — half
even the derived width, on every host, regardless of configuration. The rayon
striping in `encode_batches` is unreachable in production, because
`sessions <= 1` always takes the sequential branch.

## Falsifiable RED/GREEN evidence (counters, not elapsed time)

`production_scale_artifact_admits_the_derived_session_width` drives the real
`AdmittedProjectionArtifactV1::admit` path with a production-scale artifact
(612 MiB model + 2 MiB tokenizer under a 2 GiB ceiling) and counts how many
sessions the pool's own admission arithmetic accepts.

**RED** — reservation restored to the pre-fix `resident_byte_ceiling()`:

```
panicked at fastembed_adapter.rs:1750:
the process ceiling must admit at least the two sessions the host width
arithmetic derives, but only 1 fit at 2147483648 bytes each
test result: FAILED. 0 passed; 1 failed
```

**GREEN** — per-session estimate:

```
test production_scale_artifact_admits_the_derived_session_width ... ok
test resident_estimate_admits_more_than_one_session_under_the_process_ceiling ... ok
test resident_estimate_never_exceeds_the_ceiling_and_survives_unknown_sizes ... ok
test result: ok. 170 passed; 0 failed   (full tracedecay-semantic lib suite)
```

The failure message *is* the measurement: `1 fit at 2147483648 bytes each`.

## The fix, and why it does not exceed the memory budget

The reservation is now derived from the artifact's own declared member
lengths (`model + tokenizer`, ×5/4 headroom for ORT arena and activations),
clamped to `resident_byte_ceiling`, which the pool still enforces unchanged.
An artifact that declares no lengths falls back to the ceiling, preserving the
previous conservative behaviour exactly.

At production scale that is ~768 MiB per session, so the unchanged 2 GiB
budget admits **2** sessions — exactly the width the CPU arithmetic already
derived, and within the pool's `ResidentSessionLimit` (`max_sessions =
embedding_pool_sessions` = width + 1 = 3). **No ceiling, no width formula, and
no concurrency knob was raised.** The 96-core machine still uses ~8 cores for
embedding, not 64; that ceiling is memory, and it is respected.

## Instrumentation added for the genuinely unmeasured steps

- `semantic_embed_session_width` / `_sessions_wanted` / `_sessions_held` /
  `_session_shortfall` — the shortfall counter is the only signal that a run
  narrowed, since no wait can occur.
- `semantic_embed_encode_stripes` / `_sequential_dispatch` — dispatch shape
  actually taken.
- `semantic.embed.batch_assembly` — the one copy from canonical chunks into
  owned tensor input.
- `semantic.embed.vector_writeback` — per-vector validation and move out of
  the adapter type.

Tokenisation is not separately instrumented: FastEmbed's `embed()` tokenises
and infers in one call, so splitting it would mean changing how the model is
invoked. Model *fetch* was out of scope and is untouched.

## Batch size: measured, and deliberately not changed

`inference_batch_size` = `max_batch_size` = 32; projection pages are 512
chunks (`MAX_SEMANTIC_VECTOR_STAGE_CHUNKS_PER_BATCH`). Groups reach 32 only
when singleton-file runs coalesce; a multi-chunk file flushes the singleton
run and emits its own group, so real corpora produce many sub-32 batches.

That is a real inefficiency, but **it is not a free lever**: `projector.rs`
states group membership *is* projection identity — regrouping changes tensor
shapes, therefore vector bytes, therefore every `output_digest` and the
generation manifest digest. Enlarging batches means a full re-embed, which is
a separate decision. Reporting it, not taking it.

## Verification

```
cargo check -p tracedecay-semantic --all-targets --locked
cargo check -p tracedecay-semantic --all-targets --locked --features hotpath
cargo check -p tracedecay-semantic --all-targets --locked --features semantic-fastembed,hotpath
cargo test  -p tracedecay-semantic --lib      # 170 passed
```

No workspace build, no workspace clippy. Sizing only: vector bytes and
generation digests are unchanged, which the crate's existing width-equivalence
tests continue to cover.

### DIFF
--- a/crates/tracedecay-semantic/src/fastembed_adapter.rs
+++ b/crates/tracedecay-semantic/src/fastembed_adapter.rs
@@ -176,7 +176,40 @@ struct VerifiedEmbeddingArtifactV1 {
     max_batch_bytes: u32,
     max_threads: u32,
     resident_byte_ceiling: u64,
+    /// What ONE warmed session is expected to retain, derived from the
+    /// artifact's own declared member lengths.
+    ///
+    /// This is deliberately NOT `resident_byte_ceiling`. The ceiling is the
+    /// whole *process* semantic budget; charging it per session makes the
+    /// first session reserve the entire budget, so the pool's memory check
+    /// refuses every subsequent session and embedding silently collapses to
+    /// one session on every host. The reservation must therefore be a
+    /// per-session estimate, bounded above by the ceiling.
+    resident_bytes_estimate: u64,
     load_deadline_ms: u64,
+}
+
+/// Headroom over the artifact's declared member bytes for ORT arena,
+/// activations, and allocator slack.
+///
+/// Deliberately generous: under-reserving would let the pool admit sessions
+/// the machine cannot hold, which is a worse failure than embedding narrow.
+const RESIDENT_ESTIMATE_HEADROOM_NUMERATOR: u64 = 5;
+const RESIDENT_ESTIMATE_HEADROOM_DENOMINATOR: u64 = 4;
+
+/// Per-session resident estimate from declared member lengths, clamped into
+/// `1..=ceiling`. A zero or unknown length falls back to the ceiling, which
+/// preserves exactly the previous conservative behaviour for any artifact
+/// that does not declare its sizes.
+fn resident_bytes_estimate_for(member_bytes: u64, resident_byte_ceiling: u64) -> u64 {
+    if member_bytes == 0 {
+        return resident_byte_ceiling;
+    }
+    member_bytes
+        .saturating_mul(RESIDENT_ESTIMATE_HEADROOM_NUMERATOR)
+        .checked_div(RESIDENT_ESTIMATE_HEADROOM_DENOMINATOR)
+        .unwrap_or(resident_byte_ceiling)
+        .clamp(1, resident_byte_ceiling.max(1))
 }
 
 #[derive(Clone, Debug, PartialEq, Eq)]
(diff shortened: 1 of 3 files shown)


ScriptedAlchemy:
Integrated directly into PR #421 as aba1ad7c2, preserving the installed-width CPU permit authority while adding per-session resident-byte accounting.

PR #686, opened 2026-08-24
feat(github): cut GitHub API cost across three read paths

ScriptedAlchemy:
Three GitHub-client efficiency items that share `crates/tracedecay-usecases/src/advisory/github_runtime/`, landed as three separate commits on one branch to avoid conflicting edits.

| Commit | Item |
| --- | --- |
| `216cac58c` | `gh auth token` credential authority |
| `6ad856734` | pre-flight rate-limit gate |
| `a3d23e094` | CI-path ETag caching |

---

## Item 1 — `gh auth token` credential authority

New `github_runtime/gh_cli.rs`. An unauthenticated client gets 60 requests/hour; the token `gh auth token` already holds gets 5,000.

**Precedence in the `Public` branch** of `open_project_github_release_read_authority_v1` (`releases.rs`): an already-registered real credential *that permits `Contents`* → the local `gh` login → `anonymous()`.

**Caching.** The finding from the first attempt was re-verified against current code and holds: `resolve_registered_github_read_only_credential_v1` (`network.rs`) calls `authority.resolve(..)`, drops the secret, and retains only the authority `Arc`; `authorization_for_stored_repository` then calls `resolve` again **on every HTTP request** (`network.rs:666-700`). Per-request shelling out therefore had to be solved inside the authority. Positive *and* negative probe results are cached behind the authority's own mutex with a 300 s TTL, and the probe runs **while the lock is held**, so concurrent resolutions collapse into one child process instead of a stampede.

**Parked-entry hazard.** A `VerifiedPrivate` credential is only valid while its authority occupies the registry slot with a matching generation, so the `gh` authority genuinely takes the `(owner, repo)` slot — which would otherwise make a later real private mount return `Rejected`. `mount_profile_github_read_only_credential_authority_v1` now withdraws only the exact retained `gh` `Arc` (`Arc::ptr_eq`, via the existing `unregister_…` which already refuses a non-matching pointer) and retries the mount **once**. A real credential in that slot is never displaced — proven by `withdrawal_never_evicts_a_foreign_credential_from_the_slot`.

### WriteCapable decision — **return `Verified`, deviation documented**

The trait asks implementations to establish effective provider permissions before returning `Verified`, and offers `WriteCapable` for a token that can write. A `gh` login usually *is* write-capable. I chose to return `Verified` with the five read permissions the read-only clients ask for, rather than spend a `gh auth status` check per TTL window, because:

1. **The alternative costs the thing this exists to save.** `gh auth status` validates the token against `api.github.com` — a provider request per TTL window, which is the exact cost this authority was written to avoid.
2. **It would not even be sufficient evidence.** `gh auth status` reports OAuth *scope labels*, and reports none at all for fine-grained tokens. The trait doc already rejects declared scope strings as proof, so paying the request would buy a weaker signal, not a stronger one.
3. **The grant is bounded to no more than `anonymous()` already had.** This authority is only ever mounted for a repository the profile mount has already classified `Public`. It replaces `GitHubReadOnlyCredentialV1::anonymous()` and claims no data access anonymous did not have — only a higher request budget.
4. **A write cannot be expressed through this surface.** The credential feeds only `GitHubReadOnlyClientV1` (fixed REST `GET`s plus one static GraphQL *query* document, `GITHUB_REVIEW_THREADS_QUERY_V1`) and `GitHubCiReadOnlyClientV1` (`GET` only). There is no representable mutation for a write-capable token to reach.

The deviation is stated in the doc comment on `GhCliGitHubReadOnlyCredentialAuthorityV1` so a later reader meets it at the type, not only here.

### Token-handling audit — every place a token value can flow

| # | Site | Container | Notes |
| --- | --- | --- | --- |
| 1 | `gh auth token` child stdout | OS pipe | `stdout` piped, **`stdin` and `stderr` both `Stdio::null()`** so provider diagnostics can never reach a TraceDecay log. Invoked as an argv array `["auth","token"]` via `Command::args` — never a shell string. |
| 2 | `probe_gh_auth_token_v1` read buffer | `Zeroizing<Vec<u8>>` | Read is `.take(MAX_GH_TOKEN_BYTES_V1 + 1)`; anything over 4096 bytes is rejected, not truncated. |
| 3 | UTF-8 decode | `Zeroizing<String>` | `String::from_utf8` takes ownership of the same allocation, so the decoded buffer is zeroized. Non-UTF-8 → `None`. |
| 4 | Trim | `Zeroizing<String>` | Empty/whitespace-only → `None`. |
| 5 | `GhCliTokenSourceV1::token` return | `Option<Zeroizing<String>>` | The trait's only output. Doc forbids implementors deriving `Debug`/Serde. |
| 6 | `GitHubReadOnlyCredentialSecretV1::from_zeroizing` | `Zeroizing<String>` inside the newtype | Pre-existing type; re-validates length ≤ 4096 and rejects any ASCII control or whitespace byte. Implements **neither `Debug` nor Serde** — enforced by the pre-existing `github_credentials_are_not_debuggable_or_serializable` static assertion. |
| 7 | `CachedGhTokenV1.secret` | `Option<GitHubReadOnlyCredentialSecretV1>` | Behind the authority's mutex. Cannot derive `Debug`/Serde, because its only field that carries bytes cannot. |
| 8 | `GitHubReadOnlyCredentialAuthorityOutcomeV1::Verified { secret }` | moved newtype | Pre-existing enum; no `Debug`, no Serde. |
| 9 | `resolve_registered_github_read_only_credential_v1` | — | Explicitly `drop(secret)`; retains only the authority `Arc`, permissions, and generation. **No token bytes cross this boundary.** |
| 10 | `GitHubReadOnlyCredentialSecretV1::authorization_header` | `Zeroizing<String>` | Builds `Bearer …` inside a zeroizing container. |
| 11 | `GitHubCredentialAuthorizationV1::Private(Zeroizing<String>)` | `Zeroizing<String>` | Pre-existing carrier from credential to request builder. |
| 12 | `…get(url, etag, permission)` → `request.header("Authorization", …)` | `&str` borrowed from the `Zeroizing` | The only place a token value leaves the process, into the outbound TLS request. |

Nothing token-shaped is written to a log, a store, a `Debug` output, a serialized structure, a URL, or an error message. `GitHubHttpReadConfigV1` (which *is* `Debug`) carries only URIs and timeouts.

**Test-build safety.** `GhAuthTokenCommandSourceV1` and `probe_gh_auth_token_v1` are `#[cfg(not(any(test, feature = "test-transport")))]` — they **do not exist** in a test build, so no test can construct the one type that can spawn a probe. The default source under `cfg(test)`/`test-transport` is `NullGhCliTokenSourceV1`, which always returns `None`. (This is a superset of the requested `cfg(test)`: `--all-targets` builds integration tests without `cfg(test)`, so the feature gate closes that hole too.) Sources are injected for the tests that need one through `GhCliTokenSourceGuardV1`, which clears every retained authority and its cached probe on both ends of its lifetime. Fixture tokens are obvious placeholders (`gh-fixture-not-a-real-token`); header presence is asserted with `assert!(…contains…)`, never `assert_eq!`, so a failure cannot print header bytes.

**Degradation.** Every failure path yields anonymous, never an error: `gh` absent from `PATH` (spawn fails), not logged in / non-zero exit, empty or whitespace-only or oversized or non-UTF-8 output, a hung child (killed on a 3 s deadline), a poisoned lock, registry rejection, and a resolved credential that does not permit `Contents`.

---

## Item 2 — pre-flight rate-limit gate

**Verified absent first.** There was no pre-flight gate. Rate limiting was purely reactive: `rate_limit_checkpoint` reads `x-ratelimit-*` off a response, and `403`/`429` becomes `RateLimited` *after* the request was spent. Nothing anywhere read `/rate_limit` or checked quota before a burst.

New `github_runtime/rate_gate.rs`. `GitHubCiReadOnlyClientV1` now retains every checkpoint it observes (from both `200` and rate-limited responses), keyed by `(credential generation, owner, repository)` so one credential's observation never gates another's burst. Each paged CI loop asks the gate before its first request:

- **Exhausted** → refuse before a single request is spent, returning `RateLimited` with the proving checkpoint.
- **Degraded** → shorten the burst to the admitted request count.
- **Unknown** (no observation, or the observed window already reset) → leave the existing page bound untouched. The gate never invents a refusal it cannot prove.

A reserve of 5 requests stays back so one CI discovery pass cannot starve a concurrent review resume or release read. **The check costs no provider request of its own** — the verdict comes from headers earlier responses already carried, which is strictly cheaper than a `/rate_limit` round trip. Gated sites: `collect_workflow_runs`, `collect_workflow_jobs`, `collect_check_runs`, and `read_annotations`. The port method defaults to `Unknown`, so in-process fixtures are never gated.

---

## Item 3 — CI-path ETag caching

**Followed the existing shape, as instructed.** The review-cursor resume path already carries an ETag through `GitHubReadResumeV1`, sends `If-None-Match` in `GitHubReadOnlyClientV1::get`, and decodes `HttpResponseV1::NotModified`. `GitHubCiReadOnlyClientV1::get` now takes the same `etag: Option<&GitHubReviewEtagV1>` parameter and reuses the same decoded variant — there is no second conditional-request mechanism.

The missing half was a body store; new `github_runtime/ci_cache.rs` supplies it. Each `200` retains its body beside the ETag the provider issued. A `304` — which does not count against the rate limit — returns that body. Edge cases are explicit rather than incidental: a `200` **without** a validator drops any retained body (the next read cannot be conditional, so nothing may stay retained under it), and a `304` with nothing retained stays `Unavailable` rather than inventing an empty body.

Entries are keyed by credential generation as well as URL, so a body fetched under one credential is never served to another and a remount — which always receives a fresh generation — starts cold. Bounded per entry (2 MiB), in total (16 MiB), and in count (512).

---

## Verification

Per-crate only, as instructed — no workspace build, no workspace clippy.

```
$ cargo check -p tracedecay-usecases --all-targets --locked
    Finished `dev` profile [optimized + debuginfo] target(s) in 9.88s

$ cargo test -p tracedecay-usecases --lib --locked
test result: ok. 864 passed; 0 failed; 1 ignored; 0 measured; 0 filtered out; finished in 67.63s
```

`cargo clippy -p tracedecay-usecases --lib --no-deps -- -D warnings` reports **9 findings, all pre-existing in `host_admission.rs`** (`only_used_in_recursion` ×7, `unconditional_recursion` ×2, lines 485-504), introduced by `64dbb4bf1 perf(observation): batch admission and preparation` on the base branch. Zero findings in any file this PR touches.

RED/GREEN output for each item is in the branch report; every RED was produced by reverting the exact production seam and every assertion is on bytes, request headers, or call counts — never on elapsed time.

---

## ⚠️ Base branch does not compile (pre-existing, not from this PR)

`origin/cursor/simplify-pr421-hot-paths` at `f1327d565` **fails `cargo check`** before any change here:

1. `crates/tracedecay-runtime-core/src/background_cpu.rs` does not exist in the branch, and `pub mod background_cpu;` is missing from that crate's `lib.rs` — but `tracedecay-sessions/src/observation.rs:24` and `tracedecay-usecases/src/host_admission.rs:26` both import from it. The file **does** exist as an uncommitted working-tree file in the shared main checkout, so `64dbb4bf1` appears to have been pushed without it.
2. `tracedecay-sessions/src/runtime/ingest/failure.rs:517` has a non-exhaustive match: `ObservationApplicationError::BatchWorkerStopped` (added at `observation.rs:315`) is not covered. The main checkout has an uncommitted arm for it.

To verify this PR I applied both fixes **locally and left them uncommitted** — they are another agent's in-flight work and are deliberately not in this branch. CI on this PR will fail to build until the owner of `64dbb4bf1` pushes them.


ScriptedAlchemy:
Closing from the #421 integration queue after review. The current head can still issue nested GraphQL calls after an in-flight response reports remaining=0, and helper-only tests do not prove REST/GraphQL dispatch suppresses the network. Re-open as a clean PR only after every nested request uses the typed preflight gate and transport-level call-count tests cover initial and nested exhaustion.

PR #687, opened 2026-08-24
docs(perf): diagnose the tokio async-worker skew

ScriptedAlchemy:
## Outcome: diagnosis, no production change

The reported "2 of 16 tokio workers carry ~90% of busy time" is real and
reproducible. Its cause is **not** a scheduling or width problem, and the two
obvious fixes for a scheduling problem — resizing the pool, or adding yield
points — would both make things worse or do nothing. No production behaviour is
changed here.

What lands instead: a self-tested profiler that turns Hotpath's already-present
metrics into the falsifiable numbers, and the measured breach recorded against
the doc that currently claims the opposite.

## What the hot workers are running

`session_temporal_refresh.history` — the historical transcript-ingest pass
driven by the session-refresh scheduler
(`src/daemon/session_temporal_refresh_scheduler/worker.rs:53`).

Hotpath's `total_poll_duration_ns` is wall time **strictly inside
`Future::poll`**, so an `.await` cannot contribute to it. Across three
independent 60s windows that one label accounts for **96.7–97.6% of everything
the async pool did**.

## Why it concentrates

Exactly two long-lived tasks — one project-scoped ingestor and one
profile-scoped one (`run_session_temporal_refresh_scheduler`, `call_count: 2`).
Runnable async concurrency is 2, not 16. Work-stealing has nothing to steal
(14–23 steals per 60s window), and no pool width redistributes a slice a worker
is already inside: a poll is not preemptible until it returns.

Ruling out the other candidates from the brief:

- **Blocking work on async workers** — yes, this is it, but not at a store
  boundary. SQLite already runs on dedicated reader/writer OS threads
  (`reader/worker.rs`, `writer.rs`), and `ReadSnapshot::query` already uses
  `spawn_blocking`. The synchronous slices sit in the ingest pass itself,
  between store awaits.
- **A single `select!`/actor funnel** — no. Lock waits are nanoseconds across
  7,324 instrumented mutexes and 2 rw-locks.
- **Oversubscription** — no. 16 workers on 96 usable cores, `nproc` 96, no
  cgroup CPU limit.

## Is 16 workers the right width?

Yes, and it should not be touched. The width is *why* this breach has not
surfaced as a latency regression: **11 of 16 workers ran zero polls** in each
sampled minute, so an arriving request always finds a parked worker instead of
queueing behind a 56ms poll. Narrowing the pool to "match" observed concurrency
would delete exactly that headroom and convert a benign skew into a real tail.

The number to watch is microseconds per poll, not busy share.

## Evidence

96-core host, `--profile perf`, features `production,hotpath,hotpath-mcp`,
`--cfg tokio_unstable`, isolated profile root under `/tmp` (the user's
`~/.tracedecay` is never touched), one registered 3,114-file corpus. Three
independent daemon runs.

**Run 1 — 325s, whole daemon lifetime.** 5 of 16 workers carry 98.5% of busy
time; 5 workers completely idle; cumulative steals 0–9; the reserved indexing
pool consumes 4.22 CPU-seconds over the entire run.

**Run 2 — 60s window, via the profiler added here:**

```
window 60.0s   workers 16   blocking threads 8 (7 idle)
async pool busy 91.92s = 1.53 cores   steals 14

 wrk    busy_ms   share%      polls    us/poll   steals     parks
  14      61824    67.26       6926     8926.4        2     13084
  12      13398    14.58      10979     1220.3        2     21509
   6       8814     9.59       7343     1200.3        2     15314
  15       5892     6.41       9293      634.0        4     17940
  10       1990     2.16       3452      576.5        4      6424
  (workers 0-5, 7-9, 11, 13: zero polls)

top-1 busy share  67.26%
top-2 busy share  81.84%
workers with any busy time  5/16
worst us/poll (>=100 polls) 8926.4

in-poll execution by labelled future (excludes every .await):
   cores    poll_s     polls    us/poll  label
    1.49     89.68     35648     2515.7  session_temporal_refresh.history
```

89.68s of the pool's 91.92s busy time — **97.6%** — is synchronous execution
under one label.

**Run 3 — same scenario, with the profiler's threshold gate armed:**

```
window 60.0s   workers 16   blocking threads 6 (5 idle)
async pool busy 86.88s = 1.45 cores   steals 23

 wrk    busy_ms   share%      polls    us/poll   steals     parks
  12      50969    58.67        910    56009.9        1      1533
  14      11860    13.65       7595     1561.6        5     15223
   0       9393    10.81       6612     1420.6        8     12881
   9       8743    10.06       6881     1270.6        5     12851
  10       5913     6.81       3770     1568.4        4      7487
  (11 workers: zero polls)

top-2 busy share  72.32%
worst us/poll (>=100 polls) 56009.9

   cores    poll_s     polls    us/poll  label
    1.40     84.03     23884     3518.3  session_temporal_refresh.history

THRESHOLD BREACHED: top-2 worker busy share 72.32% exceeds 60.00%
THRESHOLD BREACHED: worst worker occupancy 56009.9 us/poll exceeds 200.0 us/poll
```

Exit status 1. **A single poll held a worker for 56 milliseconds.**

Supporting function timing from the same scenario (inclusive, so awaits are
included — these locate the pass, they do not measure occupancy):

```
    509.73 s        2   254.86 s  run_session_temporal_refresh_scheduler
    509.21 s        4   127.30 s  session_history_refresh
    188.78 s        4    47.19 s  ingest::project_provider::run_codex
     44.31 s    16386     2.70 ms  jsonl_observation_admission::admit_jsonl_observations
```

## Where a fix has to land, and why none is in this PR

The cost is CPU **placement**, not balance. This work belongs off the request
runtime the way indexing already is.

- A **yield point** redistributes the slices across workers without moving a
  single cycle off the serving pool. It would improve the busy-share metric and
  fix nothing — precisely the cosmetic rebalance the brief warns against.
- **`spawn_blocking` cannot wrap the pass**, because the pass is a future with
  interleaved store awaits, not a closure. A real fix has to batch the
  synchronous work into blocking chunks at a boundary that owns its data, which
  is a restructure of the ingest pipeline and its cancellation/cursor handling.
  That is not something I can land and prove in one session on a path this
  privacy- and cursor-sensitive, so it is scoped rather than guessed at.

Leaf attribution is left open, and the obvious instrument cannot close it:
Tokio names async worker threads and blocking-pool threads with the same
`thread_name_fn`, so `perf --comms` cannot separate the two pools. Under that
combined family the largest single product symbol during ingest is
`privacy::rules::contains_ignore_ascii_case` (5.16% of whole-process cycles,
~7.4% for the privacy-detector family) — a plausible in-poll leaf, since
redaction is pure CPU over transcript text. But SQLite parser symbols in the
same family belong to `ReadSnapshot::query`, which is correctly placed on
`spawn_blocking`. Naming the exact leaf needs a `#[hotpath::measure]` inside the
pass or `HOTPATH_FOCUS`, not a thread filter. This is written down in the doc so
the follow-up starts from the obstacle rather than rediscovering it.

## Scope of the measurement

All three runs start from a **cold, isolated profile**, so the ingest is doing a
full historical catch-up. That is a real workload (fresh install, wipe, new
profile) and the code path is identical when warm — warm passes carry fewer
bytes, not a different structure. The steady-state magnitude is not measured
here.

## What lands

- `scripts/profile-tokio-worker-balance.sh` + `scripts/lib/tokio_worker_balance.py`
  — reads only Hotpath's metrics HTTP routes, so no product code change and no
  restart. `--max-top2-share` / `--max-us-per-poll` make it a before/after gate.
  Dependency-free stdlib Python; `--self-test` covers the arithmetic over
  fixtures with no daemon.
- `docs/SERVING-PATH-PERFORMANCE.md` — an "Open breach" section following the
  existing pattern, plus a pointer from Principle 2, which currently asserts the
  invariant this measurement contradicts.

## Verification

```
$ ./scripts/profile-tokio-worker-balance.sh --self-test
tokio_worker_balance self-test: ok

$ ./scripts/profile-tokio-worker-balance.sh --port 59999 --seconds 1; echo $?
no Hotpath tokio_runtime metrics at 127.0.0.1:59999. ...
3
```

`npx commitlint --from origin/cursor/simplify-pr421-hot-paths --to HEAD` — 0
problems. No Rust source is touched, so there is no crate to rebuild; the
`production,hotpath,hotpath-mcp` binary that produced every number above built
clean.

## Note on the base branch

`origin/cursor/simplify-pr421-hot-paths` does not currently compile: commit
`64dbb4bf1` references `tracedecay_runtime_core::background_cpu`, but that
module is not in the tree (it exists uncommitted in the shared checkout). The
measurements were taken at `434f4d492`, the last building ancestor; these two
commits are then rebased onto the branch head and touch only `scripts/` and
`docs/`, so they neither depend on nor conflict with that fix.

### DIFF
--- a/docs/SERVING-PATH-PERFORMANCE.md
+++ b/docs/SERVING-PATH-PERFORMANCE.md
@@ -199,6 +199,10 @@ have no finish line, so they stay paced:
   them defer or shrink a slice while agents are actively querying. Results
   are identical; only pacing changes.
 
+The historical half of the session-refresh sweep does none of this today; see
+"Open breach: historical transcript ingest violates Principle 2" below for the
+measurement and for why the answer is not a wider or narrower worker pool.
+
 ### 3. Hash where data is born, never where it is served
 
 Content digests (payload references, canonical record digests) are computed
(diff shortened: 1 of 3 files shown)


ScriptedAlchemy:
Merged directly into PR #421 after #663 as cd0013d4e and 09c0598be; the profiler self-test and syntax checks passed before integration.

PR #688, opened 2026-08-24
fix(build): land the batching commit's missing consumers

ScriptedAlchemy:
## Merge this first — supersedes #676 and #681

`cursor/simplify-pr421-hot-paths` does not compile. `perf(observation): batch admission and preparation` landed changes whose consumers, and one whole file, were never `git add`ed, so they exist only in one local working tree. **Every branch cut from the pushed head fails before its own code is built** — that is currently every open PR in this stack, and six agents have independently hit it and worked around it four different ways.

| Missing | Referenced by |
|---|---|
| `crates/tracedecay-runtime-core/src/background_cpu.rs` (whole file + `mod` line) | `observation.rs:24`, `host_admission.rs:26` |
| `ObservationApplicationError::BatchWorkerStopped` arm | `runtime/ingest/failure.rs:517` |
| `HostAdmissionOutcome.recovery` field ×5 | `kiro.rs`, `cline_like.rs`, `ingest/tests.rs` ×2, `host_admission.rs`, `integration_test_support.rs` |
| 5 readiness counters, now `pub(crate)` in another crate after the hooks extraction | `sessions/claude_observation_benchmark/baseline.rs` |

## Deliberately *not* included

An earlier attempt (#681) also carried `crates/tracedecay-code-index/src/parallelism.rs`, whose four missing items are referenced by unpushed work. That commit changes `parallelism::install` to return a `Result` and removes `install_daemon_worker_ceiling`, which breaks `tracedecay-semantic` and `tracedecay-cli` on this base — it needs its own unpushed follow-up (`bdd465ae9`, +276 lines) to be coherent, and supplying `available_memory_bytes` to the new plan API is a design decision belonging to its author. **#681 is closed in favour of this one**; that refactor should land with its follow-up.

## One judgement call, flagged for the author

`BatchWorkerStopped` needed a classification and none was written. Classified **retryable / `Unavailable`** (`observation_batch_worker_stopped`): a worker that went away before reaching a verdict decided nothing about the payload, so re-running can succeed. Classifying it `permanent` would silently drop observations on a transient worker fault. Change this line if different semantics were intended.

The five readiness counters are exposed as accessors rather than `pub` fields — widening the fields leaked several `pub(crate)` field *types* out of the crate (E0446), a much larger API change than the benchmark needs.

## Verification

`cargo check --workspace --all-targets --locked` — clean.

### DIFF
--- a/crates/tracedecay-agent-hosts/src/hooks/analytics/readiness.rs
+++ b/crates/tracedecay-agent-hosts/src/hooks/analytics/readiness.rs
@@ -178,6 +178,31 @@ pub struct HookCompletedReadinessDistributions {
     pub(crate) rows_folded_to_other_host: u64,
     pub(crate) disposition_values_folded_to_unknown: u64,
 }
+impl HookCompletedReadinessDistributions {
+    /// Readiness counters the root-crate observation benchmark folds. The
+    /// struct moved into this crate, so its `pub(crate)` fields are no longer
+    /// reachable from there; these expose exactly the five it reads.
+    pub fn source_event(&self) -> &str {
+        &self.source_event
+    }
+
+    pub fn input_rows_received(&self) -> u64 {
+        self.input_rows_received
+    }
+
+    pub fn input_rows_processed(&self) -> u64 {
+        self.input_rows_processed
+    }
+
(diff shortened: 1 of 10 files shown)


ScriptedAlchemy:

PR #689, opened 2026-08-24
test(dashboard): verify graph payloads across many projects

ScriptedAlchemy:
Verification of whether the dashboard serves **real** graph data across many
enrolled projects, plus the harness that makes that checkable.

Two things are worth saying up front:

1. **The dashboard is not the blocker.** It reports unavailability honestly and
   never fabricates a payload. No API or dashboard behaviour change was needed
   and none was made.
2. **`cursor/simplify-pr421-hot-paths` does not compile**, and not only for the
   reason #676 fixes. That blocked the end-to-end half of this verification,
   and it is the finding I would act on first.

## What was run

- Isolated profile (`TRACEDECAY_DATA_DIR=/tmp/vdm-mp/profile/.tracedecay` with a
  matching `TRACEDECAY_DAEMON_SOCKET`). The real `~/.tracedecay` was never
  touched. Every process was killed by recorded PID; no `pkill -f`/`pgrep -f`.
- Stack: base + #676 + #670 + #671, plus the fixes and carries listed at the
  bottom.
- **12 projects enrolled**, deliberately past the old cap of 3: react-router,
  mold, lcm, context, qvm, machinen, tailwindcss, stocktok, rsdoctor, midscene,
  rstest, rspress. Then a second, cleaner run with **6** (all under the cap).

## Per-project results

`SERVED` means `/api/projects/{id}/plugins/graph/overview` returned a non-null
payload with a non-zero node count.

### Run A — 12 projects (default worker sizing)

| # | Project | Enrolled | Graph overview | HTTP | `domain_state` | Reason carried by the envelope |
|---|---|---|---|---|---|---|
| 1 | react-router | ok | no | 200 | unknown | verified code graph is not ready for the exact project root |
| 2 | mold | ok | no | 404 | — | registered project graph is not mounted |
| 3 | lcm | ok | no | 404 | — | registered project graph is not mounted |
| 4 | context | ok | no | 404 | — | registered project graph is not mounted |
| 5 | qvm | ok | no | 404 | — | registered project graph is not mounted |
| 6 | machinen | ok | no | 200 | unknown | missing_registry |
| 7 | tailwindcss | ok | no | 200 | unknown | missing_registry |
| 8 | stocktok | ok | no | 200 | unknown | missing_registry |
| 9 | rsdoctor | ok | no | 200 | unknown | missing_registry — **scheduler capacity exhausted, looping** |
| 10 | midscene | ok | no | 200 | unknown | missing_registry — **scheduler capacity exhausted, looping** |
| 11 | rstest | ok | no | 200 | unknown | missing_registry — **scheduler capacity exhausted, looping** |
| 12 | rspress | ok | no | 200 | unknown | missing_registry — **scheduler capacity exhausted, looping** |

### Run B — 6 projects, under every cap, zero daemon errors

| # | Project | Enrolled | Graph overview | `domain_state` | Reason |
|---|---|---|---|---|---|
| 1 | lcm (active) | ok | no | unknown | verified code graph is not ready for the exact project root |
| 2 | context | ok | no | unknown | missing_registry |
| 3 | qvm | ok | no | unknown | missing_registry |
| 4 | machinen | ok | no | unknown | missing_registry |
| 5 | tailwindcss | ok | no | unknown | missing_registry |
| 6 | stocktok | ok | no | unknown | missing_registry |

Run B logged **zero** capacity errors, **zero** resident-memory denials and
**zero panics** across the entire run — including react-router in run A.

## What is verified

**Enrolment past the old cap works.** All 12 `tracedecay init` calls returned
rc=0 and `/api/projects` answers `domain_state: ready`, `count: 12`. The
"graph capacity budget exhausted (limit 8)" refusal at the 4th project is gone.
#671 does what it says.

**A third cap remains, and #671 does not lift it.** Three independent limits
gate multi-project indexing:

| # | Limit | Value | Lifted by #671? |
|---|---|---|---|
| 1 | `MAX_RETAINED_PROJECT_RUNTIME_OWNERS` / `MAX_RETAINED_REMOTE_NODE_OWNERS` | 8 → 4096 | yes |
| 2 | `GraphDbRegistryConfig { max_open }` | 8 → 8192 | yes |
| 3 | **`MAX_CACHED_PROJECT_SERVERS`** (`src/daemon.rs:53`) | **still 8** | **no** |

Cap 3 is passed as `max_worktrees` to the code-index scheduler registry in
`DaemonInvocationState::default()` (`src/daemon/invocation_state.rs:48`), which
`src/daemon/bootstrap.rs:113` uses for the production daemon. Past the eighth
project:

```
code-index scheduler could not be mounted: code-index identity construction
failed: code-index scheduler capacity is exhausted
```

Measured, this lands exactly where the arithmetic predicts: with 12 projects
enrolled in order, the four that fail are **#9–#12**. They are refused *in a
hot loop* — roughly twice a second, indefinitely, with no backoff and no
terminal state. `src/daemon.rs` is unmodified on this branch, so this is
committed behaviour, not something my build introduced.

**A fourth constraint sits underneath it.**
`DEFAULT_PROCESS_RESIDENT_MEMORY_LIMIT_V1` is a process-wide **6 GiB** pool and
each mounted code index reserves `workers × 128 MiB`
(`INDEX_WORKER_RESIDENT_BUDGET_BYTES_V1`). The number of projects that can hold
an index at once is therefore `6 GiB / (workers × 128 MiB)` — a function of host
width, not of how many projects the user enrolled. On committed sizing
(`indexing_worker_target` clamps to `DEFAULT_MAX_INDEXING_WORKERS = 8`) that is
≤1 GiB per project, so roughly six projects fit — already tighter than cap 3.

I hit this directly, and it is legible:

```
resident-memory admission denied: used=6442450944 requested=6442450944 limit=6442450944
```

`6442450944` is exactly 6 GiB: one project's plan asking for the whole pool.
Setting `TRACEDECAY_INDEX_WORKERS=3` removed the denial entirely and left cap 3
as the binding constraint, which is how the #9–#12 boundary was isolated.

**The dashboard reports failure honestly.** `domain_state: unknown` +
`payload: null` is what `DashboardEnvelopeV1::unavailable` produces, and it
pushes the cause onto `coverage.omission_reasons`. The client already handles
this: `decodeEnvelopeBody` turns a null payload into
`{ outcome: 'transport', state: envelope.domain_state, detail: envelope.coverage.omission_reasons[0] }`
and `envelopeReadState` renders it as a blocked state carrying the daemon's own
sentence. Every "reason" column in the tables above came out of that field.
Patching the dashboard to paper over a broken index would have been worse than
reporting it, so nothing was patched.

## What is NOT verified, and why

**No code-index generation ever sealed**, so "renders real data" could not be
confirmed end to end. `tracedecay_status` against the isolated profile is
unambiguous even for the active project of the clean 6-project run:

```
graph_statistics:      { state: unavailable, reason: exact_scope_generation_not_ready }
code_index_freshness:  { status: warming, staleness_state: indexing,
                         latest_generation_id: null, sealed_at_micros: null }
```

Shards did grow (112 MB → 219 MB per project, 2.5 GB total), so work was
happening; it just never sealed, including for `lcm` at 141 source files.

**I am not attributing that to the base branch**, because the base cannot be
built and the binary I did produce carries 17 files of somebody's uncommitted
work-in-progress (below). Whether sealing is broken on a correctly assembled
tree is exactly the thing that becomes checkable once the branch compiles.

One lead for whoever picks it up: every reconcile failure in the clean 6-project
run was the same one —

```
code-index background reconcile failed; the served generation stays stale
  error=code-index graph activation failed: the read port was cancelled
```

I checked whether my own dashboard requests were causing those cancellations by
going completely silent for seven minutes. They kept accruing (24 → 30) with no
client touching the daemon, and nothing sealed in that window either. So the
cancellation is self-inflicted by the daemon, not request-driven, and the
reconcile never reaches a sealed generation. That is where I would start.

## What this PR adds

`dashboard/stories/live-multiproject.ts` (`npm run live:multiproject`).
`live-sweep.ts` proves each workspace renders; it cannot prove anything is
behind that render, because a workspace whose read failed still draws its chrome
and passes — which is how the Code workspace passed while
`/api/plugins/graph/overview` was answering 200 with a null payload. This sweep
asserts the payload instead: it walks `/api/projects` and requires every
enrolled project to serve a graph overview with a non-zero node count, printing
node/edge/file/language counts per project and `coverage.omission_reasons` when
a read fails. It exits non-zero, so it is usable as a gate. Every table above is
its output.

Two tests, both asserting on content, never on elapsed time:

- `dashboard/src/data/query/envelope.test.ts` — an unavailable envelope must
  carry the daemon's reason through the decode. The existing test asserted the
  state but not that the reason survived, and the reason is the only part that
  says what to fix.
- `crates/tracedecay-sessions/.../ingest/tests.rs` — covers the classifier arm
  added below.

## The base branch does not compile

Beyond the missing `background_cpu` module that #676 restores:

1. **Fixed here.** `crates/tracedecay-sessions/src/runtime/ingest/failure.rs`
   had a non-exhaustive match — the batching commit added
   `ObservationApplicationError::BatchWorkerStopped` with no arm to classify it.
   Classified retryable/`Unavailable`: a worker that went away never reached a
   verdict, so the same input can succeed on retry. Deliberately different from
   the neighbouring `BatchContainsNonDurable`, which is permanent because the
   batch itself was invalid.

2. **Not fixed here — needs its owner.** Four symbols are referenced by
   `src/daemon/code_index_scheduler.rs` and defined nowhere on the branch:
   `CodeIndexWorkerPlanInstallErrorV1`, `worker_reservation_bytes`,
   `installed_worker_status`, `with_background_cpu_permit`. All four exist in
   the main worktree's **uncommitted** `parallelism.rs` — the same
   never-`git add`ed class of defect as `background_cpu`, but far larger (~694
   added / ~170 removed lines in that file alone, cascading through
   `tracedecay-semantic`, `src/daemon/` and `crates/tracedecay-cli`).

To get a runnable daemon I carried 17 files from that uncommitted tree
**locally only — none are in this PR**. **This PR will therefore not build in CI
until #676 and the missing parallelism slice are committed.** Its own content —
one TypeScript harness, two tests, one small Rust fix — is independent of all of
it.

## Caveats on the carried build

Stated so none of this reads as stronger evidence than it is.

- Carrying `src/daemon/code_index_scheduler.rs` overwrote **#670's panic-guard
  wiring** (21 hunks against a file that had itself diverged; hand-merging was
  judged too error-prone). `reconcile_panic_guard.rs` is present but no longer
  referenced, so this run does not exercise #670's isolation path. It does show
  **zero panics across every run**, react-router included.
- On #670's other half: the workspace declares no `[profile.release]`, so
  release inherits cargo's `debug-assertions = false` and the `debug_assert!` in
  `generate_node_id` **cannot fire in a release daemon at all**. The 114-retry
  loop was reachable only from a build with assertions on — which
  `[profile.perf] inherits = "dev"` (what `cargo test-all`/`test-ci` use) is.
  The construct itself is real and still present:
  `react-router/integration/fs-routes-test.ts:287` is `test.describe("", …)`.
  #670 is right to drop the assert; a release binary simply could not have
  reproduced that panic either way.

### DIFF
--- a/crates/tracedecay-runtime-core/src/background_cpu.rs
+++ b/crates/tracedecay-runtime-core/src/background_cpu.rs
@@ -1,0 +1,461 @@
+//! Process-wide weighted admission for background CPU work.
+//!
+//! The authority counts active CPU units rather than owning an executor. Code
+//! indexing, semantic native threads, and session preparation can therefore
+//! use their existing execution substrates while sharing one hard process
+//! ceiling. FIFO waiter order prevents a continuously busy class from starving
+//! another class, and RAII releases capacity on success, cancellation, or
+//! unwind.
+
+use std::cell::Cell;
+use std::collections::VecDeque;
+use std::fmt;
+use std::num::NonZeroUsize;
+use std::sync::{Arc, Condvar, Mutex, OnceLock};
+
+#[derive(Debug)]
+struct BackgroundCpuWaiterV1 {
+    units: usize,
+}
+
+#[derive(Default)]
+struct BackgroundCpuStateV1 {
+    active_units: usize,
+    waiters: VecDeque<Arc<BackgroundCpuWaiterV1>>,
+}
+
+/// One process-wide background CPU budget shared across subsystems.
+pub struct ProcessBackgroundCpuV1 {
+    width: NonZeroUsize,
+    state: Mutex<BackgroundCpuStateV1>,
+    available: Condvar,
+}
+
+impl fmt::Debug for ProcessBackgroundCpuV1 {
+    fn fmt(&self, formatter: &mut fmt::Formatter<'_>) -> fmt::Result {
+        formatter
+            .debug_struct("ProcessBackgroundCpuV1")
+            .field("width", &self.width)
+            .field("active_units", &self.active_units())
+            .finish_non_exhaustive()
+    }
+}
+
+thread_local! {
+    static BACKGROUND_CPU_DEPTH: Cell<usize> = const { Cell::new(0) };
+    static BACKGROUND_CPU_UNITS: Cell<usize> = const { Cell::new(0) };
+}
+
+struct BackgroundCpuScopeV1;
+
+impl BackgroundCpuScopeV1 {
+    fn enter() -> Self {
+        BACKGROUND_CPU_DEPTH.with(|depth| depth.set(depth.get().saturating_add(1)));
+        Self
+    }
+}
+
+struct YieldedBackgroundCpuV1<'a> {
+    authority: &'a Arc<ProcessBackgroundCpuV1>,
+    units: usize,
+    depth: usize,
+}
+
+impl Drop for YieldedBackgroundCpuV1<'_> {
+    fn drop(&mut self) {
+        self.authority.admit_units(self.units);
+        BACKGROUND_CPU_UNITS.with(|units| units.set(self.units));
+        BACKGROUND_CPU_DEPTH.with(|depth| depth.set(self.depth));
+    }
+}
+
+impl Drop for BackgroundCpuScopeV1 {
+    fn drop(&mut self) {
+        BACKGROUND_CPU_DEPTH.with(|depth| {
+            let remaining = depth.get().saturating_sub(1);
+            depth.set(remaining);
+            if remaining == 0 {
+                BACKGROUND_CPU_UNITS.with(|units| units.set(0));
+            }
+        });
+    }
+}
+
+impl ProcessBackgroundCpuV1 {
+    fn new(width: NonZeroUsize) -> Self {
+        Self {
+            width,
+            state: Mutex::new(BackgroundCpuStateV1::default()),
+            available: Condvar::new(),
+        }
+    }
+
+    #[must_use]
+    pub const fn width(&self) -> NonZeroUsize {
+        self.width
+    }
+
+    #[must_use]
+    pub fn active_units(&self) -> usize {
+        self.state
+            .lock()
+            .unwrap_or_else(std::sync::PoisonError::into_inner)
+            .active_units
+    }
+
+    #[must_use]
+    pub fn waiting_work_units(&self) -> usize {
+        let state = self
+            .state
+            .lock()
+            .unwrap_or_else(std::sync::PoisonError::into_inner);
+        waiting_units(&state)
+    }
+
+    /// Acquire one CPU unit, waiting in FIFO order when the process budget is
+    /// full. The returned guard must remain alive for the active work unit.
+    pub fn acquire(self: &Arc<Self>) -> BackgroundCpuPermitV1 {
+        self.acquire_units(1)
+    }
+
+    /// Acquire one CPU unit only when no earlier waiter exists and capacity is
+    /// immediately available.
+    pub fn try_acquire(self: &Arc<Self>) -> Option<BackgroundCpuPermitV1> {
+        let mut state = self
+            .state
+            .lock()
+            .unwrap_or_else(std::sync::PoisonError::into_inner);
+        if !state.waiters.is_empty() || state.active_units >= self.width.get() {
+            return None;
+        }
+        state.active_units += 1;
+        record_state(&state, self.width);
+        Some(BackgroundCpuPermitV1 {
+            authority: Arc::clone(self),
+            units: 1,
+        })
+    }
+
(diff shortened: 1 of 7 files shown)


ScriptedAlchemy:
Integrated the independent multiproject dashboard sweep directly into PR #421 as 0fc6943ec. The missing background-CPU module and BatchWorkerStopped classification were already carried by #421 and were intentionally not duplicated.

PR #690, opened 2026-08-24
perf(mcp): cache tools/list and pin daemon panic isolation

ScriptedAlchemy:
Five findings from an earlier daemon/MCP audit that no commit had ever acted on. Each was re-verified against current code before any change. Two were real and are fixed; one was already fixed; one was not real as stated; one is real and deliberately **not** fixed, because fixing it would cross an authority boundary.

> **Depends on #681.** This branch carries `origin/claude/base-build-fixes` (2 of its 3 commits) plus one extra commit covering three root-crate sites #681 does not reach. The base branch does not compile without them, so no finding could be verified. See "Carried base fixes" at the end.

---

## 1. `tools/list` rebuilds the whole catalog per request — REAL, FIXED

`get_maximal_tool_definitions()` in `src/mcp/tools/definitions.rs` assembled ~200 `ToolDefinition`s, each with a full JSON Schema, on every call. There was no cache.

It is a pure function of static data: the application catalog is already a `LazyLock` snapshot, and `ast_grep_available()` is already a `OnceLock` host probe. So it is now assembled once per process behind a `LazyLock` and cloned per caller.

**Nothing session-scoped is frozen into the cache.** Every per-session pass — `apply_context_budget`, `apply_context_warming_budget`, and the profile/capability/scope filtering in `get_catalog_filtered_tool_definitions_with_budget` — already ran *after* the registry was built, and still does, on the clone. The second test below pins that invariant.

One deliberate behaviour change: `McpDispatchMetadataError` is not `Clone`, so the cache retains the rendered message and replays it as `Initialization`. A failure here is a deterministic catalog/schema defect, not a transient condition, so the message is what matters.

## 2. Idle reader threads are never reaped — ALREADY FIXED

The reader pool already has a full idle lifecycle in `crates/tracedecay-rusqlite-runtime/src/reader/pool/mod.rs`:

- `retire_idle_at(now)` walks the idle deque, retires burst workers idle longer than `idle_burst_retire` (`idle_burst_retire_ms`, default 60_000), and never breaches the `min_per_hot_shard` floor.
- Retired workers are `shutdown()` **and** `join()`ed — the thread is reaped, not just dropped.
- It is driven from `acquire_lane` on entry and after each notified wake (deliberately not on every bounded poll tick, to keep lock traffic off the hot path).

Three existing tests already cover it. No change made; evidence below.

**Residual, reported not fixed:** retirement is acquire-driven, so a pool that goes completely idle keeps its burst workers (bounded by `max_per_hot_shard`, never below the floor) until the next acquire. Adding a periodic reaper is a process-lifecycle change, and the workers it would target are exactly the ones the next acquire reclaims anyway. Not worth the risk.

## 3. Leaked `serve` processes are never reaped — NOT REAL AS STATED

The premise is that tracedecay spawns `serve` children that outlive it. It does not:

- Nothing in the workspace spawns `tracedecay serve`. Every `["serve"]` reference is **host configuration generation** (Cursor, Codex, Copilot, Kilo, Kiro, Cline, Roo, Gemini) — an MCP host launches `serve`, so there is no parent/child relationship for the daemon to reap.
- `serve` terminates on host disconnect by design: `proxy_transport_to_daemon_with_drain_bound` joins a stdin reader against the proxy loop over a `watch` EOF channel, and an in-flight request gets a bounded `drain_daemon_request_after_disconnect` whose bound is the daemon's own published dispatch ceiling.
- The child processes tracedecay *does* spawn are reaped: `ChildGuard::drop` in `crates/tracedecay-sessions/src/runtime/codex_app_server.rs` calls `kill()` then `wait()`, backed by an active-children registry and a shutdown guard the daemon awaits.

**One shape genuinely is not covered,** and I am reporting rather than fixing it: if a host dies while some descendant still holds the stdin pipe open, `serve` blocks on a read that will never EOF. There is no parent-death watchdog (no `PR_SET_PDEATHSIG`, no `getppid` poll). Both fixes are platform-specific and can kill a live session if the parent legitimately re-execs, so this needs an owner decision, not a drive-by.

## 4. Connection tasks have no panic isolation — ALREADY ISOLATED, NOW PINNED

Both accept loops (`src/daemon/bootstrap.rs`) serve clients from a `JoinSet` and keep draining `join_next`, and the workspace does not set `panic = "abort"`. So a panicking connection already unwinds into a contained `JoinError` rather than taking the daemon down.

The part worth proving is the *permit*: `DaemonClientPermit` has no `Drop` of its own — the semaphore permit is released when the last `Arc<ParkableConnectionAdmission>` drops. With `MAX_CONCURRENT_DAEMON_CLIENTS = 64`, a permit stranded per panic would silently degrade the daemon to refusing all clients. Added a regression test covering both unwind paths, because they leave admission in different states:

- a connection that panics while it still holds its slot;
- a connection that panics **while parked** — `park_admission` has already surrendered the slot and never reaches its re-acquire.

No production change; the test is the deliverable.

## 5. Every MCP request opens a new daemon connection — REAL, DELIBERATELY NOT FIXED

Confirmed: `send_daemon_request_line_with_liveness_poll` (`src/daemon/core_proxy.rs`) calls `connect_to_current_daemon_within` per request, writes the preamble, sends one line, reads to the matching id, and drops the connection. The daemon side would support reuse — it reads preface, handshake, first request, then runs a full rmcp MCP session on the connection.

**I am not pooling it, because the connection is the authority boundary.**

1. **The preamble is the authority record, and it is mutated between requests.** `write_daemon_preamble` writes the auth token and `DaemonHandshake` exactly once per connection; the daemon then serves every later request under that handshake. But the proxy rewrites `routed_handshake` *between* requests — `apply_proxy_initialize_metadata` after every response (changing `project_path`, `allow_init`, `scope_prefix`, `catalog_version`, `tool_list_changed_capable`) and `reset_proxy_handshake_for_initialize` on every `initialize`. A reused connection would keep serving under the stale preamble, so a request issued after `initialize` re-routed the project would execute against the previous project route and `allow_init` authority. That is precisely one request observing another's authority state.
2. **Holding a connection idle holds an admission slot.** Each accepted connection takes a permit from `MAX_CONCURRENT_DAEMON_CLIENTS` (64). Pooling converts a per-request concurrency bound into a per-client one, and idle proxies would occupy slots that saturation-shedding assumes are held by live work.
3. **Closing the connection is how a request is cancelled.** `drain_daemon_request_after_disconnect` and `DaemonClientDeadline` abandon an in-flight request by dropping the connection. A shared connection has no per-request cancellation channel.

A safe version would have to invalidate the pooled connection whenever the handshake bytes change, and hand the slot back while idle. That is a protocol-level change to the daemon's admission and cancellation model, not a client-side optimisation, and it needs an owner decision.

---

## Verification

Every test asserts on counts, task outcomes, or lifecycle state. None asserts on elapsed time.

### Finding 1 — RED (cache bypassed, registry rebuilt per call)

```
running 1 test
test mcp::tools::definitions::tests::maximal_tool_definitions_are_assembled_once_per_process ... FAILED

---- mcp::tools::definitions::tests::maximal_tool_definitions_are_assembled_once_per_process stdout ----
panicked at src/mcp/tools/definitions/tests.rs:336:5:
assertion `left == right` failed: the maximal tool registry was re-assembled after it had
already been built; tools/list rebuilds the whole catalog per request
  left: 9
 right: 1

test result: FAILED. 0 passed; 1 failed; 0 ignored; 0 measured; 2656 filtered out
```

Nine assemblies for one initial call plus eight reads — the pre-fix behaviour exactly.

### Finding 4 — RED (permit leaked in `with_connection_admission`)

```
running 1 test
test daemon::tests::lifecycle::a_panicking_connection_isolates_from_its_peers_and_returns_its_permit ... FAILED

---- ... stdout ----
panicked at src/daemon/tests/lifecycle.rs:1264:17:
connection task panicked holding its slot
panicked at src/daemon/tests/lifecycle.rs:1287:5:
a panicking connection stranded its admission slot: 3 of 4 free

test result: FAILED. 0 passed; 1 failed; 0 ignored; 0 measured; 2656 filtered out
```

Both RED injections were temporary and are **not** in this branch (`git diff` against the
committed files is empty for `definitions.rs` and `core_admission.rs`).

### GREEN — findings 1 and 4 together, unmodified branch

```
test daemon::tests::lifecycle::a_panicking_connection_isolates_from_its_peers_and_returns_its_permit ... ok
test daemon::tests::lifecycle::parked_requests_never_starve_admission_for_work_that_cannot_park ... ok
test mcp::tools::definitions::tests::maximal_tool_definitions_are_assembled_once_per_process ... ok
test mcp::tools::definitions::tests::per_session_budget_does_not_leak_through_the_cached_registry ... ok
test mcp::tools::definitions::tests::catalog_filtered_discovery_uses_the_deterministic_maximal_registry ... ok
test mcp::tools::definitions::tests::test_get_tool_definitions_with_budget ... ok
[... 34 more ...]
test result: ok. 40 passed; 0 failed; 0 ignored; 0 measured; 2617 filtered out; finished in 5.56s
```

The two deliberate panics print to stderr on the passing run too — that is the test firing
them, not a failure.

### GREEN — finding 2, the existing reader-pool lifecycle

```
running 28 tests
test reader::tests::exact_sql_health_snapshot_retires_its_reader_after_drop ... ok
test reader::tests::retirement_is_elapsed_time_modelled_and_never_retires_the_floor ... ok
test reader::tests::acquire_drives_burst_worker_retirement_on_entry ... ok
test reader::tests::cancellation_bounds_query_return_even_when_the_executor_is_still_running ... ok
test result: ok. 28 passed; 0 failed; 0 ignored; 0 measured; 271 filtered out; finished in 0.58s
```

### Build

```
cargo check -p tracedecay        --all-targets --offline   → 0 errors
cargo check -p tracedecay-agent-hosts --all-targets --offline → 0 errors
cargo fmt -- --check                                        → clean
```

Per-crate only; no workspace build and no workspace clippy, since other lanes were building
concurrently. `--offline` rather than `--locked` for the `Cargo.lock` reason below.

---

## Carried base fixes

`origin/cursor/simplify-pr421-hot-paths` does not compile. Carried, and droppable once the peer's unpushed work reaches origin:

- `fix(runtime-core): commit the missing background CPU module` and `fix(build): land the batching commit's unpushed consumers` — cherry-picked from `origin/claude/base-build-fixes` (#681).
- `fix(build): cover the root crate's base breakage` — mine, covering three sites #681 does not reach: two `HostAdmissionOutcome { .. }` initializers in the root crate missing the new `recovery` field (set to `None`, matching every other site), and the `HookCompletedReadinessDistributions` accessors from #675 (that commit's other hunks depend on unpushed peer work and are **not** carried).

I did **not** carry #681's `feat(index): plan indexing width against a memory budget`: it changes `parallelism::install` to return a `Result`, which does not compile against this base's `crates/tracedecay-semantic/src/embedding_parallelism.rs`. That file is rewritten by the peer's unpushed `bdd465ae9`.

`Cargo.lock` on the base is also ahead of the base's manifests (it lists `tracedecay-runtime-core` as a dependency of `tracedecay-code-index`, which that crate does not declare here), so `--locked` fails outright and verification used `--offline`.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -6143,6 +6143,7 @@ dependencies = [
  "tracedecay-code-extraction",
  "tracedecay-domain",
  "tracedecay-graph-db",
+ "tracedecay-runtime-core",
  "tracedecay-rusqlite-runtime",
  "tracedecay-store",
  "tree-sitter",
--- a/crates/tracedecay-agent-hosts/src/hooks/analytics/readiness.rs
+++ b/crates/tracedecay-agent-hosts/src/hooks/analytics/readiness.rs
@@ -179,6 +179,31 @@ pub struct HookCompletedReadinessDistributions {
     pub(crate) disposition_values_folded_to_unknown: u64,
 }
 
+impl HookCompletedReadinessDistributions {
+    /// Readiness counters read by the root-crate observation benchmark. The
+    /// struct moved into this crate, so its `pub(crate)` fields are no longer
+    /// reachable from there; these expose exactly the five it folds.
+    pub fn source_event(&self) -> &str {
+        &self.source_event
+    }
+
+    pub fn input_rows_received(&self) -> u64 {
+        self.input_rows_received
+    }
+
+    pub fn input_rows_processed(&self) -> u64 {
+        self.input_rows_processed
+    }
+
+    pub fn input_rows_dropped_at_cap(&self) -> u64 {
+        self.input_rows_dropped_at_cap
+    }
+
+    pub fn events_considered(&self) -> u64 {
+        self.events_considered
+    }
+}
+
 #[derive(Default)]
 struct MutableNumericSummary {
     present_count: u64,
--- a/crates/tracedecay-runtime-core/src/background_cpu.rs
+++ b/crates/tracedecay-runtime-core/src/background_cpu.rs
@@ -1,0 +1,461 @@
+//! Process-wide weighted admission for background CPU work.
+//!
+//! The authority counts active CPU units rather than owning an executor. Code
+//! indexing, semantic native threads, and session preparation can therefore
+//! use their existing execution substrates while sharing one hard process
+//! ceiling. FIFO waiter order prevents a continuously busy class from starving
+//! another class, and RAII releases capacity on success, cancellation, or
+//! unwind.
+
+use std::cell::Cell;
+use std::collections::VecDeque;
+use std::fmt;
+use std::num::NonZeroUsize;
+use std::sync::{Arc, Condvar, Mutex, OnceLock};
+
+#[derive(Debug)]
+struct BackgroundCpuWaiterV1 {
+    units: usize,
+}
+
+#[derive(Default)]
+struct BackgroundCpuStateV1 {
+    active_units: usize,
+    waiters: VecDeque<Arc<BackgroundCpuWaiterV1>>,
+}
+
+/// One process-wide background CPU budget shared across subsystems.
+pub struct ProcessBackgroundCpuV1 {
+    width: NonZeroUsize,
+    state: Mutex<BackgroundCpuStateV1>,
+    available: Condvar,
+}
+
+impl fmt::Debug for ProcessBackgroundCpuV1 {
+    fn fmt(&self, formatter: &mut fmt::Formatter<'_>) -> fmt::Result {
+        formatter
+            .debug_struct("ProcessBackgroundCpuV1")
+            .field("width", &self.width)
+            .field("active_units", &self.active_units())
+            .finish_non_exhaustive()
+    }
+}
+
+thread_local! {
+    static BACKGROUND_CPU_DEPTH: Cell<usize> = const { Cell::new(0) };
+    static BACKGROUND_CPU_UNITS: Cell<usize> = const { Cell::new(0) };
+}
+
+struct BackgroundCpuScopeV1;
+
+impl BackgroundCpuScopeV1 {
+    fn enter() -> Self {
+        BACKGROUND_CPU_DEPTH.with(|depth| depth.set(depth.get().saturating_add(1)));
+        Self
+    }
+}
+
+struct YieldedBackgroundCpuV1<'a> {
+    authority: &'a Arc<ProcessBackgroundCpuV1>,
+    units: usize,
+    depth: usize,
+}
+
+impl Drop for YieldedBackgroundCpuV1<'_> {
+    fn drop(&mut self) {
+        self.authority.admit_units(self.units);
+        BACKGROUND_CPU_UNITS.with(|units| units.set(self.units));
+        BACKGROUND_CPU_DEPTH.with(|depth| depth.set(self.depth));
+    }
+}
+
+impl Drop for BackgroundCpuScopeV1 {
+    fn drop(&mut self) {
+        BACKGROUND_CPU_DEPTH.with(|depth| {
+            let remaining = depth.get().saturating_sub(1);
+            depth.set(remaining);
+            if remaining == 0 {
+                BACKGROUND_CPU_UNITS.with(|units| units.set(0));
+            }
+        });
+    }
+}
+
+impl ProcessBackgroundCpuV1 {
+    fn new(width: NonZeroUsize) -> Self {
+        Self {
+            width,
+            state: Mutex::new(BackgroundCpuStateV1::default()),
+            available: Condvar::new(),
+        }
+    }
+
+    #[must_use]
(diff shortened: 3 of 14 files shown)


ScriptedAlchemy:
Integrated the two coherent findings into PR #421: static tools/list catalog caching and the daemon connection panic/admission regression proof. The first three commits were stale base repairs already superseded by #663/PR421, so they were intentionally not carried. Integrated commits: c662d39bd and 520df8b80.

ScriptedAlchemy:
Closing as integrated/superseded by PR #421.

PR #692, opened 2026-08-24
perf(graph): reuse the activated seal decode

ScriptedAlchemy:
Reconciles `docs/plans/tracedecay-v2/40-partitioned-generation-storage-and-zero-copy-replay.md`
against the actual tree and lands **Stage 1 (single-parse activation)** only.

## 1. Ground truth per stage

Established from the code, not from commit messages. Every stage-0 commit
subject in the plan exists **and** is an ancestor of this base branch — but
that alone proved nothing, so each was confirmed against the tree.

| Stage | Verdict | Evidence |
|---|---|---|
| **0a** — sealed replay resolves canonical-first, pool fallback | **Landed** | `with_verified_seal_from_roots` / `decode_verified_seal_from_roots` in `src/daemon/store_runtime/session_registry/code_graph_manifest.rs` probe the canonical root, fall back to the pool, and surface `Cancelled`/`DeadlineExceeded` without a second read. Commit `5ddd16271` is an ancestor. |
| **0b** — retirement retires into the replay pool instead of unlinking | **Landed, by a different mechanism than the plan text** | `crates/tracedecay-usecases/src/retention/code_index_generations.rs` quarantines with `rename(source → stage_root)` and then exposes the entry with `hard_link(staged → pool)`, collapsing `AlreadyExists` through `verify_existing_graph_replay_pool_entry`. The plan says "atomic rename into the replay pool"; the tree uses rename-to-quarantine plus hard-link-to-pool. Both are metadata-only and duplicate zero bytes, so the invariant holds — the wording does not. Commits `4b4e9e93a`, `0b788640d`, `4c258db81` are ancestors. |
| **0c** — eager staging copy deleted from the graph publish path | **Landed** | `stage_project_graph_replay_seal` and `install_project_graph_replay_seal_at` do not exist anywhere under `src/`; only unlink staging remains. Commit `b2a057733` is an ancestor. |
| **1** — single-parse activation | **Was genuinely untouched; landed in this PR** | `DaemonCodeGraphManifestProviderV1::hydrate_sealed_code_generation` unconditionally called `decode_verified_seal_from_roots`, a full re-read and re-parse, while `DaemonCodeIndexPublicationStoreV1` already held the identical decoded generation. The supplied-manifest ride-along (`1c972ac8e`) covered only first publication, exactly as the plan predicted; the recovery branches were not covered. |
| **2** — manifest + segment split (`format_revision` bump) | **Untouched** | `SEALED_GENERATION_FORMAT_REVISION_V1` is still `6`; the envelope is still one whole-generation JSON. Not attempted — it needs its own migration story and review. |
| **3** — Grafeo startup authority | **Untouched** | No manifest ↔ Grafeo verified-head validation exists; startup still replays. |

The plan's `Status: ACTIVE` line was **not** wrong, but the document carried no
record of stage state and its §Problem section reads as present tense while
0c had already removed the eager replay copy it describes. `docs(plan)` records
the reconciled state and marks the measurements as original evidence.

## 2. Stage 1

Graph activation now offers the decode it already holds to the code-graph
manifest provider, keyed by the exact shard the provider resolves bindings
under. Hydration consults that offer **only** on an exact
generation-**and**-sealed-state-digest match, and still applies every identity
guard (`project_id`, `repository`, `generation_id`) to the offered value.
Every miss falls through to the unchanged canonical-then-pool read.

Correctness notes:

- The offered generation was itself digest-verified at decode time
  (`decode_active_generation` refuses unless `sha256(bytes) == pointer.state_digest`),
  and the offer is matched on that same digest — never on generation id alone.
- The **separate** on-disk verification that `publish_verified_snapshot`
  performs (`verify_sealed_generation_source_from_roots`) is untouched, so a
  corrupt, byte-replaced, or symlinked seal is still refused at publication.
  `sealed_generation_publishes_and_republishes_without_eager_replay_payload`
  still passes unchanged.
- Callers with only a borrowed generation pass `None` and are behaviourally
  identical to before.

### RED / GREEN

Falsifiable by construction, not by timing: both seal roots handed to the
provider are empty, so hydration can only succeed by consuming the offer.

RED — offer consult disabled (`let offered = if true { None } else { … }`):

```
running 1 test
test daemon::store_runtime::session_registry::code_graph::sealed_publication_tests::offered_decode_hydrates_without_reading_the_sealed_payload_again ... FAILED

thread '…offered_decode_hydrates_without_reading_the_sealed_payload_again' panicked at
src/daemon/store_runtime/session_registry/code_graph/sealed_publication_tests.rs:574:10:
hydration reuses the offered decode without reading the seal: Unavailable { message:
"sealed code generation is unavailable for replay: No such file or directory (os error 2)" }

test result: FAILED. 0 passed; 1 failed; 0 ignored; 0 measured; 2654 filtered out; finished in 0.39s
```

GREEN — as committed:

```
running 1 test
test daemon::store_runtime::session_registry::code_graph::sealed_publication_tests::offered_decode_hydrates_without_reading_the_sealed_payload_again ... ok

test result: ok. 1 passed; 0 failed; 0 ignored; 0 measured; 2654 filtered out; finished in 0.21s
```

Whole suite for the touched area:

```
test result: ok. 19 passed; 0 failed; 0 ignored; 0 measured; 2636 filtered out; finished in 1.97s
```

The test also asserts the negative: a foreign sealed digest is never answered
from the offer.

### Verification

Per-crate only, no workspace build or clippy:

- `cargo check -p tracedecay --lib` — clean
- `cargo test -p tracedecay --lib --no-run` — clean
- `cargo test -p tracedecay --lib code_graph` — 19 passed, 0 failed
- `cargo clippy -p tracedecay --lib` — no findings in this crate
- `cargo clippy -p tracedecay-agent-hosts --lib`, `-p tracedecay-sessions --lib` — no findings in these crates

## 3. Pre-existing base-branch breakage carried here

`origin/cursor/simplify-pr421-hot-paths` did not build. Four independent
breaks, none caused by this work, all of them "committed but not actually
complete in the tree" — the same failure mode this PR was asked to
investigate:

1. **`tracedecay-runtime-core`** — `background_cpu` module never `git add`ed.
   Cherry-picked `origin/claude/fix-missing-background-cpu` (PR #676) as
   instructed; not authored here.
2. **`tracedecay-sessions`** — `64dbb4bf1` added
   `ObservationApplicationError::BatchWorkerStopped` without extending the
   ingest failure classifier. Fixed here.
3. **root `tracedecay` lib** — `HostAdmissionOutcome` gained a `recovery`
   field; two construction sites were never updated. Fixed here.
4. **root `tracedecay` lib test binary** — `ba0e72ae7` narrowed
   `HookCompletedReadinessDistributions` fields to `pub(crate)` while its
   assertions live in the root crate. Fixed here with read-only accessors
   rather than widening the fields.

Without 2–4 this crate cannot compile, so Stage 1 could not have been built or
tested at all. Each is its own commit and can be dropped or re-homed
independently.

## 4. Pre-existing test failures (NOT introduced here)

`cargo test -p tracedecay --lib session_registry::` has **9 failures on this
base branch**, in graph proxy binding, shard retirement, project mounting, and
memory relation graphs:

```
session_registry::tests::project_sessions_mount_uses_typed_enrollment_and_is_idempotent
session_registry::tests::duplicate_project_attaches_schedule_one_historical_convergence
session_registry::verified_graph_runtime_port_contract_tests::runtime_binding_is_absent_before_bind_and_rejects_double_bind
session_registry::verified_graph_runtime_port_contract_tests::session_relation_close_refusal_restores_route_and_retry_closes_exact_graph
session_registry::verified_graph_runtime_port_contract_tests::registry_drop_cancels_retained_trait_runtime_operations
session_registry::verified_graph_runtime_port_contract_tests::exact_shard_retirement_closes_retained_graph_after_root_is_absent
session_registry::verified_graph_runtime_port_contract_tests::exact_shard_retirement_leaves_sibling_live_and_remounts_fresh
session_registry::verified_graph_runtime_port_contract_tests::read_only_memory_database_rejects_a_writer_graph_runtime
session_registry::project_memory_relation_graph_contract_tests::registered_memory_relation_graph_survives_restart_and_isolates_topologies
```

Proven independent of this change: with the Stage-1 consult made inert, the
**same 9** still fail (70 passed / 9 failed either way). None of those files
reference `retain_code_graph_runtime`, `offer_decoded_code_generation`, or
`generation_handle`. Not fixed here — they are unrelated to storage
activation and deserve their own investigation.

## SCOPE DEVIATION

- **Fixed three unrelated base-branch compile breaks** (items 2–4 above),
  touching `crates/tracedecay-sessions`, `crates/tracedecay-agent-hosts`,
  `src/host_admission*`, and `src/sessions/claude_observation_benchmark`.
  Not requested. Done only because the crate under change does not compile
  without them; each is isolated in its own commit.
- **Carried the cherry-pick** of `origin/claude/fix-missing-background-cpu`
  as instructed by the coordinator.
- **Edited the plan document** beyond the `Status` line: added a stage-state
  paragraph directly under `Status`, and one sentence marking §Problem as
  historical evidence. The `Status` line itself was accurate, so correcting
  "the status line" alone would have left the drift unrecorded. No other plan
  content changed; the stale `seals.rs::stage_project_graph_replay_seal`
  reference and the 0b "atomic rename" wording were **left intact** as record.
- **Stages 2 and 3 not attempted**, as instructed.
- Stage 1 changed the signature of `retain_code_graph_runtime` (one added
  `Option` parameter); the semantic-vector caller passes `None` because it
  holds only a borrowed generation.


ScriptedAlchemy:
Integrated the sound graph slice into PR #421 as commits 7b19478b2 and 188ea35ea: activation now reuses the exact generation-and-digest-bound decoded seal, with publication/recovery coverage and the reconciled plan state. I deliberately excluded the unrelated readiness getters because they existed only to expose production fields to a test benchmark. Closing this superseded branch.

PR #698, opened 2026-08-24
perf(mcp): daemon and MCP runtime hygiene

ScriptedAlchemy:
Reopened against `codex/tracedecay-total-redesign-plan` (previous PR #690 auto-closed when `cursor/simplify-pr421-hot-paths` was deleted after #663 merged).

**Note the base has since landed two of these findings independently** — `perf(mcp): cache the static tools/list catalog` and `test(daemon): pin connection panic isolation`. Merge conflicts in the batching-commit consumer files were resolved in favour of the base, since Codex owns those now. Review this against the base to see what remains unique.

The audit verdicts remain the useful part: idle reader threads were **already fixed**; leaked `serve` processes were **not real as stated** (nothing in the workspace spawns `tracedecay serve`); and per-request daemon connections are real but deliberately **not** pooled — the preamble is the authority record and the proxy mutates it between requests, so reuse would run a request under a previous request's routing authority.

See #690 for full RED/GREEN evidence.


ScriptedAlchemy:
Closing because the actual MCP catalog cache and daemon panic-isolation work are already integrated into PR #421. The only remaining net diff adds public readiness getters solely for a root test benchmark, which would create a test-only production API.

PR #699, opened 2026-08-24
docs(store): document the flat WAL soft/hard ceiling under workstation profile

ScriptedAlchemy:
Answers the finding PR #680 deliberately left out of scope:

> `GlobalQueueProfileV1::ExplicitWorkstation` raises the global queue to 256 MiB while the WAL soft limit stays capped at 32 MiB — a workstation can admit 8x the soft limit in one drain. Possibly intentional (absorbed by the 256 MiB hard limit) but undocumented.

Carries #680's base-build-fix commit (`origin/claude/base-build-fixes-v2`, cherry-picked as `4b120569b`) since this branch is cut from `origin/cursor/simplify-pr421-hot-paths` before that fix landed there. This PR does not otherwise depend on #680 and touches none of the files it's changing (`crates/tracedecay-rusqlite-runtime/src/writer/worker/mod.rs`, `checkpoint/types.rs`, `checkpoint/tests.rs`).

## Verdict: intentional and safe, now documented (outcome 1 of 3)

No behaviour change. `AdmissionConfigV1::validate` still rejects the exact same inputs it did before.

## Why

Traced the actual mechanism in `tracedecay-rusqlite-runtime`'s `WriterCheckpointController`:

- **Soft is not a durability boundary.** Crossing `soft_limit_bytes` only asks the controller to *attempt* an opportunistic, non-blocking PASSIVE checkpoint (`WriterCheckpointController::evaluate`). Nothing blocks admission or readers on soft pressure alone.
- **That evaluation runs once per committed write batch**, not once per "drain." Batches are capped at a few MiB (`FOREGROUND_BATCH_MAX_BYTES` = 1 MiB, `BACKGROUND_BATCH_MAX_BYTES` = 4 MiB) and `run_scheduled_checkpoint` is called after every one (`writer/worker/mod.rs`). So under `ExplicitWorkstation`, soft pressure is still detected well before the WAL could plausibly leap past the hard ceiling in one uncontrolled jump — there is no single atomic write of queue-sized bytes.
- **Hard is the real ceiling**, and it is *already* profile-independent by design: `WAL_HARD_LIMIT_BYTES` (256 MiB) is deliberately equal to `WORKSTATION_GLOBAL_QUEUE_BYTES`, the largest in-flight admission volume any profile permits — not scaled *up* for `ExplicitWorkstation`, just sized once to dominate every profile's ceiling. Crossing it escalates to `CheckpointPressure::BlockGeneral`, gating new readers until a checkpoint can fully drain the WAL. That escalation path is untouched by this PR.
- **The "half of queue" and "== workstation queue" relationships are not derived from any code comment or design doc** — both constants were introduced together in one commit (`eb93351b2`) with no explanatory rationale; #680's framing was a plausible *reconstruction*, not a proven invariant. Given the mechanism above, having the soft threshold stay flat across profiles is consistent with how the hard ceiling already behaves (also flat, also not keyed off `global_queue_profile`): a workstation profile just means checkpoints keep firing at the same 32 MiB cadence relative to a larger admitted-write ceiling, i.e. *more* often relative to that ceiling, never less safely.

So the 8x relationship is real, was previously unexplained, and does not weaken durability: the hard ceiling remains a genuine, unscaled backstop, and soft pressure is sampled far more often than "once per queue drain" would suggest.

## What changed

- Doc comments on `WAL_SOFT_LIMIT_BYTES`, `WAL_HARD_LIMIT_BYTES`, and `WORKSTATION_GLOBAL_QUEUE_BYTES` (`crates/tracedecay-store/src/runtime/operation.rs`) explaining the mechanism above so the next reader doesn't have to re-derive it.
- A comment on the two ceiling checks in `AdmissionConfigV1::validate` noting they are deliberately flat (neither reads `global_queue_profile`).
- Two new tests in `crates/tracedecay-store/tests/storage_runtime_contract.rs`:
  - `wal_soft_limit_is_not_scaled_by_the_queue_profile` — pins `WAL_SOFT_LIMIT_BYTES == DEFAULT_GLOBAL_QUEUE_BYTES / 2` and `WAL_HARD_LIMIT_BYTES == WORKSTATION_GLOBAL_QUEUE_BYTES`, and proves `validate()` still rejects a soft limit scaled the same way against the workstation queue (128 MiB).
  - `explicit_workstation_permits_queue_bytes_far_above_the_flat_wal_soft_limit` — reproduces the exact 8x scenario (`ExplicitWorkstation` + max queue + default WAL budget) and asserts on the configured ratio and that the hard ceiling still dominates the largest queue under either profile.

Both assert on configured/effective values only — no timing, no on-disk file sizes.

## Test output

```
$ cargo test -p tracedecay-store --test storage_runtime_contract --locked -- wal_soft_limit_is_not_scaled_by_the_queue_profile explicit_workstation_permits_queue_bytes_far_above_the_flat_wal_soft_limit identity_and_budget_validation_fail_closed

running 3 tests
test explicit_workstation_permits_queue_bytes_far_above_the_flat_wal_soft_limit ... ok
test wal_soft_limit_is_not_scaled_by_the_queue_profile ... ok
test identity_and_budget_validation_fail_closed ... ok

test result: ok. 3 passed; 0 failed; 0 ignored; 0 measured; 14 filtered out; finished in 0.00s
```

Full crate suite:

```
$ cargo test -p tracedecay-store --locked
...
test result: ok. 17 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.00s
   Doc-tests tracedecay_store
test result: ok. 3 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.12s
```

Builds (per-crate only, as briefed): `cargo check -p tracedecay-store --all-targets --locked` and `cargo check -p tracedecay-rusqlite-runtime --all-targets --locked` both clean.

## Durability

No behaviour change. `WAL_HARD_LIMIT_BYTES` is untouched and remains the real ceiling; `AdmissionConfigV1::validate` accepts and rejects exactly the same configurations as before this PR.

## SCOPE DEVIATION

None. Scope is exactly the two files touched: doc comments + validate() comments in `crates/tracedecay-store/src/runtime/operation.rs`, and two new tests in `crates/tracedecay-store/tests/storage_runtime_contract.rs`. No production behaviour changed, no other files touched, and #680's file (`crates/tracedecay-rusqlite-runtime/src/writer/worker/mod.rs`) was not touched.


ScriptedAlchemy:
Closing after contract review. The proposed documentation overstates  as proof that a full queue drain cannot overshoot the hard WAL ceiling; it ignores pre-existing WAL bytes and the non-identity between admission payload bytes and WAL growth. I won't pin that as a safety invariant.

PR #700, opened 2026-08-24
perf(ledger): scope superseded pruning to the validated incarnation

ScriptedAlchemy:
Reopens #672, which auto-closed when `cursor/simplify-pr421-hot-paths` merged and was deleted. The base has moved a long way since: it landed its own copy of the retention rule in `f0b16f321`, so this is a genuine semantic reconciliation of two versions of the same feature, not a mechanical rebase.

## Where the two sides already agree, and stay agreed

The base's prune and this branch's prune converged on the same bounding scheme independently: same `MAX_PRUNED_ROWS_PER_COMMIT` name, same value of `256`, and the same placement — one unconditional bounded pass on every commit, immediately after `checkpoint::persist`. That agreement is kept verbatim. The redundant half of this branch's original change (making the delete bounded at all, and firing it on every commit rather than only on the transition commit) is **dropped** because the base already achieves it; re-expressing it would have produced two competing bounding schemes for one behaviour.

The bound is not a nicety. The cleanup shares the user mutation's savepoint and the process's sole SQLite writer transaction. An unbounded `DELETE` over a legacy backlog monopolises admission, and an interruption or `SQLITE_FULL` rolls the checkpoint advance back with it — so every retry re-attempts the same delete and the new epoch never commits. That is a livelock, not a slow write.

## What this branch still contributes

The candidate **scope**.

The base selects candidates by joining every checkpoint row and comparing each incarnation's raw scalar `authority_epoch`:

```sql
JOIN td_runtime_writer_checkpoint_v1 AS checkpoint
  ON checkpoint.shard_json = candidate.shard_json
 AND checkpoint.incarnation = candidate.incarnation
WHERE candidate.shard_json = ?1
  AND candidate.authority_epoch < checkpoint.authority_epoch
```

But `checkpoint::next` decodes and validates only the incarnation being advanced. Nothing has validated a *neighbouring* incarnation's row. A row whose scalar reads `999` while its watermark and receipt still encode `7` is `Corrupt` the moment anything loads it — yet the join above would read that `999` and retire that incarnation's **live** receipts. A lost idempotency receipt silently re-admits the duplicate write it exists to stop.

The delete is now scoped to the single incarnation whose checkpoint this commit just decoded, validated, and persisted, at that checkpoint's validated epoch:

```sql
WHERE candidate.shard_json = ?1
  AND candidate.incarnation = ?2
  AND candidate.authority_epoch < ?3
ORDER BY candidate.authority_epoch, candidate.idempotency_key
LIMIT ?4
```

A neighbour's superseded records are retired by that incarnation's own commits, under its own validated checkpoint. Ordering drops the two now-constant leading primary-key columns, which leaves the same key order over the same candidate set, so each bounded pass stays deterministic and repeated commits still converge.

Pruning remains keyed on authority supersession only — never on age, and **never on superseded incarnation**. `StoreIncarnationV1` is not monotonic: non-daemon processes derive it from random process-run bytes while the daemon uses a small counter, so a CLI attach can outrank a later daemon one. Pruning by it would have reclaimed far more and is unsound regardless.

## Conflict resolution

| file | resolution |
| --- | --- |
| `ledger/prune.rs` | Kept the base's bound and per-commit placement; replaced the checkpoint join with the validated `(incarnation, epoch)` pair the caller already holds; exposed `MAX_PRUNED_ROWS_PER_COMMIT` so the tests assert against the constant instead of a second copy of `256`. |
| `ledger/commit.rs` | The call passes the persisted checkpoint instead of the bare shard id. Still unconditional, so a backlog left by an earlier bounded pass still converges. |
| `ledger/tests.rs` | The base's two bounding tests are the same two tests this branch renamed and re-derived from the constant, so the base copies are dropped rather than duplicated under both names. Both properties it covered are retained. |

The whole delta against the base is those three files and nothing else.

## Test coverage

Both of the base's properties are preserved, and one is added:

- `one_foreground_commit_prunes_at_most_one_bounded_batch` — bounded progress (base's property, now derived from the constant)
- `later_commits_drain_the_remaining_superseded_backlog` — convergence (base's property)
- `a_corrupt_neighbouring_checkpoint_cannot_retire_its_receipts` — **new**: an unvalidated neighbouring checkpoint cannot authorise deleting that incarnation's receipts
- `a_pruned_record_fails_closed_rather_than_admitting_a_duplicate` — the safety property bounding the whole rule
- `records_are_retained_while_their_authority_still_stands`
- `advancing_authority_prunes_only_the_superseded_records`

Assertions are on row counts and outcomes; none is on elapsed time.

## Verification

```
$ cargo check -p tracedecay-rusqlite-runtime --all-targets --locked
    Finished `dev` profile [optimized + debuginfo] target(s) in 1m 39s

$ cargo test -p tracedecay-rusqlite-runtime --locked --lib ledger::
running 10 tests
test ledger::tests::a_corrupt_neighbouring_checkpoint_cannot_retire_its_receipts ... ok
test ledger::tests::a_pruned_record_fails_closed_rather_than_admitting_a_duplicate ... ok
test ledger::tests::records_are_retained_while_their_authority_still_stands ... ok
test ledger::tests::advancing_authority_prunes_only_the_superseded_records ... ok
test ledger::tests::one_foreground_commit_prunes_at_most_one_bounded_batch ... ok
test ledger::tests::later_commits_drain_the_remaining_superseded_backlog ... ok

test result: ok. 10 passed; 0 failed; 0 ignored; 0 measured; 318 filtered out; finished in 0.14s
```

Full crate suite, every target: `328 passed` in the lib plus 19 integration binaries, **0 failed** anywhere.

## Related

Sibling branch `claude/perf-ledger-row-width` (#683) is **not** being re-proposed. It was reverted from the base in `5f1c70bc1` for two blockers that still stand — its migration copies the full legacy table in the first foreground writer transaction, which is the exact unbounded-foreground-work failure this PR's bound exists to prevent, and it drops `original_receipt_json` after validating only `commit_sequence` rather than all legacy receipt invariants. That work stays closed pending a bounded/background migration with full corruption validation.


ScriptedAlchemy:
Integrated the validated-incarnation pruning fix into PR #421 as commit 84df647f5. Pruning remains bounded to 256 rows per foreground commit, now derives its candidate scope only from the exact decoded and persisted checkpoint, and retains the corrupt-neighbor regression coverage. Closing this superseded branch.

PR #701, opened 2026-08-24
fix(tests): clear leftover test/build debt on simplify-pr421-hot-paths

ScriptedAlchemy:
## Summary

Rebased onto the current `codex/tracedecay-total-redesign-plan` head and removed build repairs that the base had already absorbed.

The remaining changes are intentionally small:

- Delete a stale test-only narration comment after the route-metadata wrapper was already inlined upstream.
- Remove dependencies that have no production or test callers.
- Keep `fs2` and document its external fixture caller for `cargo-shear`.
- Remove a stale self-dependency suppression from `tracedecay-search-eval`.

## Rebase resolution

The old branch was 35 commits behind its base. Patch and range-diff review showed:

- The missing background CPU module and admission recovery fixes were already incorporated by the base.
- The base also resolved the readiness benchmark boundary without widening crate-private fields, so the obsolete visibility commit was dropped.
- Automated review findings about installing and sharing the CPU authority are addressed by the base's canonical `tracedecay-code-index::parallelism` integration.

## Verification

- `git diff --check origin/codex/tracedecay-total-redesign-plan...HEAD`
- Branch is clean and contains zero base-only commits.
- Focused Cargo verification is delegated to the refreshed PR checks while other worktrees are running active Cargo jobs.


ScriptedAlchemy:
Closing after selective reconciliation. The safe route-metadata cleanup is already semantically present on PR #421. The dependency-removal commit is stale against the current CLI: open is used by dashboard launch, regex is required by the Hotpath configuration path and feature-off tests, and the shared integration fixtures still require their dev dependencies. The branch also predates the current base and carries repair commits already integrated elsewhere.

PR #704, opened 2026-08-24
perf(lcm): measure the GC batching at the executor boundary, and fix the grep fetch budget

ScriptedAlchemy:
Reopens #673, which auto-closed when `cursor/simplify-pr421-hot-paths` was merged and deleted.

Rebased onto the base by merge. Since #673 was opened, the base landed its own `perf(lcm): batch per-row GC and delete-recovery SQL` (`cff4821cb`) covering the same four sites, so **the production halves of this branch are dropped**: `gc.rs`, `payload.rs`, and `payload/delete_recovery.rs` take the base verbatim.

The base's shape is the better one and is kept: it hoists the per-payload `DELETE FROM lcm_gc_marks` out of the prepared delete so the reap loop clears the whole batch in one bounded statement, while `delete_external_payload_in_transaction` still clears a single payload's own mark.

## What is left after dropping the redundant part

| what | why the base does not already have it |
|---|---|
| `query.rs`, `query/grep.rs` | the `grep` fetch-budget defect (below) — the base never touched this file |
| `dag.rs` | `expand_query` still expanded explicit node ids one at a time; summary hydration now resolves the whole requested page in one pass |
| `gc/tests.rs` | the gates are re-expressed to count **work**, not statement text |

### The fetch-budget defect

`grep` already set the page budget, and the `LIKE` fallback applied `compute_search_fetch_limit` a **second** time on top of it, so `limit: 2` became a SQL `LIMIT` of 80 and 72 of those rows were discarded verbatim by the truncation at the end of the same function. The FTS siblings bind the caller's limit straight through, so the rows a page actually cost depended on which index path SQLite happened to take — 8 via FTS, 80 via `LIKE`, for the identical request. The caller now owns that budget outright and the leaf spends it once; the dead heuristic goes with it.

### The test gates

Source-shape and string scans are not acceptance evidence in this repo, so nothing here inspects statement text. A `WorkCounter` records round trips and rows visited at the `Executor` boundary and never retains SQL. Batching is asserted as a **marginal across two set sizes**, which is immune to query rewrites: reap 2 payloads and 8, and compare.

That gate is what caught the base's improvement. On this branch's own structure the per-payload marginal was 5 round trips; on the base's it is **4**, because the GC-mark delete left the loop. The constant was re-measured, not assumed, and both halves were mutation-checked:

- un-hoisting `ReferencedClosureCache` into the loop → marginal 5, gate fails
- restoring the per-row `DELETE FROM lcm_gc_marks` → marginal 5, gate fails

The residual-placeholder gate is likewise measured as rows visited staying flat as decoy rows that merely name the ref are added, rather than by counting `LIKE` terms.

One assertion from #673 stays deleted as wrong on the merits: it claimed a batched GC-mark delete where a single-payload delete must clear its own mark. The base's two behavioural assertions covering exactly that — ordinary deletion clears the mark, rollback restores it — are kept.

## Verification

`cargo check -p tracedecay-sessions --all-targets --locked` and `cargo check -p tracedecay-cli --all-targets --locked` both clean.

```
cargo test -p tracedecay-sessions --locked --lib runtime::lcm
test result: ok. 118 passed; 0 failed; 0 ignored; 0 measured; 591 filtered out; finished in 1.61s
```

10 failures elsewhere in the crate (`runtime::codex::tests::recent_first_discovery_tests::*`, `runtime::ingest::scheduler::tests::profile_codex_frontier_converges_beyond_budget_and_survives_restart`) are pre-existing on the base: this branch's tree is byte-identical to the base under `runtime/codex/` and `runtime/ingest/scheduler.rs`.


ScriptedAlchemy:
Integrated the reviewed net-new LCM slice into PR #421 as commit d4c67d49f. The final base keeps its existing batched GC deletion, adds executor-bound work counters, batches explicit summary expansion, and removes the duplicate LIKE fallback fetch-budget multiplication. Closing the superseded branch.

PR #708, opened 2026-08-24
perf(ledger): scope pruning to the validated checkpoint

ScriptedAlchemy:
Reopened against `codex/tracedecay-total-redesign-plan-reopened` (prior PRs auto-closed when #421 merged and master was reverted).

Base already carries the 256-row bound; this keeps that verbatim and adds only what remained: the delete is scoped to the **validated** `(incarnation, epoch)` pair instead of joining every checkpoint row and trusting each one's raw scalar `authority_epoch`. `checkpoint::next` validates only the incarnation it advances, so a corrupt neighbour reading 999 while its watermark encodes 7 could retire that incarnation's live receipts — and a lost receipt silently re-admits the duplicate write it exists to stop.

328 tests pass, 0 failures. Superseded-incarnation pruning stays out: `StoreIncarnationV1` is non-monotonic.


ScriptedAlchemy:
Closing as an exact no-op against PR 707. GitHub reports zero changed files, and the validated-incarnation pruning is already carried by 84df647f5 in the restored delivery tree.

PR #709, opened 2026-08-24
feat(github): cut GitHub API cost across three read paths

ScriptedAlchemy:

Three items: a `gh auth token` read-only credential authority (60/hr anonymous -> 5000/hr), a pre-flight rate-limit gate that costs **no** provider request (verdict comes from `x-ratelimit-*` headers earlier responses already carried), and CI-path ETag caching reusing the existing review-cursor `If-None-Match` mechanism.

Token bytes live only in `Zeroizing` containers, nothing holding one derives `Debug` or Serde, stderr/stdin are `Stdio::null()`, and under `cfg(test)` the default source is a stub so no test can spawn `gh`. Every failure path degrades to anonymous.


ScriptedAlchemy:
Closing after review. The branch still has P1 credential-isolation and wrong-host token risks plus quota-keying, fallback, retention, and cache-accounting defects. None of this unsafe slice is being reconciled into PR 707.

PR #710, opened 2026-08-24
test(writer): pin the batch-to-frame coalescing ratio

ScriptedAlchemy:

**No production change.** The 14:1 batches-per-frame ratio was already fixed twice over, so this pins it instead of reimplementing it: `build_batches` had no assertion in this crate, and every existing test constrained how many requests get *submitted*, not how they get *grouped*.

Also documents a finding: `rusqlite.begin_immediate` is **not** the writer's begin and does not measure a SQLite statement — it is dominated by blocking on the writer actor's reply channel. Widening the writer window makes that metric worse.


ScriptedAlchemy:
Closing after review. The added build_batches fixture does not exercise worker ingress or dwell and duplicates the existing three batching tests. No production change or unique journey remains for PR 707.

PR #711, opened 2026-08-24
docs(store): document the flat WAL soft/hard ceiling

ScriptedAlchemy:

**No behaviour change.** Investigated whether the workstation profile's 256 MiB queue against a flat 32 MiB WAL soft limit is a sizing bug. It is not: crossing the soft limit only triggers an opportunistic non-blocking PASSIVE checkpoint, and that check runs once per committed batch with batches capped at 1-4 MiB — so soft pressure is sampled far more often than once per queue drain.

git history shows all four constants were introduced together with zero rationale, so the "half of queue" framing was an after-the-fact reconstruction, not an invariant. Now pinned by two tests.


ScriptedAlchemy:
Closing after review. The WAL text overstates pressure thresholds as growth bounds, the global queue budget is not enforced across shards, and the readiness accessors have no production caller. No safe slice remains.

PR #712, opened 2026-08-24
perf(mcp): daemon and MCP runtime hygiene

ScriptedAlchemy:
Reopened against `codex/tracedecay-total-redesign-plan-reopened`. Five-finding audit. `tools/list` catalog rebuild and connection panic isolation may already be in base — review against it. Verdicts: idle reader threads **already fixed**; leaked `serve` processes **not real as stated** (nothing in the workspace spawns `tracedecay serve`); per-request daemon connections real but deliberately **not** pooled, because the preamble is the authority record and the proxy mutates it between requests.


ScriptedAlchemy:
Closing as integrated or rejected. The MCP catalog cache and daemon panic-isolation patches are already present in PR 707 as c662d39bd and 520df8b80 with identical patch IDs; the only remaining diff is unused public readiness accessors.

PR #713, opened 2026-08-24
test(exact-sql): pin the split begin/lock spans

ScriptedAlchemy:
Reopened against `codex/tracedecay-total-redesign-plan-reopened`. **The span split already existed** in `perf(exact-sql): separate dispatch and lock timing` — my brief was stale. This adds only the missing falsifiable test proving `rusqlite.exact_sql.begin_immediate` and `rusqlite.exact_sql.write_lock` are recorded and distinct, driven through a real hotpath guard and asserted from parsed JSON, never on elapsed time.


ScriptedAlchemy:
Closing after review. The feature-gated test is not exercised in PR CI, inherits ambient HOTPATH_REPORT state, and adds benchmark-only public accessors. The production span split already exists in PR 707.

PR #714, opened 2026-08-24
chore: clear test leftovers and unused deps

ScriptedAlchemy:
Reopened against `codex/tracedecay-total-redesign-plan-reopened`. Deleted a test-only hooks wrapper following the two existing house-pattern commits. `cargo shear` went 7 errors -> 0; `fs2` was **restored** after finding it is used via a cross-crate `#[path]` include, with a documented ignore rather than blanket suppression. Two premises were stale: the knowingly-broken privacy test had already been rewritten, and the mcp_suite dead-code report was test-layout drift.


ScriptedAlchemy:
Closing after review. This removes regex required by default tracedecay-cli test builds, and the earlier dependency-removal reconciliation was stale against live CLI and integration callers.

PR #716, opened 2026-08-24
fix(index): stop one bad file from blocking a whole index

ScriptedAlchemy:
Reopened against `codex/tracedecay-total-redesign-plan-reopened`. **Note: an earlier version of this was reverted as "ineffective panic quarantine" and that criticism was correct** — the guard was declared but never reached from the worker. See `claude/panic-guard-wiring` (the follow-up) for the actual wiring. This branch carries the bad-file half only. Close it if the base already covers that.


ScriptedAlchemy:
Closing as incomplete and superseded. Its panic guard is not wired into the worker. The valid minimal empty-name identity behavior and focused tests were selectively applied for PR 707; none of the dead panic machinery was taken.

PR #717, opened 2026-08-24
fix(index): wire the panic guard into the reconcile worker

ScriptedAlchemy:

**This answers the "ineffective panic quarantine" revert directly.** That revert was right: `ReconcilePanicGuardV1` was declared but never reached from the worker, so a panicking unit was still retried on every wake — the original symptom was the same panic repeating 114 times.

This connects `record_panic`, `suppresses_pass` and `record_progress` to the registry worker, with a 371-line isolation test. Re-proposing reverted work is the owner's call — close this if you disagree with the direction rather than the implementation.

### DIFF
--- a/crates/tracedecay-domain/src/code_intelligence/graph.rs
+++ b/crates/tracedecay-domain/src/code_intelligence/graph.rs
@@ -896,11 +896,19 @@ pub struct CodeBlock {
 ///
 /// The ID format is `"kind:32hexchars"` where the hex portion is the first 32
 /// characters of the SHA-256 hash of the input components.
+///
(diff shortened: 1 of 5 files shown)


ScriptedAlchemy:
Closing after selective reconciliation for PR 707. Retained only the graph.rs behavior that accepts empty extractor-derived names and its two deterministic identity tests. RED reproduced both debug-assert panics; GREEN passed 2 of 2 in isolated target /tmp/tracedecay-empty-name-tdd.SbLUtK. The reviewed panic and capacity machinery was not taken.

PR #718, opened 2026-08-24
fix(daemon): stop capping how many projects can enrol

ScriptedAlchemy:
Reopened against `codex/tracedecay-total-redesign-plan-reopened` **for visibility, not for merge as-is.**

A prior version was reverted as "unsafe unmetered capacity expansion" and that bar is right: raising ceilings without resident-memory admission trades a typed refusal for possible OOM.

What this branch has that is worth keeping is the **arithmetic and the structural finding**: two graph owners per mounted project plus two profile-wide is exactly 8, so three projects fill the budget; and `reserve_capacity_eviction` can never reclaim a live project because its owner attachment is held for the life of the mount, so nothing bounds residency today.

The owner's requirement is "no cap on projects that we can enroll". The version that satisfies both is resident-memory-admitted elasticity, not this constant bump.


ScriptedAlchemy:
Closing as evidence-only and unsafe to merge. Raising retained project ceilings without resident-memory admission or idle-owner eviction trades a typed refusal for OOM risk, and the arithmetic-only guards do not exercise daemon mount behavior.

PR #707, opened 2026-08-24
feat: TraceDecay V2 delivery (PR8–PR13) + daemon performance, memory, and reliability overhaul

ScriptedAlchemy:
## Order 25 — perf routing (2026-09-16; #707 held)

| Finding | Ticket / draft | Owner lane | Status |
|---|---|---|---|
| Deferred query mount Notify (r4023832024) | [#1391](https://github.com/ScriptedAlchemy/tracedecay/issues/1391) | [fcb451b2](fcb451b2-c0a2-4a91-9fdf-01432ebc95c3) `/fast/tmp/td-1391-deferred-mount` | **assigned** |
| Candidate-batch cancel (r4023833605) | [#1392](https://github.com/ScriptedAlchemy/tracedecay/issues/1392) / [#1372](https://github.com/ScriptedAlchemy/tracedecay/issues/1372) / [#1385](https://github.com/ScriptedAlchemy/tracedecay/pull/1385) | [b0c507d5](b0c507d5-879e-482c-a681-143262f42bc0) `/fast/tmp/td-1385-token-cutover` | **assigned expand** |
| rebuild/freshness clones (r4023833146) | [#1368](https://github.com/ScriptedAlchemy/tracedecay/pull/1368) + [#1389](https://github.com/ScriptedAlchemy/tracedecay/pull/1389) | existing | parked mapping |
| Discovery queue merge (r4023846345) | — | — | **DISMISS**; git-discovery deadline/single-flight only |
| fastembed residue (r4023833721) | [#1393](https://github.com/ScriptedAlchemy/tracedecay/issues/1393) | qualification owner | assigned ticket |
| Lexical / ast-grep / grep alloc hypotheses | measure then split PRs | [61f929fb](61f929fb-6649-484e-8b2b-d7c003d2b096) `/fast/tmp/td-707-perf-profiles` | **measure-first**; report tables; no PR from static counts |
| New ETag/cache-header subsystem | — | — | **NON-ACTION** (SSE path via #1389) |
| [#1388](https://github.com/ScriptedAlchemy/tracedecay/pull/1388) → [#1395](https://github.com/ScriptedAlchemy/tracedecay/issues/1395) direct-seal admit | [#1395](https://github.com/ScriptedAlchemy/tracedecay/issues/1395) | [3ec8d342](3ec8d342-6982-4421-bc52-359a5b68fc4a) `/fast/tmp/td-1394-seal-admit` | **replacement owner**; keep graph-db gate; measure ≥5 runs |

Perf tables

[...]

PR #751, opened 2026-08-29
test(scheduler): retain canonical graph-off store path

BleedingDev:
## Summary

- retain the canonical scoped store used to seed the graph-off fixture
- assert and remove the witness from that exact scope
- reuse the same scope for artifact inspection

## Why

On macOS, the fixture was seeded through `/private/var/...` but later recomputed its store key from the aliased `/var/...` path. It then tried to remove a witness from an unseeded sibling scope and failed with ENOENT before exercising scheduler behavior.

No production code or behavioral assertions change.

## Verification

- reproduced the original witness-removal failure
- corrected focused test passes
- all four `graph_off_` scheduler tests pass
- aggregate clean build with the companion branches: 4 passed
- Rustfmt and diff checks pass

Supports ScriptedAlchemy/tracedecay#707.

### DIFF
--- a/crates/tracedecay-code-index-runtime/src/code_index_scheduler/tests.rs
+++ b/crates/tracedecay-code-index-runtime/src/code_index_scheduler/tests.rs
@@ -10139,7 +10139,7 @@ async fn graph_off_overflow_preserves_text_owner_progress_without_full_decode() 
     let (scope, privacy_domain) = {
         let mut scheduler = scheduler(
             &fixture,
-            scoped_store,
+            scoped_store.clone(),
             Arc::new(SharedCodeIndexBytePoolV1::default()),
         );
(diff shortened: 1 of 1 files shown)


ScriptedAlchemy:
Absorbed into `codex/tracedecay-total-redesign-plan-reopened` via merge commit 70a731212d (pushed at 56094ea680). The fixture change merged cleanly onto the reworked scheduler tests; verification on the merged tree: `cargo check --offline --no-default-features -p tracedecay --lib` green, and all four `graph_off_` scheduler tests pass (4 passed / 0 failed). A pre-existing test-lane compile break in `git_transactions/tests.rs` (bare `GitIndexTransactionStoreError::Unavailable` after the variant gained a reason string) was fixed alongside in 1b64d5f597 so the suite could run.

PR #770, opened 2026-08-31
fix(daemon-control): stabilize macOS service resolution

BleedingDev:
## Summary

- continue PATH lookup past non-executable `launchctl` or `id` shadows
- preserve strict rejection for explicitly injected absolute program paths
- fix the macOS test-module `ServiceRunner` compile failure
- isolate service-unit advice tests from the developer's real LaunchAgents directory

## Why

The pinned #707 head does not compile its daemon-control tests on macOS because an ungated test uses a Linux-gated import. After fixing that blocker, one fixture reads the host's real launchd state. PATH lookup also incorrectly treats an unusable early entry as terminal.

Based on #707 commit `e8e30fb323ebdfb53380b682ef6aba5a5c384979`.

## Verification

- `cargo test -p tracedecay-daemon-control --lib --locked`
- 81 passed, 0 failed
- deterministic PATH tests for both `launchctl` and `id`
- explicit-path rejection test


ScriptedAlchemy:
Merged into PR #707's integration branch at [`bb812b05920bae1dae824e0b378ee111b9fc00e0`](https://github.com/ScriptedAlchemy/tracedecay/commit/bb812b05920bae1dae824e0b378ee111b9fc00e0). Validation: 106 daemon-control tests passed and the combined feature-off gate passed.

PR #771, opened 2026-08-31
fix(sdk): refresh inline project selector schemas

BleedingDev:
## Summary

- refresh the generated TypeScript schemas for all 13 selector-aware retained requests
- emit the canonical closed nullable `project_selector` object inline
- leave handwritten SDK types and runtime behavior unchanged

## Why

PR #707 changed the canonical selector schema, but the committed TypeScript artifact was stale. The SDK packages job fails its deterministic codegen check on that drift.

Based on #707 commit `e8e30fb323ebdfb53380b682ef6aba5a5c384979`.

## Verification

- `scripts/check-sdk-codegen.sh`
- codegen tests: 11 passed
- TypeScript typecheck
- TypeScript tests: 34 passed
- Rust SDK tests: 40 passed, 3 expected ignored


ScriptedAlchemy:
Superseded on the current `codex/tracedecay-total-redesign-plan-reopened` tip.

This PR changes only `sdks/typescript/src/operations.ts`. The current base and PR head resolve that file to the identical Git blob:

`af929d1473a6bd99177f9a324e18f051e922fbc9`

The generated selector schema payload is therefore already byte-for-byte present on the base, with no remaining substantive change to merge.

Issue #837, opened 2026-09-04
Semantic vector publication times out under the fixed 30-second graph deadline


HashemKhalifa:
## tracedecay version

0.1.0-beta.37+1576b68e3df520e2113170518a98de97a31fdda8, built from source on macOS.

The same publication boundary is still present at PR #707 commit a5002ff6a85db7f02d80bfac383005b535deaa22.

## What happened?

After the stale-stage recovery covered by #792 allowed a retained semantic vector generation to resume, staging advanced to 161,492 of 168,659 chunks. Final publication then failed in `publish_generation` with a graph operation deadline error, and semantic runtime returned to degraded/runtime_failure.

This is downstream of #792. Commit a5002ff fixes stale occupancy at `begin_generation`, but `publish_generation_records` still creates one authority with `Instant::now() + GRAPH_OPERATION_DEADLINE` where `GRAPH_OPERATION_DEADLINE` is 30 seconds. That same authority is reused across both:

1. `prepare_publication_from_staged_native`
2. `publish_ready_stage`

The prepare phase calls `recovered_generation_digest_from_database`, which decodes and hashes the full staged generation. Publication then performs additional whole-generation work under the remaining portion of the same deadline.

Existing repository scaling measurements for the native staging proof are approximately linear:

- 2,048 chunks: 497 ms
- 4,096 chunks: 1,021 ms
- 8,192 chunks: 2,150 ms
- 22,016 chunks: 5,704 ms
- approximately 243 to 259 microseconds per chunk

At 168,659 chunks, the digest proof alone projects to roughly 41 to 44 seconds, before sealed publication work. The fixed 30-second deadline therefore cannot admit this valid corpus.

## Steps to reproduce

1. Run the PR #707 semantic recovery stack with a retained staged generation of roughly 168k chunks.
2. Allow the retained generation to resume and finish staging.
3. Drive final semantic publication.
4. Observe `publish_generation` terminate at the graph authority deadline and semantic runtime become degraded.

## Error output

```shell
semantic publication phase: publish_generation
graph authority check: deadline exceeded
runtime state: degraded
runtime reason: runtime_failure
active generation: null
```

## Expected behavior

Daemon-owned whole-generation verification and publication should be bounded by lifecycle cancellation and a background-operation budget appropriate to the existing repository policy. Interactive graph operations should retain their short deadline.

Each independently corpus-scaled publication phase should receive a fresh authority so one full-generation pass cannot consume the budget of the next pass.

[...]

Issue #792, opened 2026-09-01
Semantic vector publication deterministically fails at first commit batch with store_concurrent_mutation


HashemKhalifa:
## Observed

On PR #707 current head 53b7999fb92fa98cb23f10508d94932e022264eb, shipping as 0.1.0-beta.37+53b7999fb92fa98cb23f10508d94932e022264eb.dirty, semantic indexing for an existing registered project fails deterministically at the first committed batch.

The model artifact is installed and verified. Runtime initially reports indexing with 168,391 total units, advances to 289 completed units, then transitions to:

- lifecycle state: failed
- detail: semantic runtime publication commit_batch failed: store_concurrent_mutation
- retryable: true
- runtime state: degraded
- runtime reason: runtime_failure
- active generation: null

A supported daemon restart repeats the same 0 -> 289 -> failed transition. The reproduction occurred after the earlier disk-full condition had been cleared; no new disk-full event accompanied this failure.

## Related graph state

The latest #765 recovery stack improves the code-index side: the daemon serves an observed stale generation with 63,695 symbols and 82,738 edges while the replacement generation refreshes. However, semantic-vector publication still conflicts independently, so the #765 same-store/reseat fixes do not cover this commit_batch path.

## Expected

A retryable concurrent mutation during semantic batch publication should be serialized or retried from the durable checkpoint, allowing vector indexing to progress beyond the first batch. Runtime and lifecycle should expose the same terminal state if retries are exhausted.

## Actual

The first conflict immediately terminates semantic indexing. Restarting repeats the same failure, leaving calibration and strict semantic retrieval unavailable.

## Reproduction outline

1. Run the daemon from PR #707 head 53b7999fb.
2. Open a registered project whose code generation contains 168,391 semantic units.
3. Ensure JinaEmbeddingsV2BaseCode is installed with auto-download enabled.
4. Observe semantic runtime move from installed to indexing.
5. At 289 completed units, observe commit_batch/store_concurrent_mutation and degraded/runtime_failure.
6. Restart the daemon and observe the same transition again.

Related: #765, but this report is specifically the semantic vector generation commit_batch path rather than code graph activation/seating.

HashemKhalifa:
Additional live reproduction evidence on beta.37 with PR #707 commit 159989b2575c76c533257ff921a934b61f065c9c:

- The code generation is complete/current and the graph is ready before the terminal semantic result.
[...]

Issue #753, opened 2026-08-30
Semantic activation has no shipped operator journey; model 'loading' state persists 16+ min past load_deadline_ms with no typed failure; day-old eval client gets connection reset


ScriptedAlchemy:
## Context

Attempted the first real-operator-profile verification of semantic retrieval at full scale (project `/fast/projects/tracedecay`, ~4,790 files / ~226k chunks, model JinaEmbeddingsV2BaseCode installed and verified on disk). Three distinct gaps surfaced; filing together because they gate the same journey.

## 1. No shipped surface can activate semantic retrieval

The operator profile has `semantic.runtime.v1 → active_profile: null` (registry default). Activation requires a native evaluated-profile publication (`semantic_evaluate_and_publish`, profile `hybrid-conservative`) followed by a CAS `configuration_set` of `active_profile` — but the only clients of that invocation are the dev binaries `tracedecay-search-eval-direct` / `tracedecay-search-eval`, which are not shipped (`~/.local/bin/tracedecay` has no subcommand for it, no MCP tool exposes it, 246-tool catalog checked). A real operator cannot reach semantic-ready through any product journey; strict_semantic search truthfully reports `unavailable` forever. (The invocation itself is additionally broken by #752.)

## 2. `loading` lifecycle state ran 16+ minutes with `load_deadline_ms=30000`, no typed transition

The effective config pins `resources.load_deadline_ms: 30000`. During the one observed projection attempt (02:41–02:57 UTC, quiet host: load1 ≈ 20), `tracedecay tool runtime` reported `semantic_runtime.state = "loading"` continuously for ≥16 minutes while daemon RSS climbed from 11.4 GB to 18.5+ GB, until a kernel OOM kill ended the process. Either the load deadline is not enforced on this path, or the surfaced lifecycle state was stale — both are progress-reporting truthfulness violations (the operator cannot distinguish a hung load from a working one, and `Indexing{completed_units,total_units}` was never reached so no progress was ever visible).

Also note: after the aborted attempt, subsequent sealed generations (04:13 `fresh|ready`, 04:27, 05:10) never re-triggered projection within the same daemon lifetime — whether that is the publish-failure memo backoff or the deferred-activation retry being single-shot, the runtime stayed silently at `installed` with no typed record of why nothing was scheduled.

## 3. One-day protocol drift breaks the eval client with a raw connection reset

[...]

Issue #852, opened 2026-09-04
perf(code-index): bound pre-seat text projection memory below admission budget


ScriptedAlchemy:
## Defect

The preserved operator profile on exact pushed PR #707 commit `250c5e01b` reached **26.31 GiB RSS before any project graph engine opened, sealed-store publication occurred, or a graph generation seated**. This is a pre-seat text/code-index projection working-set defect, distinct from #799's retained Grafeo-engine peak and #830's post-publication release boundary.

Observed live progression for generation 29:

- 1.82 → 8.58 → 13.15 → 18.97 → 22.36 → 23.74 → 26.31 GiB RSS
- graph file remained 2,361,815,040 bytes with no WAL
- shutdown cancelled inside the published text loop; the graph gate is evaluated only afterward
- profile was preserved and the daemon was stopped through its normal lifecycle before OOM

A prior Hotpath allocation capture on the same ~4–5k-file shape attributes the largest overlapping owners to:

- `code_index.generation.decode.bundle`: 7.4 GB
- `publish.segment_encode`: 10.8 GB
- `decode.segment`: 1.7 GB
- `restore.file_admit`: 1.4 GB
- incremental materialization: 1.1 GB
- `load_active` deep clone: 533 MB

The decoded active generation, restored file artifacts/evidence, incremental candidate, whole-generation encoder payload/segment/index marker buffers, and one text-source batch can coexist. The current 1.5 GiB ledger bounds only batch/cache estimates; it does not account for the decoded generation, encoder, SQLite/allocator overhead, or arena retention. `0a3642dd4` only changes scheduling throughput and `63777d1d0` only improves diagnostics; neither bounds these owners.

## Required outcome

- Decompose and eliminate overlapping whole-generation owners across decode, active-generation loading, incremental materialization, partitioned sealing, and text publication.
- Preserve exact generation/source identity, content-addressed validation, cancellation, rollback, and the existing admission ceiling.
- Make the actual end-to-end working set observable and admission-accounted; do not raise limits or mask the failure with stale RSS telemetry.
- Keep #800's delta-proportional persistence work distinct, but reuse it where it genuinely deletes whole-generation work.

## Acceptance evidence

- A scale regression that includes active-generation decode, one-file incremental materialization, segment sealing, and text publication—not an isolated batch estimator.
- Hotpath/allocator evidence showing the former whole-generation buffers no longer overlap.
[...]

PR #882, opened 2026-09-05
fix(code-index): preserve and bound live migration replay

HashemKhalifa:
## Summary

Retained pre-paging evidence manifests can now be authenticated and restored without allocating owned identity maps for every symbol and chunk in the corpus. Encountered legacy markers resolve directly against canonical descriptors and decoded files. The full manifest parser and streaming retention reader share bounded layout validation while retaining separate authentication responsibilities; missing `pages` identifies historical format, and explicit `null` is rejected.

Semantic generation restart and publication use fresh, finite 15-minute background authority per phase. Tests cover the exact budget and lifecycle cancellation during snapshot refresh. The existing lexical cursor preservation and adaptive subdivision of refused page batches remain included.

## Validation of the follow-up fixes

- Codec: 10 unit tests and 2 integration tests passed.
- Semantic vector generations: 31 tests passed, including cancellation and fresh per-phase authority.
- Formatting and diff whitespace checks passed.
- Two scoped commits: `74da6279f` (codec) and `316e8e73f` (semantic authority).

## Earlier live evidence

The earlier revision resumed a retained generation, completed 3,863 source files, and remounted the verified generation after restart without replay. This is historical evidence; matching-runtime and BOT CLI/MCP verification of the follow-up revision is still in progress.

## Acceptance gaps

- A genuine historical legacy-fixture peak-RSS measurement remains outstanding. Removing pages from a current-writer fixture is not valid historical evidence.
- Approved, green fork workflows remain outstanding; a skipped integrity check is not CI acceptance.
- Strict semantic profile activation and BOT CLI/MCP usability on the follow-up runtime are not yet claimed complete.

Follow-up validation for #707 and #792; shared validation addresses #886. Related: #753 and #837.


ScriptedAlchemy:
Architecture review of `a239f44418f61293eaba19714827775123685c91` in the context of #707.

Filed **#886** for a concrete simplification exposed by this patch: `parse_partitioned_manifest` and `partitioned_segment_identities_from_reader` independently validate the same evidence descriptor invariants, so this migration now adds its legacy/paged decision in both places. Share a small private descriptor validator, not the whole reader. Retention's caller-authenticated metadata projection must remain distinct from serving authentication and must not start decoding the full generation.

The existing inline review already covers the two more serious concerns, so I am not filing duplicate issues for them: the corpus-wide owned identity maps during legacy restore, and replacing the interactive deadline with a very long background budget without a cost/progress proof. Keep the memory acceptance attached to #811/#852 and the semantic lifecycle acceptance attached to #753/#837. Successful completion on the reported 3,863-file profile is useful evidence, but it does not by itself bound peak memory or prove prompt cancellation of a stalled restart.

For the compatibility acceptance in #886, retain historical-writer fixture bytes with real marker-bearing evidence. The added test formed by removing `pages` from a current manifest is useful for descriptor compatibility, but is not by itself evidence for every historical identity-restoration case. Preserve the supported live format; do not delete recovery support merely because it is called legacy.

Source/diff review only; I did not run the reported tests or reproduce the operator journey.

ScriptedAlchemy (changes requested):
Two P1s remain unresolved and should block this PR from landing into #707:

1. **Legacy restore must not allocate corpus-wide identity maps.** The current compatibility decoder retains the decoded files + evidence DOM and then builds owned `BTreeMap<String, String>` entries for every symbol/chunk. Replace that with marker parsing/direct indexed resolution (or another demonstrably bounded lookup) so migration cost is proportional to evidence actually encountered, not total corpus identities. This is especially important because #852 already establishes that simultaneous generation representations are the dominant memory risk.

2. **Do not replace a 30s timeout with the 168-hour background budget.** Corpus-scale restart can have a different budget from an interactive request, but it still needs a measured finite bound plus progress/cancellation checkpoints. The acceptance should prove a stalled phase releases the semantic worker promptly; successful completion of the 3,863-file operator profile proves the happy path, not bounded failure behavior.

Also complete #886 while touching this codec: share the common paged/legacy descriptor validation between the full parser and streaming retention reader without merging their authentication responsibilities.

The missing changeset is not independently concerning if this PR is intentionally absorbed into unreleased #707 before release; do not add release metadata merely to silence the bot in that case.

After the fixes: rerun the focused codec/usecase tests already listed in the PR, add a peak-memory measurement for the legacy migration fixture, add a cancellation/stall test for restart, and get the forked-workflow runs approved/green. Current workflow results are `action_required`, so there is no CI evidence on this head yet.

ScriptedAlchemy (changes requested):
Current-head deep review (`316e8e73f`). The earlier corpus-wide identity-map P1 and 168-hour authority P1 have been materially addressed: legacy markers now resolve directly from persisted indices, descriptor layout validation is shared, explicit null is rejected, and background graph authority is finite at 15 minutes. Those prior objections should not be carried forward unchanged.

One new merge blocker remains in the current branch-add wait implementation: `BRANCH_GENERATION_SEATING_POLL_INTERVAL = 10ms` drives the whole exact-generation wait loop, and the loop calls `dashboard_freshness(...).await` before every select. A long but healthy generation can therefore execute ~100 freshness probes/sec for up to 30 minutes (~180k probes per waiter), plus serving-scope reads, with unrelated publication wakeups causing more iterations. Multiple branch-add callers multiply that background demand. The comment says polling exists only to bridge the narrow publication-before-seating race; scope it to that race instead of polling the entire generation lifetime. Prefer a seating/serving-generation notification tied to the existing owner; otherwise only start a bounded, slower confirmation poll after a matching publication event. Keep the hard deadline and idle-progress semantics.

I would also keep the PR's own stated acceptance gaps explicit blockers for claiming #707 live-migration acceptance: green fork workflows on this exact head, matching-runtime BOT CLI/MCP verification, and at least one genuine historical pre-paging fixture peak-RSS measurement. The synthetic 'remove pages from current writer' fixture proves descriptor compatibility, not historical memory behavior.

The 15-minute corpus authority is now qualitatively correct (finite + lifecycle cancellation), but the acceptance should record the measurement that justifies 15 minutes rather than only asserting the constant in a unit test. A constant-equality test protects configuration, not the cost model.

ScriptedAlchemy:
Addressed the `branch_add` seating-poll finding in c6130cb23.

**Root cause.** The scheduler broadcasts a generation publication when reconcile *seals* (`registry.rs`, `publish_generation` in the worker loop), which happens before the sealed generation takes the serving slot (`ServingSwapOutcomeV1` swap, several hundred lines later). `await_exact_branch_generation` needs the *seated* slot, and no second publication follows the seat — so the only way it could observe the transition was to poll for it. That is what `BRANCH_GENERATION_SEATING_POLL_INTERVAL` (10ms) was doing, for up to the 30 minute hard deadline: ~180k iterations per waiter, each doing two async registry lookups.

**Fix.** `CodeIndexSchedulerRegistryV1` now carries a `tokio::sync::watch` seating counter (`serving_seats` / `subscribe_serving_seats`) advanced *after* the serving slot is written, in both seat paths:
- the reconcile serving swap, when `ServingSwapOutcomeV1::installs()`
- the ignored-dependency serving swap

`branch_add` subscribes before it requests the refresh and awaits that transition alongside the publication stream. The remaining timer branch is `sleep_until(idle_deadline.min(hard_deadline))`, so both existing bounds still fire and neither is polled for. A closed seat channel disables that select branch rather than spinning; the deadline branch still bounds the wait.

I reused the existing publication broadcast rather than adding a second one — it fires too early for this waiter by design, so a distinct seating signal was needed. No existing notify/watch covered seating (`grep -rn "serving_generation_epoch"` shows the two seat sites bump only an `AtomicU64`, which is not awaitable).

**Evidence.**

New test `serving_seat_wake_arrives_only_after_the_slot_is_seated` (`crates/tracedecay-code-index-runtime/src/code_index_scheduler/tests.rs`) subscribes to the seat counter before mounting, then asserts the wake arrives *and* that the serving slot is already populated at that moment — the invariant that makes the poll unnecessary. It fails if anyone moves the signal back before the seat.

```
test code_index_scheduler::tests::serving_seat_wake_arrives_only_after_the_slot_is_seated ... ok
test result: ok. 1 passed; 0 failed; ... finished in 0.37s
```

0.37s against the 10ms×N poll it replaces, with no poll interval left in the loop.

New test `the_hard_deadline_bounds_every_seating_wake` covers the deadline selection, including an idle deadline extended past the hard bound:

```
test daemon::branch_add::wait_policy_tests::the_hard_deadline_bounds_every_seating_wake ... ok
test daemon::branch_add::wait_policy_tests::pending_graph_activation_keeps_exact_branch_wait_live ... ok
test result: ok. 2 passed; 0 failed
```

`cargo clippy -p tracedecay -p tracedecay-code-index-runtime --all-targets -- -D warnings` — 0 errors, 0 warnings. That required fixing two pre-existing clippy failures in `branch_add.rs` (`items_after_test_module`, `duration_suboptimal_units`); both reproduce on 316e8e73f without my change and would block CI on their own.

`cargo test -p tracedecay --features test-helpers,test-transport --test daemon_suite git_watch_test` fails 3 of 5 (`concurrent_reconciliations_converge_on_one_sealed_generation`, `linked_worktree_requires_mount_then_serves_only_its_exact_generation`, `one_reconciliation_covers_a_fifty_commit_frontier`). **Identical failing set on 316e8e73f with the change stashed** — pre-existing on this branch (indexing does not converge in time on a loaded box), not a regression from this commit. Flagging rather than fixing, as it is outside this finding.

ScriptedAlchemy:
## Validation on head c6130cb23 (2026-09-05)

| Item | Verdict | Evidence |
|---|---|---|
| Seating wake tests | PASS | `serving_seat_wake_arrives_only_after_the_slot_is_seated` 1 passed (code-index-runtime); `daemon::branch_add` 2 passed (`the_hard_deadline_bounds_every_seating_wake`, `pending_graph_activation_keeps_exact_branch_wait_live`) |
| `cargo clippy --workspace --all-targets --locked -- -D warnings` | PASS | 0 warnings, 4m39s |
| Exact-runtime CLI/MCP generation identity | PASS | isolated daemon (`scripts/with-isolated-tracedecay-daemon.sh`), `tracedecay_runtime`, CLI `tracedecay_search` and MCP `tools/call tracedecay_search` all report the same `generation.v1.8ece8036.00000001…`, freshness `current`/`fresh`, exact/graph/lexical coverage `complete` |
| Semantic stage on exact runtime | UNVERIFIED | semantic stage reports `calibration_unavailable` until `tracedecay semantic activate` runs; activation evaluation exceeded the 10 min probe window. Re-running with a longer budget. |
| Legacy (pre-paging) restore peak RSS | **FAIL** | no historical writer fixture exists on this head; scaled synthetic harness: 500 files → generation 21.2 MB, legacy VmHWM delta 52,428 KiB (2.53×), paged 256 KiB; 2000 files → 85.0 MB, legacy 221,488 KiB (2.67×), paged 272 KiB. Legacy restore is linear in corpus: `decode_legacy_generation_evidence` reads the whole evidence segment (`whole_reads=2001`), parses a full `serde_json::Value` DOM, mutates identities in place, then `from_value`s. The identity-map P1 is fixed (index-based lookups); the remaining cost is the DOM. Extrapolated to the 3,863-file operator profile: ~430 MiB transient. |
| Trial merge onto `codex/tracedecay-total-redesign-plan-reopened` (e5dffc9cc) | one cosmetic conflict | `code_index_scheduler/registry.rs`: base renamed `publish_generation` → `broadcast_generation_publication`; #882 inserted `subscribe_serving_seats`/`record_serving_seat` above it. Keep both, use the new name. |

Not merging yet. Next: stream legacy evidence without the DOM so restore RSS stays bounded, keep the RSS harness as a real bounded test, then re-run the semantic activation probe.

ScriptedAlchemy:
Semantic activation probe (run 3, after waiting for index readiness): `tracedecay semantic activate` is rejected with `hybrid-conservative:train fallback bytes changed`, and `tracedecay_runtime` reports `semantic_runtime.state=degraded reason=invalid_runtime_status` before any activation. CLI and MCP search agree on the exact generation but semantic stays `calibration_unavailable`. Filed as #905; the exact-runtime semantic validation item stays open until that is fixed.

ScriptedAlchemy:
Legacy restore blocker addressed in `4f9091548`, with the base merged in `3c83c9e7a`.

**Root cause.** `decode_legacy_generation_evidence` read the whole pre-paging evidence segment into a `Vec<u8>`, parsed a full `serde_json::Value` DOM from it, rewrote identity markers in that tree, then `from_value`d into the typed payload. Peak RSS was segment + DOM + payload and grew with the corpus — reproduced at **2.348x** the on-disk generation at 500 files and **2.370x** at 2000, i.e. ~5.5x the evidence segment itself in both cases.

**Fix.** Two changes, both reusing what the paged form already does:

1. The pre-paging segment is now read through `PartitionedEvidencePageReaderV1`. It carries no page table, so the reader ranges it in 256 KiB chunks and authenticates it with the aggregate segment digest it always had — there are no per-page digests to skip. `verify_partitioned_sealed` takes the same path, so verification no longer materializes a legacy segment either (that branch is now deleted rather than special-cased).
2. Identity markers are restored by a `serde` transcoder wrapped around that reader, so the typed payload is built directly from the stream. The classification rules are the replaced DOM walk's, unchanged: a string is substituted by the object key that encloses it, reset at every object member and inherited through arrays. An invalid symbol or chunk key still fails with its original contract message, carried out of `serde`'s error type rather than stringified into it.

**Numbers.** Measured as peak RSS (VmHWM) growth over a baseline warmed by a paged restore of the *same* generation. Warming isolates the pre-paging path's own cost: the restored generation's own footprint (~1.2x the on-disk size, unavoidable — `decode_partitioned_sealed` returns it) is already in the allocator, so what is left is the extra the legacy path asks for.

| files | generation on disk | evidence segment | before | after |
| --- | --- | --- | --- | --- |
| 300 | 12,750,156 B | 5,402,171 B | 1.279x (15.9 MB) | **0.232x** (2.9 MB) |
| 1500 | 63,747,800 B | 27,005,771 B | 1.272x (79.2 MB) | **0.168x** (10.5 MB) |

The ratio now shrinks as the corpus grows instead of holding constant, which is the linearity the review flagged. Absolute peak for a full restore at 500 files went from 47.6 MB to the paged form's own ~24 MB.

**Test.** The measurement harness is a real bounded test now, not an `#[ignore]`d probe: `legacy_generation_restore_does_not_materialize_its_evidence_segment` builds a 300-file generation, rewrites its descriptor into the pre-paging shape, and fails if the legacy restore's peak growth reaches half the on-disk generation size. It takes its reading in a child process because VmHWM is process-wide, and returns early where `/proc/self/status` and `clear_refs` are unavailable. Confirmed it fails on the pre-fix code (1.279x and 1.272x against a 0.5x bound) and passes after (0.232x / 0.168x, stable across runs).

`partitioned_codec_reads_pre_paging_evidence_descriptor` now asserts the read shape this produces — no whole-segment read of the evidence, and no single read larger than one page — instead of the previous "exactly one whole read". Its identity assertion (a legacy manifest restores byte-identically to the paged one) is unchanged.

**Verification.** `cargo test -p tracedecay-code-index` green (all suites). `cargo clippy --workspace --all-targets --locked -- -D warnings` clean. `cargo test -p tracedecay-code-index-runtime --lib` has two failures — `registry_feeds_publications_and_bounded_freshness_reads` and `semantic_schedule_order_tests::panicking_semantic_hook_does_not_retire_later_reconciliation`, both `Elapsed(())` waiting on a generation publication — that reproduce identically on `origin/codex/tracedecay-total-redesign-plan-reopened` at `e5dffc9cc` with none of this branch's work, so they are pre-existing on the base and not from this change or the merge.

**Merge.** `3c83c9e7a` merges the base. The only conflict was `registry.rs`, where the base renamed `publish_generation` to `broadcast_generation_publication` and this branch inserted `subscribe_serving_seats` / `record_serving_seat` immediately above it; both sides are kept, with the base's new name and its doc comment.

ScriptedAlchemy:
Merging into codex/tracedecay-total-redesign-plan-reopened per owner decision (#707 is the long-lived integration branch). Validation closed: branch_add seating is watch-driven (c6130cb23; serving_seat_wake_arrives_only_after_the_slot_is_seated, the_hard_deadline_bounds_every_seating_wake); legacy pre-paging restore streams the evidence segment (4f9091548; 0.23× generation at 300 files, 0.17× at 1500, was 1.28×/2.7×), bounded test not ignored; workspace clippy clean; CLI and MCP report the same exact generation on an isolated daemon. The semantic activation defect found by that probe (#905) is fixed on the base branch (c5fe03731…4e6ecf029) and is inherited by this merge.

Issue #933, opened 2026-09-06
[707/P1] Make crash and snapshot fixtures exercise recovery without illegally copying live Windows stores


ScriptedAlchemy:
Priority: P1
Blocks undraft #707: yes
Labels: bug, ci, P1, v2, windows (suggested)

### Symptom

**Classification:** Durability test preconditions; possible underlying file-sharing defects.

Four Windows crash/snapshot tests fail on a native lock while constructing the scenario; one payload test misses its intended size refusal.

Baseline: [run 34039753429](https://github.com/ScriptedAlchemy/tracedecay/actions/runs/34039753429), head `9d17537b4`. Exact test rows below come from the archived reports; Dashboard comes from its job log.

- **windows-2** — `tracedecay-graph-db::durability_crash_contract crc_faulted_store_is_quarantined_with_its_wal_sidecar_and_rebuilt` — failed initial and retry ([job](https://github.com/ScriptedAlchemy/tracedecay/actions/runs/34039753429/job/101514939956)).
- **windows-3** — `tracedecay-runtime-core sqlite_read_snapshot::tests::copied_snapshot_survives_empty_writer_sidecar_cleanup` — failed initial and retry ([job](https://github.com/ScriptedAlchemy/tracedecay/actions/runs/34039753429/job/101514939885)).
- **windows-5** — `tracedecay-graph-db runtime::tests::reopen_of_a_dirty_store_copy_replays_the_walled_commit` — failed initial and retry ([job](https://github.com/ScriptedAlchemy/tracedecay/actions/runs/34039753429/job/101514939889)).
- **windows-5** — `tracedecay-graph-db::durability_crash_contract reopen_collapses_replayed_wal_history_from_an_unclean_shutdown` — failed initial and retry ([job](https://github.com/ScriptedAlchemy/tracedecay/actions/runs/34039753429/job/101514939889)).
- **windows-2** — `tracedecay-lcm payload::filesystem_authority::authority_tests::oversized_payload_is_rejected_before_allocation_or_read` — failed initial and retry ([job](https://github.com/ScriptedAlchemy/tracedecay/actions/runs/34039753429/job/101514939956)).

### Root cause

[...]

PR #951, opened 2026-09-06
fix(snapshot): live Copy-mode backup; bound crash-child fixtures

ScriptedAlchemy:
## Two distinct changes

This PR still fixes https://github.com/ScriptedAlchemy/tracedecay/issues/933, stacked on `codex/tracedecay-total-redesign-plan-reopened` (PR #707). Do not undraft or merge #707.

The production snapshot path and the crash fixtures are separate claims.

### 1. Production: Copy-mode live snapshot (not a test-only change)

`SnapshotMode::Copy` and `backup_live_sqlite_database` use the SQLite online backup API. They do **not** `fs::copy` a locked Windows main/WAL/SHM family, and they do **not** open a changing source with `immutable=1`.

**Publication boundary**

- Cancel/deadline is checked before any scratch is created.
- Staging is reserved with `create_new` (owner-only on Unix). A colliding candidate is skipped — not opened, truncated, or deleted. Exhausting candidates is a typed `AlreadyExists`.
- Failure retires only the family this attempt exclusively created (`-wal`/`-shm`/`-journal` + main). Pre-existing destinations and foreign scratch are left untouched.
- Unix: `rename(2)` atomically replaces an existing destination. Failure leaves the old file at its original path.
- Non-Unix: an existing destination is rejected (`AlreadyExists`). Displace/restore is not used; it cannot keep the old snapshot at its path until replace succeeds. Scratch and LCM callers already publish to a new path.

Acceptance in `sqlite_read_snapshot_backup_tests.rs`:

- Concurrent WAL writer → coherent committed prefix.
- WAL-resident rows included without checkpointing the source family.
- Cancel/deadline interrupt page-copy work and DELETE-journal Busy/Locked retries (Busy/Locked probed first).
- Failed/cancelled backup never publishes a new destination and never rewrites a prior destination.
- Exclusive staging refuses a collision without deleting it; reservation steps past occupied names and retires only the owned file.
- Unix atomically replaces an existing destination; non-Unix refuses that replace and keeps the old file.
- Foreign snapshots leave source main/WAL/SHM untouched.
- Successful backup is not a freshness claim after the source changes.
- Windows: live WAL writer survives backup to a **new** path; raw `fs::copy` still fails with OS 32/33.

### 2. Fixtures: subprocess unclean-exit, not power-loss

Grafeo crash/reopen tests spawn a child that reaches the durable WAL phase and exits without a clean close. The parent bounds that process (45s), kills and reaps it on timeout, and fails if the durable-phase marker is missing. Only then is the abandoned container + sidecar copied.

This proves unclean **process** exit recovery. It does not prove machine power-loss durability.

The oversized LCM payload test drops the creator handle before size admission.

### 3. `TRACEDECAY_SQLITE_UNSAFE_FAST`

Removed the unused Windows CI export. Crash children still `.env_remove` it. Writer policy is checked directly (`WAL` / `synchronous=NORMAL` / `wal_autocheckpoint=0`).

## Test plan

- [x] Linux lib + crash fixtures at `fc060bec3` (Zack): `tracedecay-runtime-core` 757 passed; all 18 `backup_tests::*`; `durability_crash_contract` 7 passed
- [x] clippy `-D warnings` on the edited crates
- [ ] Native Windows shards on this head (create-only dest, live backup, crash-child reap). Hosted `CI` is queued behind an older same-PR run (`cancel-in-progress: false`).

## Checklist

- [ ] `CHANGELOG.md` updated (under `[Unreleased]` if no version bump)
- [x] Breaking changes documented (if any)

<div><a href="https://cursor.com/agents/bc-dde38e91-6ed1-4a21-a2df-c4ca2924fdd6?cursor_ref=pr_footer&cursor_cta=open_in_web"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a>&nbsp;<a href="https://cursor.com/background-agent?bcId=bc-dde38e91-6ed1-4a21-a2df-c4ca2924fdd6&cursor_ref=pr_footer&cursor_cta=open_in_cursor"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;</div>


ScriptedAlchemy (commented):
## Architecture review — HOLD for production snapshot and fixture-lifecycle gates

Reviewed `8518e088005fcada58f71fffeb9f8ba893b20f68`; source review only, exact-head CI pending. Child reaches durable phase -> exits without clean close -> parent joins -> copy abandoned image is the correct crash fixture. Closing the oversized payload's creator before testing size admission is also correct.

**P1:** `sqlite_read_snapshot.rs` changes production Copy-mode semantics, not just tests. Keep this change visibly separate in the PR's claims and prove a snapshot of a concurrently written WAL database is coherent, cancellation/deadline interrupts Busy/Locked and page-copy work, partial scratch is retired, and the caller never receives an incompletely backed-up database. Preserve the Foreign-source no-mutation contract for the source main/WAL/SHM family; `SQLITE_OPEN_READ_ONLY` alone is not the entire acceptance proof. Exercise absent/cleaned sidecars and Windows live-writer behavior. No `immutable=1` on a changing source to bypass locking.

**P1 test isolation:** the new `refuse_sqlite_unsafe_fast` performs process-global unsafe environment mutation in ordinary library tests. Prefer `.env_remove` on the controlled crash child and remove the audited-unused CI export, rather than mutating shared libtest environment to prove an inert flag is absent. Verify the real runtime journal/sync policy directly. A comment calling an unread variable a speed knob is misleading.

Bound and reap crash child processes on every failure path; a child that never publishes the durable-phase marker must fail the fixture, not hang the suite or leave locks behind. Keep subprocess-crash proof distinct from a claim about machine power-loss durability. Native Windows run and the affected snapshot/backup suites on the #707 merge tree remain required.

ScriptedAlchemy (commented):
## System-level re-review at `f24ab494b` — two concrete fixes required

The production snapshot change is not merely a fixture repair. I read the online-backup implementation and the new backup tests. Removing the unused shard-wide unsafe environment setting and separating the crash child from the live parent are improvements. This review does not claim to have run the Rust suites.

### P1 — cleanup must not delete a pre-existing destination

`sqlite_read_snapshot.rs::backup_live_sqlite_database_with` unconditionally removes `destination` whenever `run_online_backup` errors. That includes the very first cancellation checkpoint or failure to open the source, before this attempt has created or published a destination. The public `backup_live_sqlite_database` path has no new-destination precondition or ownership receipt, so an unreadable/missing source can delete an existing valid backup. `publish_complete_backup` likewise unlinks the destination before renaming and deletes it again on rename failure, so failed replacement need not preserve the prior snapshot.

Delete only scratch material created by this attempt. Either explicitly enforce create-only destination semantics without touching an existing path, or retain the old destination until an authorized atomic replacement succeeds. Use attempt-owned staging (the fixed `.backup-partial` name also cannot distinguish two public helper calls), and reject source/destination aliases. Test an existing destination with source-open failure, cancellation before first step, failure after copying and publication failure; its bytes and identity must survive each failed attempt. A scratch-copy caller whose destination is always new can keep that narrower contract, but it must not make the public helper's cleanup destructive.

### P2 — the new Busy/Locked test does not actually establish reader contention

`sqlite_read_snapshot_backup_tests.rs::live_backup_deadline_interrupts_busy_locked_retries` uses WAL mode plus `BEGIN EXCLUSIVE` on another connection, then expects read-side backup to time out. SQLite's transaction contract makes EXCLUSIVE equivalent to IMMEDIATE in WAL mode; this does not exclude snapshot readers. An isolated mechanism check here with SQLite 3.46.1 copied the fixture successfully in 0.143 ms while the writer transaction remained open (backup status DONE, both pages copied). This is not execution of the pinned rusqlite test, but it confirms the fixture premise is wrong, in agreement with SQLite's Transaction and Online Backup API documentation.

Use an actual rollback-journal exclusive-reader conflict, or a controlled backup-step boundary, and prove a Busy/Locked observation occurs before testing deadline/cancellation. Otherwise a fast runner fails `expect_err`, while a loaded runner can accidentally pass by using up 50 ms on unrelated work. Keep the WAL-concurrent-writer test separately; there success is the desired behavior.

### Preserve the system contract

The copied image must be transactionally coherent and its later attach token must still validate the source generation; a successful backup alone does not authorize a freshness claim after the source changes. Keep checkpoint/deadline checks around the cooperative page loop and never use `immutable=1` on the live source. Crash-child evidence proves process-crash recovery, not power-loss durability. The returned exact-head CI/hotpath-coverage are pending and SDK/profile runs queued.

ScriptedAlchemy (commented):
## Follow-up on the fix pushed during review — `a3e902b7a`

The final inventory showed this newer head, so I re-read the backup implementation and the new tests before carrying the earlier findings forward.

**Resolved in source:** source-open and cooperative cancellation failures now retire the attempt's staging path rather than deleting the pre-existing destination. The new tests assert preserved destination bytes and file identity. The Busy/Locked fixture now uses DELETE journal mode, and explicitly observes a Busy/Locked step before testing the deadline. Withdraw the earlier WAL-EXCLUSIVE fixture objection. The WAL concurrent-writer test correctly remains a success path. I did not execute these repository tests.

**Remaining Windows publication gap:** `replace_existing_destination` under `cfg(not(unix))` first renames the existing destination to `displaced`, then renames staging to destination. If the second move fails, restoration is attempted with `let _ = fs::rename(&displaced, destination)`. A failed restoration is silently discarded; the function returns the staging-rename error without reporting that the old snapshot is no longer at its original path. A process exit between those two moves likewise leaves the old file displaced with no recovery record. Thus the implementation still does not establish the documented promise that the old destination stays in place until replacement succeeds. The current forced-publish-failure test exits before any displacement and cannot exercise this branch.

Use a supported atomic replacement authority, or reject replacing an existing destination where that contract cannot be provided. If non-atomic displacement is an intentional requirement, give it explicit recoverable ownership and report restoration failure rather than discarding it. Add native Windows failure injection after displacement and on restoration itself; prove recovery of the exact old file, not merely that some bytes remain at an unreported scratch name. This is distinct from the now-fixed early-error deletion.

Also make scratch ownership real at creation: PID/counter naming alone is not exclusive creation, and `Connection::open_with_flags(...CREATE)` can open a stale same-name file. Reserve the staging file exclusively before treating it as this attempt's deletable material, or refuse collisions. Do not delete a colliding file on an early checkpoint/source-open error.

The original findings have therefore been narrowed, not left stale: early failure cleanup and the lock-test premise are corrected; safe replacement and scratch ownership remain the publication boundary to finish.

ScriptedAlchemy:
## Evidence for the `a3e902b7a` follow-up — both remaining items are closed at `fc060bec3`

Head `fc060bec3` = PR branch + merge of `codex/tracedecay-total-redesign-plan-reopened@79735b580` (`cdf9d0e2b`, clean, no conflicts; the base did not touch any PR-owned file) + one new test commit. 0 commits behind base.

### Scratch ownership at creation — `960433915`

- `reserve_attempt_staging` creates the staging file with `OpenOptions::create_new` (0600 on Unix) and steps past `AlreadyExists` to the next `{dest}.{pid}.{n}.backup-partial` candidate; a colliding name is never opened, truncated, or deleted. Exhausting 32 candidates is a typed `AlreadyExists` error with nothing removed.
- The cancel/deadline checkpoint now runs *before* any reservation, so a first-checkpoint failure touches no filesystem entry.
- `run_online_backup` and the test probe open staging with `SQLITE_OPEN_READ_WRITE` only — `SQLITE_OPEN_CREATE` is gone, so SQLite can no longer adopt a stale same-name file.
- Failure paths call `retire_failed_backup(error, &[staging])` for the reserved path only (cleanup failure is reported in the error, not discarded).

### Windows publication — `960433915`

- `replace_existing_destination` under `cfg(not(unix))` no longer displaces the old file. It retires the attempt's staging and returns `AlreadyExists` ("cannot atomically replace an existing destination"). The old destination is never renamed, so no process exit or failed restore can leave it displaced. Unix keeps atomic `rename(2)`.
- Every production Copy-mode caller (`copy_snapshot_family`) publishes into a fresh private scratch directory, so the reject path is unreachable there; only the public helper can hit it, and it now fails closed.

### New in this push — `fc060bec3`

`backup_tests::early_source_open_error_skips_colliding_scratch_and_retires_only_owned_staging`: pre-seeds 16 foreign files at the next candidate staging names, fails the attempt at source open, and asserts (a) the error is not `AlreadyExists`, (b) every foreign file keeps its bytes, (c) the set of `.backup-partial` files afterwards equals exactly the foreign set. Verified falsifiable: a `remove_file`-then-`create_new` reservation mutant fails it. `backup_staging_path` now takes the id so the test can name future candidates without consuming them.

### Local runs at `fc060bec3` (hosted CI is not executing jobs today)

| Command | Result |
|---|---|
| `cargo test -p tracedecay-runtime-core` | lib `757 passed; 0 failed; 1 ignored` (all 18 `backup_tests::*`, 2 `cancellation_tests::*`), integration suites 2/5/1/15/1/1/4/1/4/3 passed, doctests 1+3 passed |
| `cargo test -p tracedecay-store-runtime` | `87 passed; 0 failed; 1 ignored` |
| `cargo test -p tracedecay-graph-db --features test-helpers` | lib `175 passed; 0 failed; 4 ignored`; `durability_crash_contract` `7 passed; 0 failed`; `backup_contract` 11, `registry_contract` 31, `runtime_contract` 46, `verified_generation_contract` 57, others green |
| `cargo check -p tracedecay-runtime-core -p tracedecay-store-runtime -p tracedecay-graph-db --tests --target x86_64-pc-windows-gnu` | Finished, 0 errors (14 warnings, all pre-existing base `cfg(windows)` dead-code/unused-import in `storage/`, `bounded_git_command.rs`, `registry_contract.rs`, code-index crates — none in PR-owned files) |
| `cargo clippy -p tracedecay-runtime-core -p tracedecay-store-runtime -p tracedecay-graph-db --all-targets -- -D warnings` | Finished, 0 warnings |
| `cargo fmt --check` | clean |

Directly relevant tests, all green: `live_backup_preserves_existing_destination_when_{source_cannot_be_opened,cancelled_before_first_step,cancelled_after_copying,publish_fails}`, `exclusive_staging_refuses_a_collision_without_deleting_it`, `early_source_open_error_skips_colliding_scratch_and_retires_only_owned_staging`, `retiring_owned_staging_removes_the_complete_sqlite_family`, `live_backup_replaces_an_existing_destination_atomically` (unix), `live_backup_refuses_to_replace_an_existing_destination` (not-unix; compiled under the windows-gnu check), `rollback_journal_exclusive_lock_makes_the_first_backup_step_busy`, `live_backup_deadline_interrupts_busy_locked_retries`.

Not claimed: a native Windows execution of the `cfg(not(unix))` test — it is compile-checked here; the crash-child suite proves process-crash recovery, not power-loss durability.

PR #956, opened 2026-09-06
test: make socket fixtures signal admission instead of racing accept and handshake

### LINKED ISSUE #932: [707/P1] Make socket fixtures signal admission instead of racing accept and handshake
Priority: P1
Blocks undraft #707: yes
Labels: test, ci, P1, v2, windows, macos (suggested)

### Symptom

**Classification:** Test infrastructure; confirmed WouldBlock subcase and separate handshake hypothesis.

Windows local test servers panic on WouldBlock; macOS controlled invocation never completes its handshake.

Baseline: [run 34039753429](https://github.com/ScriptedAlchemy/tracedecay/actions/runs/34039753429), head `9d17537b4`. Exact test rows below come from the archived reports; Dashboard comes from its job log.

- **macos** — `tracedecay-daemon-protocol client::controlled_invocation_tests::remote_effect_without_authoritative_settlement_returns_reset_required` — failed initial and retry ([job](https://github.com/ScriptedAlchemy/tracedecay/actions/runs/34039753429/job/101514387881)).
- **windows-4** — `tracedecay-sdk::client multi_root_operations_reach_their_exact_project_application_routes` — passed retry; CI failed-on-flaky ([job](https://github.com/ScriptedAlchemy/tracedecay/actions/runs/34039753429/job/101514939860)).
- **windows-5** — `tracedecay-semantic model_lifecycle::tests::fresh_hub_acquisition_downloads_then_reuses_private_cache_offline` — passed retry; CI failed-on-flaky ([job](https://github.com/ScriptedAlchemy/tracedecay/actions/runs/34039753429/job/101514939889)).

### Root cause

Two Windows fixtures panic with native WouldBlock 10035 in their local mock server; their caller then reports an SDK/admission or model DownloadFailed symptom. The macOS remote-effect test fails during invocation handshake before testing settlement uncertainty. These are fixture protocol-readiness failures, not evidence that remote effects were incorrectly committed or that the public model hub was unavailable.

### Exact fix

Repair the mock server helpers in `crates/tracedecay-sdk/tests/client.rs`, `tracedecay-semantic/src/model_lifecycle/tests_first.rs`, and the daemon-protocol controlled-invocation fixture. Bind/listen first and communicate readiness. Treat nonblocking WouldBlock as not-ready using async readiness or a dedicated blocking accept thread with owned shutdown, not unwrap or busy-spin. Exchange the exact handshake before injecting the target failure. Retain server threads/tasks and surface their errors to the parent; use phase signals for fault injection and cleanup all sockets at teardown.

Source/implementation entry points:
- [`crates/tracedecay-sdk/tests/client.rs`](https://github.com/ScriptedAlchemy/tracedecay/blob/9d17537b43cd49e9719c932c651bc152d6119390/crates/tracedecay-sdk/tests/client.rs)
- [`crates/tracedecay-semantic/src/model_lifecycle/tests_first.rs`](https://github.com/ScriptedAlchemy/tracedecay/blob/9d17537b43cd49e9719c932c651bc152d6119390/crates/tracedecay-semantic/src/model_lifecycle/tests_first.rs)
- [`crates/tracedecay-daemon-protocol/src/client/controlled_invocation_tests.rs`](https://github.com/ScriptedAlchemy/tracedecay/blob/9d17537b43cd49e9719c932c651bc152d6119390/crates/tracedecay-daemon-protocol/src/client/controlled_invocation_tests.rs)

### Acceptance

The malformed SDK reply fails after actual HTTP admission; hub acquisition completes against the local fixture and then reuses cache offline; missing authoritative remote settlement yields ResetRequired after the operation is admitted. No fixture-server panic, stale socket or timing sleep is necessary.

### Notes / related issues

NEW shared fixture-readiness issue with explicit subcases, #896/#863 tracking. If the macOS handshake remains after deterministic setup, split its wire-protocol failure rather than attributing it to the Windows WouldBlock mechanism.

All work targets `codex/tracedecay-total-redesign-plan-reopened`. Keep #707 draft; no undraft or merge is authorized by this draft.

---
_Filed from ChatGPT 6 Pro review of PR #707 CI baseline [run 34039753429](https://github.com/ScriptedAlchemy/tracedecay/actions/runs/34039753429) (head `9d17537b4`). Local draft #19 (P1). Do not undraft/merge #707 without Zack authorization._

ScriptedAlchemy:
## Summary

- Keep the SDK `serve_until_stopped` server and the semantic `FixtureHub` listeners in blocking mode, stopped through an owned flag plus one wake-up connection instead of a nonblocking accept loop that busy-polled `WouldBlock`.
- Rework the daemon-protocol controlled-invocation fixture so the fake daemon signals `admitted` and `control_observed` phases, and the tests pause/advance the tokio clock past `DAEMON_TOOL_RESPONSE_GRACE` only after every real loopback exchange has completed.
- Fixture servers now exit on release and are joined, so their panics surface to the test instead of being aborted.

## Motivation

Fixes #932 (P1, blocks undraft of #707). Targets `codex/tracedecay-total-redesign-plan-reopened`; #707 stays draft.

Two distinct fixture-readiness defects were behind the baseline failures:

- **Windows (`tracedecay-sdk::client multi_root_operations_reach_their_exact_project_application_routes`, `tracedecay-semantic model_lifecycle::tests::fresh_hub_acquisition_downloads_then_reuses_private_cache_offline`)** — both fixtures set the *listener* nonblocking. On Windows an accepted socket inherits the listener's nonblocking flag, so the server-side request `read` failed with WSAEWOULDBLOCK (10035), the server thread panicked, and the client saw a reset connection (reported as an SDK protocol/admission error or `DownloadFailed`).
- **macOS (`tracedecay-daemon-protocol client::controlled_invocation_tests::remote_effect_without_authoritative_settlement_returns_reset_required`)** — the test ran with `start_paused = true` while the choreography used real loopback sockets. A paused tokio clock auto-advances to the next timer whenever the runtime would otherwise block on I/O, so the client's 10s admission-deadline `sleep` fired before loopback connect readiness was observed; the client dropped the connection and the fixture read EOF at `"invocation handshake"` (junit: panic at line 1072, then `request admission: RecvError`). Linux loopback readiness is synchronous enough that the zero-timeout I/O poll always won, which is why it only showed on macOS. I reproduced the mechanism standalone: with a paused clock, a `select!` between a socket read and a 10s `sleep` resolves to the sleep in ~100µs when the peer writes 5ms later.

## Changes

- `crates/tracedecay-sdk/tests/client.rs` — `StoppableServer` replaces the `mpsc` stop channel; blocking `accept`, stop flag + wake-up connect, `stop()` joins and returns the recorded requests.
- `crates/tracedecay-semantic/src/model_lifecycle/tests_first.rs` — `FixtureHub` binds/listens before handing out the endpoint, accepts blocking until every expected request is served, and `finish()` stops it via the same wake-up pattern (a short acquisition now fails the request-count assertion instead of hanging or relying on the old 5s deadline/5ms poll).
- `crates/tracedecay-daemon-protocol/src/client/controlled_invocation_tests.rs` — `UnsettledDaemon` carries `admitted`, `control_observed`, `release`, and the server handle; `settle_without_daemon_response` drives the two unsettled tests; `reset_then_reconnect_client` also signals when it has read the typed cancellation. All three former `start_paused` tests now run in real time until the control phase is observed, then `virtualize_response_grace()` pauses and advances; the reconnect test calls `tokio::time::resume()` before its second, real-I/O invocation.

Non-obvious: in the `CancellationConnectionRejected` case the client's bounded (250ms) cancel-delivery attempt still runs under the paused clock after `control_observed`; every path it can take ends in the same indeterminate settlement, which is documented on the field.

## Test plan

- [x] `cargo test -p tracedecay-daemon-protocol --lib controlled_invocation_tests` — 15 passed; the three fixed tests looped 30x with no failure
- [x] `cargo test -p tracedecay-sdk --test client` — 18 passed; `multi_root_operations_reach_their_exact_project_application_routes` looped 30x
- [x] `cargo test -p tracedecay-semantic --features semantic-fastembed --lib model_lifecycle::tests` — 35 passed; `fresh_hub_acquisition_downloads_then_reuses_private_cache_offline` looped 30x
- [x] `cargo clippy --tests` on the three crates (with and without `semantic-fastembed`) — no new warnings; `cargo fmt --check` clean
- [x] `npm run lint:commit` passes for both commits
- [ ] Windows/macOS CI on this PR is the real acceptance check for the platform-specific paths; if the macOS handshake failure persists after this deterministic setup, split it as a wire-protocol issue per the note in #932

## Checklist

- [ ] `CHANGELOG.md` updated — not applicable (test-only change)
- [x] Breaking changes documented (if any) — none

<div><a href="https://cursor.com/agents/bc-53886c9c-0d17-4d66-a7fb-a4cfb82e23ea?cursor_ref=pr_footer&cursor_cta=open_in_web"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a>&nbsp;<a href="https://cursor.com/background-agent?bcId=bc-53886c9c-0d17-4d66-a7fb-a4cfb82e23ea&cursor_ref=pr_footer&cursor_cta=open_in_cursor"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;</div>


ScriptedAlchemy:
This branch no longer merges cleanly with `codex/tracedecay-total-redesign-plan-reopened` (tip 5808a7511). Conflicting files when merging the base in: crates/tracedecay-daemon-protocol/src/client/controlled_invocation_tests.rs. Needs a rebase or base merge by the author; I left the branch untouched. The peer-landed work most likely behind the conflict: 5808a7511 (registrar keyed by store authority), 2cc98085f (capability digest), b0988105a (git argument paths on Windows), 6504d0ca2 (private DACL), d265aa4f5 (host fixture paths).

ScriptedAlchemy (commented):
## Review — correct phase separation; HOLD for cleanup and merge-state evidence

Reviewed `547452e663a0da5fb90f31332155945e92149ea1`; source review only. GitHub currently reports conflicts and CI pending. Keeping real time for socket admission, then advancing virtual time only after the typed control phase, is the right correction. Keep the reconnect's resume before its next real I/O.

**P1 fixture lifetime:** blocking accept plus a wake connection fixes the listener loop, but explicit `stop()`/`finish()` on the happy path is not full ownership. Ensure unwind/early-return drops also signal release and reap the server; an accepted connection stalled in a partial request must not make stop/join block forever, because waking the listener only unblocks accept, not that read. Test stop-before-first-request, partial request, client failure, and server panic; preserve the original test failure while reporting cleanup failure.

The `admitted`/`control_observed` signals must follow decoding and validating the actual request IDs, not merely connection acceptance. The new control parsing does that; retain it through conflict resolution. A phase barrier that never arrives needs a bounded fixture failure and child cleanup, not another increased production timeout.

After resolving the current conflicts, run the three original failing cases natively on Windows/macOS and the relevant SDK/model/daemon-protocol suites. This is a fixture repair, not evidence to weaken protocol admission or reset-required settlement. Target #707 only; keep its draft state.

ScriptedAlchemy (commented):
## System-level re-review at `547452e66` — normal choreography improved; failure cleanup remains incomplete

Re-read the SDK StoppableServer and semantic FixtureHub changes against the protocol phase-barrier approach. Keeping real loopback I/O in real time and switching to virtual time only for the response-grace phase is appropriate. The admitted/control-observed signals describe actual protocol progress; they are not substitutes for testing the real cancellation/reset result.

**P2 — make the fixture owner retire on failure, not just on the happy path.** StoppableServer has `stop(self)` but no Drop cleanup; FixtureHub relies on `finish(self)`. A test panic or early return before these calls detaches the worker, which can remain in blocking `accept` (the hub's former lifetime deadline is gone). Also, once the worker is inside `request(&mut stream)` or `serve_fixture_hub_request`, opening a new wake-up connection does not interrupt that accepted socket's read. A partially sent request whose client stays connected can therefore strand `stop()/finish()` at join.

Retain a small scoped cleanup owner that signals stop on unwind and can interrupt/bound the active accepted connection as well as accept. Avoid double-panic in Drop; surface worker failures through the explicit join on the normal path. Test three cases independently: fewer requests than expected, panic/early exit before normal finish, and a client that sends only part of a request and holds the socket open. Do not restore the nonblocking-accepted-socket bug or inflate product deadlines to mask fixture hangs.

Keep this classified as fixture lifecycle work, not a production transport defect. The real system's post-admission cancellation contract remains: absent authoritative settlement is reset-required/indeterminate, not a fabricated successful cancellation or automatic retry of a possibly committed effect. The controlled tests must keep asserting that outcome after the barrier changes.

All four returned exact-head workflows are cancelled. No native test execution was performed in this review.

Issue #929, opened 2026-09-06
[707/P1] Make LCM summary visibility agree with seeded committed summary state


ScriptedAlchemy:
Priority: P1
Blocks undraft #707: yes
Labels: bug, ci, P1, v2 (suggested)

### Symptom

**Classification:** Dashboard backend/read-model contract; not frontend test failure.

Dashboard job 101514387922: summary_nodes_total 0 vs 1 in two API tests; messages_total=3 and sessions_total=1 assertions precede and pass.

Baseline: [run 34039753429](https://github.com/ScriptedAlchemy/tracedecay/actions/runs/34039753429), head `9d17537b4`. Exact test rows below come from the archived reports; Dashboard comes from its job log.

- **dashboard-linux** — `tracedecay::dashboard_api_test api::lcm_endpoints_cover_seeded_fts_and_like_fallback` — failed single dashboard run ([job](https://github.com/ScriptedAlchemy/tracedecay/actions/runs/34039753429/job/101514387922)).
- **dashboard-linux** — `tracedecay::dashboard_api_test api::lcm_serves_project_session_store_without_global_override` — failed single dashboard run ([job](https://github.com/ScriptedAlchemy/tracedecay/actions/runs/34039753429/job/101514387922)).

### Root cause

Both Dashboard tests reach HTTP 200 and the expected project-store scope, and pass the message/session count assertions before summary_nodes_total returns 0 instead of 1. This narrows the failure to summary publication/availability/read filtering or fixture setup. It does not support a blanket wrong-database diagnosis, and no frontend failure was observed.

### Exact fix

At `crates/tracedecay/tests/dashboard_api_test/api.rs:682` and `:851`, inspect how `dashboard_api_support` seeds summary nodes and any required visibility/availability/temporal bindings. Compare that committed record with the overview reader's authoritative summary query and the search reader. Seed through the real summary publication contract when a raw node insert is insufficient; if valid committed nodes are excluded, repair the reader's join/filter. Keep project-session storage authoritative with and without the global accounting override. Do not change expected counts to zero or return unscoped summaries.

Source/implementation entry points:
- [`crates/tracedecay/tests/dashboard_api_test/api.rs`](https://github.com/ScriptedAlchemy/tracedecay/blob/9d17537b43cd49e9719c932c651bc152d6119390/crates/tracedecay/tests/dashboard_api_test/api.rs)

### Acceptance

[...]

PR #957, opened 2026-09-06
fix(dashboard): seed LCM fixture messages once so summaries stay visible

ScriptedAlchemy:
## Summary

- Fixes the two Dashboard failures from #929 (`api::lcm_endpoints_cover_seeded_fts_and_like_fallback`, `api::lcm_serves_project_session_store_without_global_override`): `summary_nodes_total` came back `0` instead of `1` after the message/session counts had already passed.
- The dashboard LCM fixture seeded each message twice (direct raw LCM ingest, then the canonical observation projection). Since the summary-convergence work (`28f15b1f4`, `ef3c912d1`), the projection's rewrite of the already-ingested `lcm_raw_messages` row is a **retained raw revision**: invalidation work is queued and every summary of the session is hidden from `current`-mode reads until the LCM effects loop drains it. The dashboard tests have no such loop, so the committed, `available` summary was never served.
- Production ingest writes the raw row only through the observation projection; the fixture now does the same, and asserts (via the convergence-queue contract the effects loop reads) that the seeded session carries no retained raw revision before publishing the summary.

Refs #929. Targets the #707 branch; #707 stays draft.

## Motivation

Issue #929 asked to make LCM summary visibility agree with the seeded committed summary state, targeting summary publication/visibility rather than a store-resolver rewrite. Locally the failure reproduced deterministically; dumping the project store after seeding showed:

- `lcm_summary_convergence_dirty_raw` rows for all three seeded raw messages (store ids 1/3/5, the gaps being the `INSERT ... ON CONFLICT DO UPDATE` from the second write), with matching `pending` rows in `lcm_summary_convergence_invalidation_work`;
- the summary node published, `available` in the active generation, with a project-owned anchor — i.e. a valid committed node that the `?N <> 'current' OR NOT EXISTS (... dirty_raw ...)` predicate in the root summary browse/search queries and record hydration deliberately excludes.

So the reader was honoring its contract; the fixture had put the store into a partial-invalidation state that production never reaches on plain ingest (`lcm_ingest_raw_message` has no production callers — only tests use it). No reader/filter change was needed, and no expected counts were touched.

## Changes

- `crates/tracedecay/tests/dashboard_api_test/dashboard_api_support.rs`
  - `seed_lcm_fixture`: drop the redundant `lcm_ingest_raw_message_for_test` call per message; the observation projection already writes `session_messages` and `lcm_raw_messages` (with the content-bound receipt hydration needs).
  - After seeding messages and before publishing the summary, assert `lcm_retained_raw_revision_for_test(...) == None` so a regression fails at the seeding step with a named reason instead of a bare `0 vs 1` on the overview.
- `crates/tracedecay/tests/dashboard_api_test/runtime.rs`
  - Remove the now-unused `lcm_ingest_raw_message_for_test`.
  - Add `lcm_retained_raw_revision_for_test`, a thin read over `tracedecay_lcm::summary_convergence::candidate_for_session(...).stale_from_store_id` on a read snapshot.

## Test plan

- [x] `cargo test -p tracedecay --test dashboard_api_test --features test-transport` — all 65 tests pass concurrently (default threads).
- [x] The two target tests pass isolated (`--test-threads=1`), 3/3 repeated runs; before the change both failed deterministically with `left: Number(0) / right: 1`.
- [x] Guard verified: temporarily restoring the second raw write makes the fixture fail at the new assertion with `left: Some(1) / right: None` ("seeded LCM messages must not leave a retained raw revision pending invalidation").
- [x] `cargo clippy -p tracedecay --test dashboard_api_test --features test-transport -- -D warnings` clean; `cargo fmt --all -- --check` clean.
- [x] Commit subject passes `npm run lint:commit`.
- [ ] Full `cargo nextest run --workspace --no-fail-fast` not run locally (test-only change to one integration binary); CI covers it.

## Checklist

- [ ] `CHANGELOG.md` updated — not applicable (test fixture only, no runtime behavior change)
- [x] Breaking changes documented (none)

<div><a href="https://cursor.com/agents/bc-995d2e46-5111-4335-97a6-2fc432c36814?cursor_ref=pr_footer&cursor_cta=open_in_web"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a>&nbsp;<a href="https://cursor.com/background-agent?bcId=bc-995d2e46-5111-4335-97a6-2fc432c36814&cursor_ref=pr_footer&cursor_cta=open_in_cursor"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;</div>


ScriptedAlchemy:
Superseded by the production-authority fix integrated on #707 as `ad0dd1bbf`. The fixture was not the sole defect: `ef3c912d1` made the temporal store hide every Current-mode summary for any dirty-raw row, while `51d6dd5e9` had already corrected the same predicate in LCM replay/condensation to hide only when pending `summary_node` invalidation proves a partially applied stale-summary walk. Protection hydration legitimately creates raw-revision work even for freshly ingested sessions, so removing one fixture write would leave the two production read authorities inconsistent.

`ad0dd1bbf` ports the canonical partial-invalidation predicate to all seven temporal read sites and adds direct regressions for both states (raw-message work only serves the available summary; pending summary-node work hides it). Evidence: dashboard 65/65, temporal-store 135/135, LCM 159 passed / 2 ignored, strict clippy. No fixture behavior or expected count was weakened.

Issue #926, opened 2026-09-06
[707/P1] Keep analyzer capabilities and diagnostics readiness aligned with the live client


ScriptedAlchemy:
Priority: P1
Blocks undraft #707: yes
Labels: bug, ci, P1, v2, macos (suggested)

### Symptom

**Classification:** Product LSP lifecycle plus one timing-sensitive fixture.

Linux providerUnavailable after initialization; Linux/macOS diagnostics_read remains retryable; macOS fake partial-frame barrier missed.

Baseline: [run 34039753429](https://github.com/ScriptedAlchemy/tracedecay/actions/runs/34039753429), head `9d17537b4`. Exact test rows below come from the archived reports; Dashboard comes from its job log.

- **linux** — `tracedecay::runtime_acceptance_suite runtime_surface_acceptance::production_lsp_negotiates_and_projects_canonical_context` — failed initial and retry ([job](https://github.com/ScriptedAlchemy/tracedecay/actions/runs/34039753429/job/101514387873)).
- **linux** — `tracedecay::runtime_acceptance_suite application_production_reachability::operation_family_executes_through_cli_mcp_and_http` — failed initial and retry ([job](https://github.com/ScriptedAlchemy/tracedecay/actions/runs/34039753429/job/101514387873)).
- **macos** — `tracedecay::runtime_acceptance_suite application_production_reachability::operation_family_executes_through_cli_mcp_and_http` — failed initial and retry ([job](https://github.com/ScriptedAlchemy/tracedecay/actions/runs/34039753429/job/101514387881)).
- **macos** — `tracedecay-lsp::analyzer_runtime broker_refresh::broker_cancels_partial_refresh_without_poisoning_warm_client` — passed retry; CI failed-on-flaky ([job](https://github.com/ScriptedAlchemy/tracedecay/actions/runs/34039753429/job/101514387881)).

### Root cause

On Linux, the LSP test passes initialization but documentSymbol returns providerUnavailable. Both Linux/macOS operation-family tests loop on diagnostics pre-admission unavailability. That differs from the macOS initialize-root mismatch and Windows project-open failures, which are assigned elsewhere. Analyzer supervision defects have since been patched; the original report cannot prove those cases still fail.

### Exact fix

[...]

Issue #913, opened 2026-09-06
test: regression gates for the 2026-09-06 lifecycle and identity fixes


ScriptedAlchemy:
Regression coverage the architecture review asked for on the changes landed 2026-09-06 (tip c214ba423), to be added as tests without changing timeouts, retries or assertions:

- **Semantic source scope is the checkout-scoped shard (c1c6dd6dc / 21af59bd9):** prove two linked worktrees of one project keep isolated semantic bindings and retention follows the exact publication dependency; give stores that still hold a Branch-form `semantic_vector_source_scope_bindings` row an explicit typed rebuild/refusal path instead of silent rebinding (the refusal now names both tuples; no migration is written).
- **Published-generation reader snapshot (851344cce):** concurrent publish / retire / read coverage — a consistent historical snapshot must not be mistaken for a guarantee that its graph stays retained after the read ends; keep the subsequent verified-generation / lease checks.
- **Context Scout read-first startup (03d61aac0):** discovery is not claim acquisition — two concurrent startups seeing the same entry must yield exactly one successful durable claim.
- **Disabled linked-route deferred owners (14249028f):** the disabled check is at spawn time; prove a disabled → enabled configuration transition recreates the dependent owners, or explicitly require a route remount, so automatic-indexing disablement is never permanent loss of a supported capability.
- **#880 rebuild boundary (e85119957):** reset → re-ingest → real temporal/graph query, including rejection of pre-reset cursors and stale graph coverage; row-count success alone is insufficient now that assertions/supersession are deleted as derivations.
- **Federated roots (3f8f0ad66 / 5fdca6182):** keep negative tests for foreign profiles, substituted factories and duplicate/missing roots; do not apply the checkout-identity relaxation to this authorization pairing.
- **Vector-store facade (c214ba423):** the current-thread liveness regression exists; add the CPU-quota acceptance the review named — a contending semantic operation must not prevent the transaction holder from committing, bootstrap from responding, or route cancellation/shutdown from completing, and no progress may depend on the 30 s idle lease.

Targets `codex/tracedecay-total-redesign-plan-reopened` (#707). Refs #753, #901, #880, #912.

ScriptedAlchemy:
Closing with evidence on `codex/tracedecay-total-redesign-plan-reopened` (all landed via cb300fa62 and its chain; every test passes three consecutive runs, no ignores/retries/sleeps):

[...]

Issue #896, opened 2026-09-05
test(macos): residual deadline and unclassified failures after the canonicalization fixes


ScriptedAlchemy:
After the macOS-only fixes landed on PR #707 (4ec65ea4b projectless resolved roots, c31b086c8 bootstrap keys, 703d9d063 codex fixtures, non-UTF-8 probes ff17aba14/1650b56b9/0a091549c, d63f499a2 `/dev/full` probe, 6bcf1ad8e paging contract, bdefc4e45 `/usr/bin/grep`), two macOS-only families from run 33897253099 remain unaddressed:

Timing on the slower runner (4): `daemon::tests::bootstrap::direct_tool_cache_miss_returns_warming_while_project_opens_in_background` (bootstrap.rs:3279 `Elapsed`), analyzer_runtime.rs:87, controlled_invocation_tests.rs:974, activation_tests.rs:174 (`did not reach phase … before deadline`). Fix what stalls or move to the shared slow-runner budget; do not just raise deadlines.

Unclassified (4): mcp/server/rmcp.rs:946 (`left: Null right: Bool(true)`), background_refresh_writer_tests.rs:46 (`retained background refresh graph is stale`), tests/scheduler_config.rs:701, and the sessions backfill test `non_utf8_canonical_worktree_resumes_exactly_then_fails_typed_publish` (git_correlation/backfill/bounded/tests.rs:799) which still needs the runtime non-UTF-8 probe that 2551e5b3c added for the fixture.

Parsed failure sets are in the sweep artifacts; see #863 for the shared Linux+macOS set.

ScriptedAlchemy:
## macOS-only residual set, run 33978239206 (job 101338647437)

Method: took the final `FAIL`/`TMT` lines from the macOS job and subtracted the Linux job's (101338647423) set. 44 macOS failures, 28 shared with Linux (those belong to #863), **16 macOS-only**. Grouped by root cause below. Linux per-test durations come from that run's `nextest-junit-Linux` artifact.

### F1 — Fixture temp roots keep the `/var` symlink (6 tests) — fixed, `28ccccbe3`

macOS puts `TempDir` under `/var/folders/...`, and `/var` is a symlink to `/private/var`. Production canonicalizes a project root before it hashes the code-index scope, keys the retained project graph, or decides whether an ignored dependency escaped the worktree — so a fixture handing out the symlinked spelling names a different project than the one the runtime mounted.

[...]

PR #958, opened 2026-09-06
fix(lsp): keep the analyzer supervisor aligned with the shared client

ScriptedAlchemy:
## Summary

- Move the analyzer's stdio client slot and its `AnalyzerSupervisor` into one `SharedAnalyzerClient` owned by the broker, so the diagnostics refresh lane and the semantic request lane describe the same mounted process and the same incarnation.
- The refresh lane now records its own lifecycle (start attempt, ready, answered refresh as service, retirement on failure or abort) and honours restart exhaustion instead of spawning behind the semantic lane's back. New non-budget `AnalyzerEvent::Retired` covers a client the lane released without a result.
- Give the two phase-gated fake-server tests the existing loaded-runner start budget so a slow `python3` spawn cannot make the refresh give up before the barrier is reached (the macOS `partial-frame-flushed` flake).

## Motivation

Closes the remaining gaps on #926 against the current `codex/tracedecay-total-redesign-plan-reopened` tip. The baseline failures in #926 had two mechanisms that have since landed on the branch and are **not** duplicated here:

- Linux `providerUnavailable` after a refresh-lane failure → `c33b975d2` (supervision keyed by root URI) and `8e5d7b048` (open the addressed document before semantics).
- `diagnostics_read` perpetual pre-admission retry → `40f019653` (unpublished scope is terminal).
- The positive acceptance job installing the pinned `rust-analyzer` component → `d0db23047`.

What was still true at tip, verified by reading `collect_refresh_batch`, `finish_refresh_snapshot`, and `StdioLspSemanticAuthority`:

1. **`Ready` with no client.** The refresh lane shares the semantic lane's slot but not its supervisor. On any refresh failure it cleared the slot (killing the process) and told the supervisor nothing; on an empty slot it spawned a new process the supervisor never saw. Readiness reported `Ready` over a dead process, and a replacement started by the refresh lane was served on the retired incarnation's attempt, so its requests could forgive a restart budget that process never earned.
2. **Two processes, two truths.** `finish_refresh_snapshot` removed the slot from the broker map on failure, so the next `prepare_refresh` minted a fresh slot while the semantic authority kept its orphaned one — two analyzers for one root, with the semantic lane restarting the first.
3. **Exhaustion was one-sided.** `MAX_ANALYZER_RESTARTS` documents exhaustion as "a stable health state, not an invitation for a bridge or client to start its own analyzer", but only the semantic lane honoured it.
4. **macOS `broker_cancels_partial_refresh_without_poisoning_warm_client`** already used an explicit phase signal; the flake was that the first refresh ran under `collect_diagnostics(500ms)`, whose derived initialize budget is 3s, so a starved `python3` spawn made the refresh give up and kill the fake before it could reach the phase, and the barrier then waited out its own 10s deadline (also listed on #896).

## Changes

- `crates/tracedecay-lsp/src/analyzer/broker/shared_client.rs` (new): `SharedAnalyzerClient` = client slot + supervisor, with the attempt-fenced `begin_start` / `mark_ready` / `record` moved here from `semantic_authority.rs`. Both lanes lock the slot before consulting the supervisor, so a start still has exactly one owner.
- `broker/refresh.rs`: `collect_refresh_batch` claims an attempt before spawning, refuses to spawn when the slot is terminal, marks ready, records an answered refresh as `Ready` (service), and concludes via a `RefreshUseOfAnalyzer` drop guard so an aborted partial refresh is recorded rather than discovered later. A refresh ending without a publication records `Retired`.
- `provider.rs`: `AnalyzerEvent::Retired` — `Starting | Ready → RestartBackoff`, no restart budget spent, coverage `analyzer-retired`. The lane cannot distinguish a silent analyzer past the quiet window from a dead one at that boundary, and charging it would let the refresh lane's own 2s quiet window exhaust the semantic surface under load (the regression `c33b975d2` warned about).
- `broker.rs`: map holds `Arc<SharedAnalyzerClient>`; `semantic_authority_if_available` keys by the canonical workspace root exactly as `prepare_refresh` does; the `Err` arm of `finish_refresh_snapshot` no longer drops the shared entry.
- `semantic_authority.rs`: uses the shared object; the old "client reached the slot without this lane seeing a start" special case is gone because the refresh lane now records its starts.
- Tests: supervisor unit test for `Retired`; `refresh_lane_lifecycle_is_recorded_on_the_shared_supervisor` (start visible, retirement visible, restart is a new incarnation, answered refresh counts as service); `an_exhausted_restart_budget_stops_the_refresh_lane_from_respawning`; the cancel-partial test now asserts the drop-guard evidence and that recovery is a new attempt; readiness root compared by path since the slot is keyed canonically; `phase_gated_fake_lsp_timeouts()` for the two barrier tests.

Not changed, deliberately: the refresh lane's own quiet-window/kill policy, and the broker-level engine backoff (a crash loop observed only by the refresh lane remains unbounded there, as before).

## Test plan

- [x] `cargo test -p tracedecay-lsp` — 213 unit + 36/4/4/4 integration, green on three consecutive runs
- [x] `cargo clippy -p tracedecay-lsp --all-targets` — no warnings from this crate (the only workspace warning, `GraphProjectionLabelPage` in `tracedecay-graph-db`, is pre-existing on the base branch)
- [x] `cargo test -p tracedecay-usecases --lib -- lsp_support lsp_runtime dashboard_diagnostics` — 45 passed
- [x] `cargo test -p tracedecay-cli --features test-transport --test opencode_one_analyzer_journey` — 2 passed (one-analyzer-per-language ownership still holds over the shared slot)
- [x] `runtime_acceptance_suite` `production_lsp_negotiates_and_projects_canonical_context` + `operation_family_executes_through_cli_mcp_and_http` with the real `rust-analyzer 1.97.1` component on Linux: pass idle (9.4s), and pass on 3/3 runs with 6 CPU hogs pinning a 4-core box (17–19s), which is the load condition behind the baseline `providerUnavailable`
- [x] `npm run lint:commit -- --from origin/codex/tracedecay-total-redesign-plan-reopened --to HEAD`
- [ ] `cargo nextest run --workspace --no-fail-fast` (CI)

## Checklist

- [ ] `CHANGELOG.md` updated (under `[Unreleased]` if no version bump) — not touched; #707 batches its own changelog
- [x] Breaking changes documented (if any) — none; `StdioLspSemanticAuthority::from_shared_client` is `pub(crate)`, `AnalyzerEvent` gains a variant

Refs #926, #913, #896. Targets #707's branch; do not undraft or merge #707.

<div><a href="https://cursor.com/agents/bc-74178037-4bc2-4d72-95e0-f8d99db9c822?cursor_ref=pr_footer&cursor_cta=open_in_web"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a>&nbsp;<a href="https://cursor.com/background-agent?bcId=bc-74178037-4bc2-4d72-95e0-f8d99db9c822&cursor_ref=pr_footer&cursor_cta=open_in_cursor"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;</div>


ScriptedAlchemy (commented):
## Architecture review — shared owner is correct; REVISE failure classification

Reviewed `5fa4d2aa57a2d9fa6c973037f747470b67b50d37`; source review only, exact-head CI pending. Keep `SharedAnalyzerClient`: one slot, one supervisor, one incarnation across semantic requests and refreshes removes the competing owner.

**P1:** `broker/refresh.rs::collect_refresh_batch` maps every post-start diagnostics error to the non-budget `AnalyzerEvent::Retired`, although it returns `RefreshFailure::crashed`. A process that initializes and then repeatedly fails during refresh can keep restarting without consuming the shared restart budget. The PR body explicitly acknowledges this remaining loop. Separate known process exit/transport/protocol failures from cancellation or a publication/quiet-window policy decision. Charge genuine failures to the same supervisor; keep neutral retirement for an owner intentionally relinquishing a client. Do not charge a healthy client's mere lack of diagnostics as a crash.

Add a refresh-only crash-loop regression proving exhaustion without any semantic requests, plus cancellation/quiet-window coverage proving those events do not spuriously exhaust the analyzer. If the client error type currently erases the distinction, preserve it at the client boundary rather than adding another broker budget.

Keep the healthy-uptime-plus-served-requests reset, attempt fences, and root authorization checks when integrating current #707. Verify stale refresh completion cannot retire a newer incarnation and that aborting a partial refresh reaps its process before a successor is served. The loaded real-analyzer results are useful; rerun the affected Windows/macOS paths on the final merge tree. #707 remains draft.

ScriptedAlchemy:
Addressed the architecture review at final head `452cd0b88` (classification implementation `29fe0b289`) after merging explicit base floor `5204bfadd`.

- Added typed post-initialize `LspRefreshError` classes at the stdio client boundary. Process exit/EOF transport, write transport, and malformed/partial protocol failures now map to budget-consuming shared-supervisor events; no-publication quiet-window policy maps to neutral `Retired` and diagnostics state `Available`, not `Crashed`.
- Kept the accepted single `SharedAnalyzerClient` authority. Its one slot is now explicitly `Vacant | Live | Reaping`, and every successor waits for the retired process reaper before it can start or serve. Attempt fencing, URI/root authorization, and healthy-uptime-plus-served-request reset behavior are unchanged.
- Preserved terminal typing: restart exhaustion refuses diagnostics starts and `upstream_capabilities()` returns `AnalyzerRuntimeError::Unavailable`; a neutral quiet-window retirement remains restartable, and capability negotiation plus the next diagnostic refresh reuse the same incarnation.

Falsifiable coverage added/strengthened:
- `refresh_only_post_initialize_crash_loop_exhausts_restart_budget`
- `refresh_quiet_window_stays_budget_neutral_and_reuses_live_client`
- `broker_cancels_partial_refresh_without_poisoning_warm_client` (repeated beyond the restart budget)
- `broker_bounds_lsp_document_write_hangs` and `broker_drops_lsp_client_after_partial_diagnostics_frame_timeout`
- `stale_refresh_retirement_cannot_retire_a_newer_incarnation`
- `successor_waits_for_the_retired_client_reaper`

Final-tree evidence:
- `cargo test -p tracedecay-lsp` — pass (full crate suite)
- `cargo clippy -p tracedecay-lsp --all-targets -- -D warnings` — pass, 0 warnings
- `cargo build -p tracedecay-cli --bin tracedecay` with the seeded dashboard digest — pass
- `cargo test -p tracedecay --features test-helpers --test runtime_acceptance_suite -- runtime_surface_acceptance::production_lsp_negotiates_and_projects_canonical_context application_production_reachability::operation_family_executes_through_cli_mcp_and_http --exact --nocapture` — 2 passed, 0 failed
- `npm run lint:commit -- --from origin/codex/tracedecay-total-redesign-plan-reopened --to HEAD` — pass

The push triggered hosted CI for `452cd0b88`; the scope/slice jobs that gate the Windows/macOS matrix are still pending: https://github.com/ScriptedAlchemy/tracedecay/actions/runs/34074681757. I am not claiming those hosted OS results from this Linux lane. The PR remains unmerged.

ScriptedAlchemy (commented):
## System-level re-review at `452cd0b88` — failure classification fixed; reaping still cancellation-unsafe

The previous blanket-`Retired` finding is resolved in the inspected source. `LspRefreshError` distinguishes NoPublication, ProcessExited, Transport and InvalidResponse, and the refresh lane maps them to the corresponding shared-supervisor events. The slot-owning refresh guard also improves abort handling. Do not revert those changes or continue describing all refresh errors as uncharged.

### P1 — a cancelled reaper waiter makes the slot reusable before settlement

In `crates/tracedecay-lsp/src/analyzer/broker/shared_client.rs::client`, `slot.take_reaper()` replaces `Reaping` with `Vacant` and moves the future into a local variable before `reaping.await`. If this caller is aborted at that await, both the local reaper and mutex guard are dropped, leaving `Vacant` in the shared slot. The next semantic/refresh request can acquire it and start a replacement without observing completion of the previous process's reap. An error from the reaper also exits through `?` with the same vacant state. This contradicts the new invariant that a successor cannot be started or served before the retired process is reaped. Kill-on-drop is not evidence of joined exit.

Keep the reaper in the shared slot while polling it, or transfer it once to a retained settlement task with shared completion. Cancellation of a waiter must not cancel ownership of process settlement. On a reap error, retain an explicit failed/retiring state that cannot authorize a new spawn until the old process's termination is known. Do not add a second analyzer registry.

Extend the existing `successor_waits_for_the_retired_client_reaper` barrier test: wait for reaper entry, abort the first `client()` waiter, start a second waiter, and assert it still cannot acquire until the same reaper is released. Add the analogous reaper-error case. The current test only covers an uninterrupted waiter and therefore misses the state-loss interleaving. I have not executed these tests here.

### System integration

The broker's stable root-keyed slot must outlive both lanes, and retirement must cover process exit as well as supervisor state. Preserve attempt fencing and the distinction between a caller cancellation, a quiet-window policy choice and a genuine process failure. Re-run the production one-analyzer journey through refresh failure -> cancelled recovery waiter -> semantic recovery, including shutdown, after the fix. The three returned exact-head workflows are cancelled, so native-platform acceptance is not established by this review.

Issue #917, opened 2026-09-06
[707/P0] Make strict indexing converge without confusing current text with graph seating


ScriptedAlchemy:
Priority: P0
Blocks undraft #707: yes
Labels: bug, ci, P0, v2, windows (suggested)

### Symptom

**Classification:** Product indexing and scale; mixed repeat failures and timing-sensitive cases.

PR checkout dogfood (job 101514387912): warming/partial_refresh_in_progress, text bulk_commit, graph pending; plus scheduler/retained-generation failures below.

Baseline: [run 34039753429](https://github.com/ScriptedAlchemy/tracedecay/actions/runs/34039753429), head `9d17537b4`. Exact test rows below come from the archived reports; Dashboard comes from its job log.

- **linux** — `tracedecay daemon::code_index_runtime_graph_activation_tests::dirty_restart_recovers_the_retained_graph_before_rebuilding_its_successor` — passed retry; CI failed-on-flaky ([job](https://github.com/ScriptedAlchemy/tracedecay/actions/runs/34039753429/job/101514387873)).
- **windows-2** — `tracedecay-code-index-runtime code_index_scheduler::tests::noop_reconcile_tests::unchanged_reconcile_retries_semantic_admission_for_the_serving_generation` — passed retry; CI failed-on-flaky ([job](https://github.com/ScriptedAlchemy/tracedecay/actions/runs/34039753429/job/101514939956)).
- **windows-2** — `tracedecay-code-index-runtime code_index_scheduler::tests::mount_with_retained_generation_verifies_cadence_promptly` — failed initial and retry ([job](https://github.com/ScriptedAlchemy/tracedecay/actions/runs/34039753429/job/101514939956)).
- **windows-2** — `tracedecay-code-index-runtime code_index_scheduler::tests::poisoned_scheduler_lock_does_not_retire_the_background_worker` — failed initial and retry ([job](https://github.com/ScriptedAlchemy/tracedecay/actions/runs/34039753429/job/101514939956)).
- **windows-2** — `tracedecay-code-index-runtime code_index_scheduler::tests::resident_memory_graph_refusal_seats_text_serving_without_graph` — passed retry; CI failed-on-flaky ([job](https://github.com/ScriptedAlchemy/tracedecay/actions/runs/34039753429/job/101514939956)).
[...]

Issue #800, opened 2026-09-03
perf(code-index): make incremental capture and seal delta-proportional


ScriptedAlchemy:
## Problem

A one-file sync still performs repository-sized work in two authorities:

- `crates/tracedecay-code-index-runtime/src/code_index_scheduler/classification.rs::candidate_paths` and `capture_authoritative_snapshot_with_active_generation_reuse` read, sanitize, and hash every tracked path.
- `crates/tracedecay-code-index/src/production/partitioned_codec.rs::encode_partitioned_sealed` and runtime publication reserialize every file segment, hash it during encoding, re-hash before publish, then reread/re-hash unchanged content-addressed segments.

The local O(files²) ordinal lookup has been fixed separately. Reusing prior capture bytes and segment descriptors changes persistence/lifetime authority and needs an explicit product design.

## Required design

- Bind unchanged capture rows and segment descriptors to the exact prior generation and source identity.
- Read/encode/hash only classified changed files during a one-file sync.
- Preserve fail-closed validation, cancellation, rollback, and content-addressed identity.
- Do not weaken digest checks or increase budgets.

## Acceptance evidence

- Production efficiency-scorecard incremental journey at 1k and 10k unchanged files with one edit.
- Captured bytes, encoded segment bytes, and existing-segment bytes read remain proportional to the edited file/segment, not repository size.
- Cold index, incompatible rebuild, cancellation, and tamper tests remain unchanged in correctness.

ScriptedAlchemy:
Fresh PR #707 dogfood evidence on `a41f511dd` (strict 600s gate, existing `--profile perf`, no timeout change):

- Timed out at 600.174s with 4,895/4,895 files and 453,770 chunks; text state was still `verification`, graph remained pending.
- Production-scale Hotpath decomposition (688.56s observed): `code_index.generation.publish.segment_encode` = 377.04s in one call (54.76%); `query.artifact.batch.scheduler_wake` = 28.27s/8; query artifact open/schema verification ~=17.98s/17.96s.
- Dominant path is `crates/tracedecay-code-index/src/production/partitioned_codec.rs` plus capture/seal publication. Graph projection did not start and is not causal.
- The existing delta-only scope may not improve fresh `parent=None` publication; please include first-publication encoder throughput or leave that exact remainder open.

No source change or deadline/profile-only workaround was retained from this diagnosis.

ScriptedAlchemy:
[...]

PR #962, opened 2026-09-06
fix(code-index): attribute strict readiness phases; refuse graph seat over RSS watermark (#917)

ScriptedAlchemy:
## Summary

- Attributes the strict-readiness journey phase by phase (evidence below) instead of assuming semantic publication (#837): on the 4925-file checkout the text projection alone consumes the 600 s budget on a dev-profile binary, and the phase that can never fit is the **native graph seat**, which grew the daemon from 6.5 GiB to 15.4 GiB and was OOM-killed on a 16 GiB host.
- Sealed graph publication now answers to the daemon's measured-RSS admission authority: over the high watermark it aborts at its existing cancellation checkpoints and reports the typed `GraphBudgetKind::ResidentMemory` budget. The scheduler already classifies that name as a graph **refusal** that keeps exact/lexical serving seated (no retry into the same wall); the persistent activation path now records the `Refused` state so `status` names the reason instead of a `pending` graph that will never seat.
- The dogfood readiness log records the identities that keep text readiness and graph seating distinct (sealed source digest, generation, coverage/staleness, graph state + typed reason, graph-statistics state, projection elapsed and commit latency, daemon RSS + kernel peak), and prints one `tracedecay_ci_readiness_phases` line on both PASS and timeout with first-progress / text-complete / graph-ready / status-current times and peak daemon RSS.

Closes nothing on its own; addresses #917. Targets `codex/tracedecay-total-redesign-plan-reopened` (#707 stays draft).

## Motivation

#917: PR checkout dogfood strict readiness (600 s) has failed on every non-cancelled run of this branch (27/27). The supplied summary stopped at `bulk_commit`, `partial_refresh_in_progress`, graph `pending`, which does not identify the blocking phase or owner.

### Evidence (local reproduction, dev-profile binary, 4 cores / 16 GiB, no other load, 5 s probes with daemon RSS sampled)

| t (s) | phase | daemon RSS |
|---|---|---|
| 0–20 | capture / extract / chunk (no progress phase published) | 0.3 → 3.5 GiB |
| 20–76 | seal + publish (no progress phase published) | 3.5 GiB flat |
| 76–96 | sealed-source scan (`source_scan files=0/0`) | 3.9 → 8.1 GiB |
| 96–399 | text bulk projection: 4925 files, 3918 pages, 880 MB payload, ~0.9 s/commit (CI: ~2.2 s) | 8.4 GiB flat |
| 399–424 | `index_build` | 8.7 GiB |
| 424–434 | `verification` | 8.6 GiB |
| 434 | text `ready`, `coverage=complete`, `status=warming`, graph `pending` | 11.9 GiB peak, trimmed to 6.5 GiB |
| 450–780 | native graph seat (`publish_verified_snapshot`) | 6.5 → **15.4 GiB → OOM-killed** (`anon-rss:15420332kB`), graph still `pending`; grafeo WAL wrote 73 × 64 MiB segments |

The daemon's own sampler logged `daemon_resident_memory_over_budget observed=10.95 GiB high_watermark=11.33 GiB` at t≈514 s ("refusing new growth") — the publication did not consult it and kept allocating for another ~4 minutes until the kernel killed the process, which took the already-complete text serving down with the optional graph.

CI baseline (run 34039753429 and the four runs before it) shows the same shape at ~1.6× slower per commit: ~105 s with `phase=None`, then text projection at 12–20 files/s ending at 4375–4938/4938 files or entering `index_build` at t≈597 s. Graph seating was never reached in CI; locally it needs > 330 s and > 15 GiB. Semantic publication is not on this path.

### What this PR does *not* do (by design, per #917)

- Does not relax freshness, accept a "latest generation" blindly, raise the 600 s bound, force rebuilds, or make semantic wait on graph seating.
- Does not change the strict validator. Strict readiness still requires text `current` **and** graph `ready`.

### What to expect from dogfood/tests after this PR

- The attempts log and the `tracedecay_ci_readiness_phases` line will name the phase a timeout died in and the peak daemon RSS. On the current CI runner the expected verdict is still a timeout, now attributed to: text projection ≈ 500 s on the dev binary, then graph seat.
- Once text completes on a runner (or with a faster build), the graph seat on this repo will report `code_graph_serving.state=refused` with the resident-memory reason within ~30–60 s of crossing the watermark instead of OOM-killing the daemon after minutes; `status`/text tooling keeps working.
- Strict readiness for this repo on a 16 GiB runner is gated by graph-publication memory (scale work, #852) and by text projection throughput on an `opt-level = 1` binary; both are outside this PR's fix scope and are called out below.

## Changes

- `crates/tracedecay-graph-db/src/error.rs`: `GraphBudgetKind::ResidentMemory` (`"resident_memory"`), round-trips through `from_name`.
- `crates/tracedecay-store-runtime/src/session_registry/code_graph.rs`: `ResidentMemoryGuardedGraphCancellationV1` wraps the request cancellation used by `publish_verified_snapshot`; trips (with one `code_graph_publication_refused_resident_memory` warn) when the pressure cell is `OverBudget`. A `Cancelled` caused by the watermark (and not by the request) is mapped to `BudgetExhausted { ResidentMemory, limit }` at the build claim, the manifest build, and the publication. `RetainedCodeGraphRuntimeV1` carries the cell (bound from the manifest provider; test-only override).
- `crates/tracedecay-store-runtime/src/session_registry/code_graph_manifest.rs`: the provider keeps and exposes the pressure cell it already registers its reclaimer on.
- `crates/tracedecay-code-index-runtime/src/code_index_scheduler/graph_activation.rs`: persistent activation records `Refused` on the resident-memory budget before propagating; shared reason constant reused by the injected test path; `is_graph_activation_refusal` compares against the enum name.
- Test: `sealed_publication_refuses_over_the_resident_memory_watermark` — real sealed generation through the production registry with an isolated pressure cell: over the watermark → `BudgetExhausted { ResidentMemory }`, journal untouched, classified as a graph refusal; request cancellation keeps its identity; back under the low watermark the same runtime publishes and advances the verified head.
- `scripts/with-isolated-tracedecay-daemon.sh`: exports `TRACEDECAY_DAEMON_PID`.
- `scripts/lib/portable_process.py`: `resident-memory --pid` (Linux `/proc`, `ps` elsewhere; exited process → `unavailable`, never an error).
- `scripts/ci-pr-dogfood-smoke.sh`: per-attempt attribution columns + daemon memory; `report_readiness_phases` on ready and timeout.
- `scripts/test-pr-dogfood-portability.py`: asserts the phases line on both outcomes and the helper subcommand.

## Test plan

- [x] `cargo test -p tracedecay-store-runtime --lib -- sealed_publication_tests` (6 passed incl. the new refusal test, 1 pre-existing ignored measurement)
- [x] `python3 scripts/test-pr-dogfood-portability.py` (15 passed), `scripts/test-check-pr-dogfood-output.py`, `scripts/test-efficiency-scorecard.py`
- [x] `npm run lint:commit -- --from origin/codex/tracedecay-total-redesign-plan-reopened --to HEAD`
- [ ] `cargo check --workspace --all-targets` (running)
- [ ] `cargo nextest run -p tracedecay-code-index-runtime` scheduler graph-activation tests (running)
- [ ] Local dogfood re-run with the patched binary: expect text `ready` at ≈430 s, then `graph=refused:…resident-memory…` instead of an OOM kill; timings and peak RSS to be posted here

## Out of scope / follow-ups (evidence-backed)

- **Graph publication memory** is what blocks strict readiness on this repo on a 16 GiB host; bounding it is #852-scale work (corpus-sized manifest + staged rows + engine WAL each add a copy on top of the 6.5 GiB text baseline).
- **Text projection throughput**: ~500 s in CI for 3900 pages is dominated by per-batch SQLite commits (`journal_mode=DELETE`, `synchronous=NORMAL`, ~2.2 s per ~24-page batch on the runner). The dogfood builds the CLI with the default dev profile (`opt-level = 1`); the repo's `perf` profile exists for "generation-scale code-index … at production-like speed" and was measured at ~1.5× faster for this phase. Switching the dogfood build to it is a CI-config decision, deliberately not made here.
- **Windows scheduler flakes** listed in #917 (`unchanged_reconcile_retries_semantic_admission…`, `semantic_schedule_*`) are 3 s test-timeouts on shards that also had 80–114 unrelated failures each; they are a separate Windows cluster and no timeout was raised here.

## Checklist

- [ ] `CHANGELOG.md` updated (under `[Unreleased]` if no version bump) — not applicable for this branch's draft PR flow
- [x] Breaking changes documented (if any) — none; `GraphBudgetKind` gains a variant

<div><a href="https://cursor.com/agents/bc-578e34cc-a2c3-4d87-aef2-318d7593bb2d?cursor_ref=pr_footer&cursor_cta=open_in_web"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a>&nbsp;<a href="https://cursor.com/background-agent?bcId=bc-578e34cc-a2c3-4d87-aef2-318d7593bb2d&cursor_ref=pr_footer&cursor_cta=open_in_cursor"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;</div>


ScriptedAlchemy (commented):
## Architecture review — HOLD for the pressure-trip invariant and scale proof

Reviewed `8a049a7a1a43b5fffdf098fc9fe0c299119cd4dc`; source review only; exact-head CI is pending. Refusing optional graph growth while retaining verified text is the correct containment boundary. This is not completion of strict readiness or #852.

- **P1 — make the publication's pressure cancellation monotonic.** In `session_registry/code_graph.rs`, `ResidentMemoryGuardedGraphCancellationV1::is_cancelled()` sets `tripped` under OverBudget but returns false on a later call if the pressure cell recovers. Once this particular publication has tripped, it must continue to return true; only a new admitted attempt should resume under recovered pressure. Keep explicit request cancellation's classification precedence. Add a test that trips, moves below the low watermark, and verifies the old attempt remains cancelled while a new attempt can publish.
- **P1 acceptance — test an in-progress trip, not only over-budget admission.** Drive a real publication through at least one staging/checkpoint phase, then raise pressure. Verify no serving head advances, no partial graph is exposed, all abandoned staging/leases are recoverable or retired, and the prior text route still answers. The patched large-corpus reproduction must show bounded overshoot and a typed refusal before process death. A periodically sampled watermark plus unbounded work between checkpoints is not a hard memory ceiling.
- **P1 integration — consolidate #963's overlapping dogfood instrumentation.** Both PRs change the harness, PID export and phase-report format. Retain one PID variable, one portable memory reader and one phase schema; preserve #963's BusyFollowUp and interruption fixes. Do not merge both instrumentation implementations independently.

Keep the strict validator and 600-second bound unchanged. Record the graph refusal separately from text-current status; do not call a refused graph strict-ready. After containment, the remaining work is measured text throughput and graph peak-memory reduction, not larger timeouts. No #707 draft-state change is authorized.

ScriptedAlchemy (commented):
## System-level re-review at `8a049a7a1` — retain text availability, make interruption monotonic

Re-read the pressure wrapper together with `GraphPublicationProbe::interruption` / `try_begin_commit`. The earlier latch finding is still present: `is_cancelled()` sets `tripped` when pressure is OverBudget, but does not consult that bit on subsequent calls. A pressure transition below the watermark therefore changes this same attempt's answer from cancelled to not-cancelled, while `refused_by_resident_memory()` still says it was refused.

**P1 — latch the refusal for this publication attempt.** Check the previously tripped bit as well as current pressure. A new attempt may be admitted after pressure recovers; the already-interrupted attempt may not revive. Preserve request/lifecycle cancellation attribution and the irreversible commit boundary. This is a directly observable inconsistency in the probe, not a claim that I reproduced an OOM or that every sequential checked loop resumes after returning an error.

Add a deterministic guard-level regression: normal -> pressure high -> `is_cancelled()==true` -> pressure low -> the same guard remains cancelled; a separately created guard is admissible. Also exercise a real mid-publication trip rather than only starting above the watermark, checking that the verified head and journal reflect the actual commit boundary and previously seated text stays usable.

**System boundary:** resident-memory refusal is an availability containment mechanism, not proof that generation-scale work is memory-bounded. The source manifest, staged graph rows, retained text and engine allocations can coexist. Attribute those owners and measure high-watermark-to-checkpoint overshoot on the integrated path; changing the sampler/timeout or claiming the refused graph is ready would hide the actual capacity problem. Semantic admission must continue to use accepted artifact/source-content/capability/privacy compatibility, not become dependent on native graph seating because strict dogfood also checks graph readiness.

Consolidate the script/PID/memory reporting with #963, but preserve the pressure change and the lost-wake fix as independently testable changes. A single phase report should distinguish text current, graph refused and semantic readiness rather than reporting one aggregate mounted flag. The existing strict validator should remain strict.

Exact-head CI is pending and profile/coverage queued. Source review only; no native/full-corpus execution claimed. #707 remains draft.

ScriptedAlchemy:
Superseded by #963; no unique #962 behavior was dropped.

Mapping:
- `e8cf06f04` (resident-memory graph refusal) -> #963 `4a93cc22c`, with monotonic per-attempt trip, cancellation-precedence, in-progress abort, and strict text-vs-graph assertions in `890883e7f`.
- `ac2f129f0` (dogfood PID/RSS/phase evidence) -> #963 `499d6d67e`, consolidated with #963's instrumentation into exactly one `TRACEDECAY_DAEMON_PID`, one portable `resident-memory` reader, and one `tracedecay_ci_readiness_phases` schema.
- #963 retains `d129af38a`, `76b9c70fb`, and `45a978163` for interruption attribution, text-current waits, and the coalesced BusyFollowUp lost-wake fix.
- Latest #707 is merged through `850b7a8c0`; surviving head is `10ed30e137a1307a7b5801981fabf5f0775ec30a`.

Final evidence is posted on #963. #917 stays open because resident-memory refusal preserves text but does not satisfy strict graph readiness. Neither child is being merged here; parent integration will sequence #963 with #800 work.

Issue #838, opened 2026-09-04
perf(semantic): native evaluation takes ~10 minutes for a one-file project


ScriptedAlchemy:
## Evidence

Measured on the `#753` in-process journey (`public_semantic_activation_rollback_and_exact_retry_preserve_graph_authority`, `crates/tracedecay/src/daemon/production_harness/semantic_activation_journey_test.rs`) on a 96-core host with the `JinaEmbeddingsV2BaseCode` distribution fixture (613 MB):

| phase | wall clock |
|---|---|
| project open + first code generation published (one file, one symbol) | ~1 s |
| `DaemonInvocationRequest::semantic_activate` → `execute_semantic_evaluation` (native evidence at `current` and `10x` scales) → activation committed | ~9 min 47 s (07:36:57 → 07:46:44) |
| readiness convergence wait | 60 s (bounded by the test) |
| harness teardown after the panic while the activation reconciler kept retrying | ~10 min |

The CLI journey (`crates/tracedecay-cli/tests/core_cli_suite/semantic_activation_test.rs`) took 401 s end to end on the same host.

The corpus is a single `pub fn` with one symbol. Ten minutes of evaluation cannot be embedding cost; it is fixed overhead (model verification/load per scale or per profile, calibration workload cases, the `10x` replicated corpus projection, resource ceilings under `SemanticResourceCeilings::default()`), and it makes every activation iteration and every journey run a 20-25 minute loop.

## Ask

- Attribute the evaluation wall clock per stage (model load, per-scale projection, per-profile measurement, calibration, publication) with hotpath labels or `tracing` spans, on the one-file fixture.
- Bound the fixed overhead so a trivial corpus evaluates in seconds, keeping the genuine-evidence contract (`REQUIRED_RESOURCE_SCALES`, exact `10x` chunk count) intact.
- Make the harness teardown cancel the activation reconciler promptly instead of letting it retry for the full shutdown deadline.

Refs #753.

ScriptedAlchemy:
[...]

PR #963, opened 2026-09-06
fix(code-index): preserve text freshness and refuse graph pressure (#917)

ScriptedAlchemy:
## Summary

- Preserve text freshness when an out-of-band edit arrives while a reconcile still owns its pass: one coalesced `BusyFollowUp` remains pending under concurrent reads, later edits are not lost, and shutdown/deadline interruptions retain their typed ownership without retry storms.
- Refuse optional native-graph publication at the daemon's measured resident-memory watermark. A pressure trip is permanent for that publication attempt; a later request cancellation keeps cancellation identity precedence; a recovered cell admits a new attempt.
- Keep text and graph truth independent. Resident-memory refusal preserves exact/lexical text serving and records a typed graph refusal, but it does **not** satisfy strict graph readiness.
- Consolidate #962 and #963 dogfood evidence into one authority: `TRACEDECAY_DAEMON_PID`, the portable `resident-memory` reader, and one `tracedecay_ci_readiness_phases` schema. The strict validator and 600-second bound are unchanged.

This PR supersedes #962 and contains its unique memory-refusal work. It targets #707 (`codex/tracedecay-total-redesign-plan-reopened`) and includes base `850b7a8c089ade73573d58fa4711876eb2ef7199`. Keep #917 open: pressure containment prevents OOM loss of text service, but the remaining graph/text scale work still has to reach the unchanged strict contract.

Do not merge yet; the parent integration lane will sequence this with #800 work.

## Review fixes

- Monotonic per-attempt pressure cancellation, with explicit request cancellation kept distinct from `GraphBudgetKind::ResidentMemory`.
- Real publication enters manifest projection before pressure trips; no journal/head is exposed, and a new recovered attempt publishes normally.
- 32 simultaneous freshness reads retain one follow-up; a source edit during the next owner pass is served by the successor.
- Text-only waits use the text-current seat; graph/pinned-cursor paths retain complete-seat assertions.
- Reconcile timing is named `wake_to_gates_held_micros`, not SQL-writer wait.
- Latest #707 integration fixes the former #962 hotpath slice compile failure and preserves the generation-free capability digest contract.

## Final evidence (`10ed30e137a1307a7b5801981fabf5f0775ec30a`)

- `cargo test -p tracedecay-code-index-runtime --lib code_index_scheduler:: -- --test-threads=1` — 318 passed, 0 failed, 1 ignored.
- `cargo test -p tracedecay-store-runtime --lib sealed_publication_tests -- --test-threads=1` — 7 passed, 0 failed, 1 ignored.
- `cargo test -p tracedecay-rusqlite-runtime -p tracedecay-global-db --locked --features tracedecay-rusqlite-runtime/hotpath,tracedecay-global-db/hotpath,tracedecay-global-db/test-helpers` — passed (the previously failing slice).
- `cargo clippy -p tracedecay-graph-db -p tracedecay-code-index-runtime -p tracedecay-store-runtime --all-targets -- -D warnings` — 0 errors, 0 warnings.
- `python3 scripts/test-pr-dogfood-portability.py` — 15 passed; `python3 scripts/test-check-pr-dogfood-output.py` — 14 passed.
- Bounded publication scale: 2 scopes × 96 files, one overlapping projection, 147.8 MiB retained growth, 168.2 MiB sampled peak growth, 5.406 s wall; all configured RSS ceilings passed.
- Commitlint and scoped rustfmt checks passed.

### DIFF
--- a/crates/tracedecay-code-index-runtime/src/code_index_scheduler.rs
+++ b/crates/tracedecay-code-index-runtime/src/code_index_scheduler.rs
@@ -5260,6 +5260,23 @@ impl CodeIndexSchedulerErrorV1 {
         }
     }
 
+    /// The typed interruption a reconcile pass stopped on, when it did.
+    ///
+    /// An interrupted pass is not a failed pass: the epoch that cancelled it
+    /// was advanced by an observed source change (or shutdown) whose caller
+    /// also posted the wake that re-runs the pass, so the worker attributes
+    /// it as superseded instead of reporting the served generation stale.
+    pub fn reconcile_interruption(
+        &self,
+    ) -> Option<crate::code_index::production::CodeIndexInterruptionV1> {
+        match self {
+            Self::Production(CodeIndexProductionErrorV1::Interrupted(interruption)) => {
+                Some(*interruption)
+            }
+            _ => None,
+        }
+    }
+
     pub fn is_graph_activation_refusal(&self) -> bool {
         matches!(self, Self::GraphActivationRefused(_))
             || matches!(
--- a/crates/tracedecay-code-index-runtime/src/code_index_scheduler.rs
+++ b/crates/tracedecay-code-index-runtime/src/code_index_scheduler.rs
@@ -5387,7 +5387,7 @@ impl CodeIndexSchedulerErrorV1 {
             || matches!(
                 self,
                 Self::GraphProjection(CodeGraphProjectionError::BudgetExhausted { budget, .. })
-                    if budget == "resident_memory"
+                    if budget == tracedecay_graph_db::GraphBudgetKind::ResidentMemory.as_str()
             )
     }
 
--- a/crates/tracedecay-code-index-runtime/src/code_index_scheduler/graph_activation.rs
+++ b/crates/tracedecay-code-index-runtime/src/code_index_scheduler/graph_activation.rs
@@ -248,6 +248,22 @@ fn take_injected_activation_gate(
         .remove(worktree_id.as_str())
 }
 
+/// The typed reason a generation reports while its native graph is refused by
+/// the measured-RSS admission watermark. Shared by the persistent publication
+/// path and the injected test refusal so status and tests name one verdict.
+pub(crate) const RESIDENT_MEMORY_GRAPH_REFUSAL_REASON: &str =
+    "code graph activation was refused by the resident-memory policy";
+
+/// Whether a projection error is the resident-memory budget refusal that
+/// [`CodeIndexSchedulerErrorV1::is_graph_activation_refusal`] recognizes.
+pub(crate) fn is_resident_memory_refusal(error: &CodeGraphProjectionError) -> bool {
+    matches!(
+        error,
+        CodeGraphProjectionError::BudgetExhausted { budget, .. }
+            if budget == tracedecay_graph_db::GraphBudgetKind::ResidentMemory.as_str()
+    )
+}
+
 #[derive(Clone)]
 pub enum CodeGraphActivationAuthorityV1 {
     Persistent {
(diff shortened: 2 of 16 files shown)


ScriptedAlchemy (commented):
## Architecture review — direction accepted; HOLD for integration evidence

Reviewed `45a97816318cdab4913259ff20c84914bc6359aa`; source review only. Exact-head CI is pending. The out-of-band edit arriving during owner work is a real lost-wake case, and using the existing coalesced BusyFollowUp is preferable to polling or another scheduler.

- **P1 — preserve wake ownership under repeated reads.** Extend the new regression with many simultaneous freshness queries while a pass is held: at most one pending follow-up, no cancellation/restart storm, and a source edit arriving during the next pass still eventually serves. Cover shutdown/deadline separately so reclassifying Interrupted does not silently drop the last necessary wake or retry after terminal cancellation.
- **P1 — keep test intent exact.** Text-current waits are appropriate for lexical/text cases. Preserve complete graph-seat assertions for graph traversal, semantic scheduling and the initial pinned-cursor generation where the test actually needs them. A newer text ID alone is not evidence that an immutable old cursor or optional graph remains correct.
- **P1 — merge the instrumentation with #962 once.** Both PRs modify `ci-pr-dogfood-smoke.sh`, the harness PID export and phase reports. Keep one schema and one portable RSS implementation; retain this PR's lost-wake/interruption fixes and #962's memory-refusal behavior independently. Do not leave two competing timelines/PID variables.
- **P2 — name timings by what they measure.** The described `writer_wait_micros` spans wake-to-all-gates-held, not just the SQL writer. Keep queue/admission/lock execution separate or rename the aggregate so #838/#852 attribution does not optimize the wrong owner.

The serial local suite and 60 focused repetitions are useful, not replacement evidence for Windows nextest and the merge tree. Keep #917 open until real dogfood reaches its unchanged strict contract; phase attribution by itself does not repair the measured throughput. Any later switch to a production-like build profile must be explicit benchmark configuration, not concealed as a correctness fix. Target #707 only; keep #707 draft.

ScriptedAlchemy (commented):
## System-level re-review at `45a978163` — preserve the wake fix and distinct serving states

I re-read the registry diff and the ignored-dependency interruption mapping. No new confirmed correctness defect found in those changes.

The follow-up wake is necessary for the actual ownership interval, not just for a busy mutex: the worker can have released the scheduler mutex but still own text projection or graph seating for source state A. A query observing a later out-of-band change B cannot assume that A's pass will revisit the source. Posting one coalesced BusyFollowUp in that branch repairs the missing obligation without invalidating the in-flight generation. Keep the pending-arrival check and do not replace this with repeated polling or an epoch advance on every reader.

The shared-reader interruption mapping is also appropriate: ordinary source capture should report a production interruption, while the explicit ignored-dependency admission boundary translates it back to its own typed refusal. Preserve cancellation/deadline distinctions and rollback of the temporary admission roster. Merely renaming the log would not fix the retry-classification issue; this change does more than that.

Acceptance should cover A published/text-projecting -> B committed -> many concurrent freshness readers -> exactly one pending follow-up -> B text-current, with the graph lane delayed or refused. The relevant text-only tests should wait on text identity; tests for graph traversal/complete serving must retain their graph assertions. Also test shutdown and source supersession separately from a deadline: a deadline is not itself evidence of a newer source observation or a queued wake, even though all can be typed interruptions. Do not call a deadline-bound pass successful progress based solely on its INFO log.

Merge the dogfood instrumentation with #962 as one implementation: one exported daemon PID, consistent units, one phase timeline, and explicit unavailable memory samples. Preserve both independent production fixes. A clearly attributed timeout remains a timeout; this PR neither proves strict graph readiness nor completes semantic activation.

Exact-head CI and profiling are pending; coverage is queued. The source change is reasonable, with native/integrated acceptance outstanding rather than a newly invented merge-blocking defect. No tests executed in this review.

ScriptedAlchemy:
Consolidation evidence at `10ed30e137a1307a7b5801981fabf5f0775ec30a` (base #707 `850b7a8c089ade73573d58fa4711876eb2ef7199`, zero behind locally):

- scheduler module: 318 passed, 0 failed, 1 ignored;
- store-runtime sealed publication: 7 passed, 0 failed, 1 ignored;
- former #962 hotpath slice: passed;
- strict clippy across graph-db/code-index-runtime/store-runtime: 0 errors, 0 warnings;
- dogfood portability/output: 15 + 14 passed;
- bounded 2-scope/96-file publication: one overlapping projection, retained growth 154,931,200 bytes, sampled peak growth 176,361,472 bytes, wall 5,406 ms.

Contract mapping: #963 remains authoritative for BusyFollowUp/interruption/text-current behavior; #962's resident-memory refusal is carried by `4a93cc22c` and hardened by `890883e7f`; dogfood instrumentation converges in `499d6d67e` to one PID/RSS/phase-report authority. Text remains available on refusal, while strict graph readiness remains unsatisfied.

Issue #909, opened 2026-09-06
Decode-free semantic readiness: seal source-side commitments on CodeGenerationManifestV1


ScriptedAlchemy:
## Context

#753 readiness now compares `vectors.source_manifest_digest()` with the serving generation's `ProjectionBatchRequestV1.changes.manifest_digest` or the full-replay recomputed digest (`vectors_serve_source`, `crates/tracedecay-usecases/src/semantic_runtime/production.rs`). Both sides are reachable today only by materialising (decoding) the serving generation via `serving_code_scope`, the decoded consumer the restart lane wants gone.

## Shape (architecture review, 2026-09-06)

- The source-side commitment belongs on the sealed code-generation contract: compute the required digests once during sealing, authenticate them with the generation, record them on `CodeGenerationManifestV1`, and expose them read-only from the seated text level (`LatestCodeTextGenerationV1` / `VerifiedSealedTextGenerationMetadataV1`). No model/profile lifecycle state on the manifest.
- Keep the vector artifact's accepted-source record: sealed metadata answers "what source does this generation represent", vector provenance answers "what source produced these vectors". Readiness compares two independently authenticated records; never copy a digest from the candidate vector record into serving metadata (that recreates the self-comparison that caused the #753 stall).
- `ChangedCodeChunkSetV1::compute_digest()` includes `from_generation`, `to_generation` and the added/deleted/reused partitions, so an incremental digest and a full-replay digest are not interchangeable. Preserve both supported representations explicitly.
- Historical manifests without the fields need explicit verified migration or typed unavailability, not a fabricated default (revision <6 sealed generations already refuse and rebuild).

## Acceptance

- Semantic readiness/restore decisions read only seated metadata plus the vector provenance record; no decoded generation on that path.
- Same source identity → activation remains valid across generation id churn; different source identity → typed refusal.
- Semantic activation journey (activate → ready → strict query → restart → query → rollback → retry) green.

Targets `codex/tracedecay-total-redesign-plan-reopened` (#707). Refs #753, #908.

ScriptedAlchemy:
[...]

PR #964, opened 2026-09-06
perf(semantic): retain projection batches across qualifications

ScriptedAlchemy:
## Summary

- 23dbbe7ed: journey stage ledger attributes exclusively (per-run reset instead of a process-global ARMED flag, no folding of shared report samples, VmHWM labelled process-lifetime). Reviewer's three ledger corrections on 410b4dbcf.
- 742eb1660: daemon-lifetime cache of immutable projection batches keyed by the full admitted projection identity and composed embedding inputs, privacy-partitioned, byte-bounded, with deduplicated concurrent construction. Request cancellation, leases and publication bindings are still per request; calibration and acceptance stay fresh. Targets the 456 s per-qualification clean projection measured on #838.

## Status

UNVERIFIED. The lane was stopped on owner request before its gates (fmt, clippy, unit tests, cold vs warm journey run with VmHWM) completed; clippy was still queued behind the journey run. Per the reviewer's ordering this lands only after one uncached passing journey and the #913 cache gates (cold/cached equivalence, miss/refuse on changed inputs, no cross-request poisoning, eviction/shutdown bound). Do not merge without those.

Refs #838, #852. Does not undraft or merge #707.


ScriptedAlchemy (commented):
## Architecture review — HOLD

Reviewed `742eb1660dc4352a2374ebc317cf072059c613ce`; source review only. CI is queued, and the PR correctly describes itself as unverified. Keep this behind the uncached #753 journey and #913 cache acceptance.

- **P1 — shutdown is not a cache-retirement fence.** In `crates/tracedecay-semantic/src/semantic_evaluation.rs`, `SemanticEvaluationProjectionBatchStoreV1::release()` clears `entries`/`retained_bytes` but leaves in-flight claims able to call `install()` afterward. The worker owner explicitly calls release even when workers outlive its shutdown deadline. A late builder can therefore retain bytes again after shutdown reported their release. Fence new claims/installations with store retirement or a generation token and settle existing claims; test `claim -> release -> late install` and waiter cancellation. Do not discard the existing request-owned cancellation/settlement rules.
- **P2 — `last_used_epoch` is not a multi-request pin.** Request A touches K, request B touches K, then B drops: K's last epoch is B, so the eviction filter considers it unpinned even while A is active. Either track actual per-entry borrowers or simplify/document eviction as an allowed cache miss; do not claim that every active request's touched batches remain pinned. Add this overlapping-request case.
- **P1 acceptance — the 80 MiB counter bounds completed entries, not total cache-associated memory.** In-flight keys, composed input copies and cloned hit vectors also consume memory. Bound concurrent construction through the real owner, account for those peaks, and measure cold/warm RSS with multiple projects. `release()` alone is not evidence of reclaimed resident bytes.

The cache key's full admitted projection plus exact composed inputs is the right direction. Keep calibration, measurement receipts, current privacy admission, leases and activation CAS per request. Require cold/cached output equivalence; changed projection/privacy/input misses; cancelled-builder recovery without poisoning waiters; eviction and shutdown races; fmt/clippy and the real journey before integration into #707. No approval to undraft or merge #707.

ScriptedAlchemy (commented):
## System-level re-review at `742eb1660` — do not hoist before the lifetime and measurement contracts agree

I traced the cache into `DaemonSemanticEvaluationWorkerOwnerV1::shutdown`, not just the cache unit tests. The exact projection/input/privacy key and key-ordered build claims are useful; acceptance/calibration/publication bindings remain request-owned. Keep that separation.

### P1 — shutdown release is only an eviction, not a retirement fence

The owner calls `projection_batch_store.release()` even when workers outlive its shutdown deadline. `release()` clears entries but leaves in-flight claims and admission enabled; a worker holding a build guard can subsequently call `install` and repopulate the store. The cache's own test explicitly demonstrates post-release reuse, so this is not a bug in a clear-and-reuse API by itself. The bug is using that API to promise that shutdown returns the retained bytes while workers may still install them. Fence store admission/installation at the owner's terminal shutdown, or defer the release assertion until all admitted workers settle. A normal pressure-driven cache clear can remain a separate nonterminal operation. Test a builder parked before install, shutdown with that worker pending, then release the builder and verify no terminal cache resurrection.

### P2 — last-used epoch cannot pin all active borrowers

A touches K; B touches K, replacing K's `last_used_epoch`; B drops; C can now evict K even though A is still active. The `active_fills.contains(last_used_epoch)` predicate therefore does not implement the documented request pin. Since hits clone owned vectors, this interleaving does not establish a dangling reference or wrong result; its proven consequence is that A may re-embed the batch and the no-thrashing claim is false. Either track the required per-entry active-request membership or simplify to an explicitly evictable cache and measure the tradeoff. Add the A/B/C interleaving; the existing same-request and sequential-request tests do not cover it.

### P2 — report cache accounting separately from total live memory

`retained_bytes` accounts completed entries, not cloned hit buffers, the complete keys retained in `in_flight` and build guards, per-call distinct/group maps, or construction output. The counter being <= max is not proof that all concurrent evaluation memory is <= that bound. `cache_entry_bytes` also estimates identity via JSON size rather than actual owned capacities/map storage. Keep the completed-entry budget, label it accurately, bound concurrent construction independently and measure daemon peak RSS for cold/warm/overlapping requests. Do not multiply the large qualification corpus by active requests while crediting only entries.

Require an uncached passing operator journey before evaluating the speedup, then cold/cached output and decision equivalence, input/model/tokenizer/runtime/privacy invalidation, cancellation isolation and terminal shutdown. No need to persist evaluation verdicts or current authority in this cache. All returned workflows are queued; no tests or performance gates were run in this review. The PR's UNVERIFIED status remains appropriate.

ScriptedAlchemy:
## HOLD resolution and acceptance evidence

Pushed exact head `c0587f1717b3208cd56c40fd12dfb903858d0e23` on base `a8c322a02ce7f5453dab0a3d89393b9b729f6b4d`.

### Ownership / retirement

- Store retirement is irreversible under the cache state lock. It clears completed entries, wakes exact-batch waiters, rejects every later claim, and prevents an outliving build guard from installing after shutdown.
- Entries now retain the full set of active request borrowers. Dropping request B cannot erase request A's pin; eviction considers an entry only when its borrower set is empty.
- Misses retain one batched runtime dispatch (the first attempt split N misses into N dispatches and timed out the live journey; the regression now fixes that directly).
- Only exact immutable computation is shared: admitted projection + execution width + tensor shape + exact composed documents. Source/profile/calibration/privacy acceptance, cancellation, lifecycle verification, publication leases, and activation CAS remain live per request.
- `production.rs` is byte-identical to the current #707 base, retaining #909's sealed-source/decode-free authority and the active #753 lifecycle behavior.

RED evidence: `cc-8228` late install repopulated after release; `cc-8229` B-drop erased A's pin; `cc-8230` post-release claims rebuilt; `cc-8566` one admitted cold batch became 3 runtime dispatches.

### Memory truth

The store reports completed batches, in-flight key copies, active composed lookup keys, warm-hit vector clones, current total, and peak separately. Shutdown logs all categories; outliving claims remain accounted until settlement. The 80 MiB limit is now explicitly the completed-batch retention limit, not an RSS claim.

Real Jina FastEmbed, two independent project stores, exact cold/warm vector and receipt bytes (`cc-8718`, 1/1):

- RSS: before 12,533,760; cold 731,213,824; warm 1,329,025,024; two stores 1,237,311,488; after retirement 993,116,160 bytes.
- First store: cold 30,024 accounted bytes; warm 48,648 (30,024 retained + 18,624 active hit vectors). Second store: 30,024 retained.
- Both retired stores report 0 current accounted bytes while RSS remains 993,116,160, explicitly proving that retirement is not presented as allocator/RSS reclamation.

### Verification

- `cc-8716`: `tracedecay-semantic` **209/209**.
- `cc-8717`: semantic runtime use cases **113/113**, including source/profile/privacy/calibration drift refusal.
- `cc-8416`: semantic worker ownership/lifecycle **14/14** (shared cross-project admission included).
- `cc-8417`: daemon semantic invocation **5/5**.
- `cc-8652`: full FastEmbed activation journey **1/1**, 1231.19 s; restart, strict query, source rollback, failed-transition refusal with unchanged authority, lifecycle retry, exact recovery, and clean teardown all passed. Process-lifetime VmHWM: 9,954,440 KiB.
- `cc-8713`: strict clippy across semantic/usecases/code-index-runtime/daemon-service/composition, all targets + test helpers: **0 errors, 0 warnings**.
- Scoped rustfmt check and `git diff --check` passed.

Local ancestry is **0 behind** #707; GitHub reports `mergeable: MERGEABLE`. Hosted checks are newly queued, so `mergeStateStatus` is currently `UNSTABLE` only for pending checks. PR remains draft for parent-controlled integration.

PR #1196, opened 2026-09-10
test(query): assert the typed pruning partial for common-term candidates

ScriptedAlchemy:
Fixes a deterministic failure on the #707 tip: `tracedecay-query` `search_quality_suite` `scaling::common_term_candidates_are_bounded_by_the_rarest_source` panicked at `candidate_producers.rs:5460` calling `.complete()` on a result that has been an intentional typed `RetrieverOutcome::Partial` since `d3fe1df45` (`admit_candidate_sources` / `candidate_admission_outcome` in `retrieval/lexical.rs`): common-term candidates are bounded by the rarest source and the pruning is *reported*, not silently truncated.

The test now matches the `Partial` and asserts the exact reason — `RetrievalFailure::CandidateSourcesPruned { term_sources: [("shared_flag", 16385)], document_frequency_budget: 16384 }` (`MAX_LEXICAL_CANDIDATE_DOCUMENTS_V1 + 1` / `MAX_LEXICAL_CANDIDATE_DOCUMENTS_V1`) — while keeping the original `candidates.len() == 1` and `coverage.eligible == 1` assertions. No production change.

Verification: exact test 1/1; `scaling::` module 1 passed + 1 ignored (pre-existing cold/warm benchmark); clippy `-D warnings --tests`; fmt; commitlint.

Not included, deliberately: the other deterministic tip failure, `tracedecay-application` `accepted_profile_authority::tests::portable_report_requires_the_current_workload_digest`, cannot be fixed by regenerating the embedded native-qualification asset — the real generator (`tracedecay-search-eval qualify-native --profile hybrid-conservative`) ran to completion and the candidate **failed the production quality bar** (validation `natural_language` `ndcg_at_10_ppm` 734639 vs baseline 734639, required gain ≥ 1). Filed separately as a product finding against `718376c49`.


ScriptedAlchemy:
Superseded: the same fix (assert the typed `CandidateSourcesPruned` partial for the mixed-selectivity query, common-only query still complete) landed on the integration branch as 785cbdf78 with the full search_quality_suite green (75 passed).

PR #1319, opened 2026-09-15
fix(test): restore prebuilt CLI fixture lookup

ScriptedAlchemy:
## Why

PR #1317 made absent CLI failures unmistakable, but it also removed the profile-sibling lookup used by hosted test jobs. Those jobs build `target/<profile>/tracedecay` without setting `TRACEDECAY_TEST_BIN`, so root journeys stopped before startup.

Restore the existing cross-package lookup and retain the `CLI binary not built:` prefix when the selected path is absent.

## Scope

- Restore the profile-sibling fallback in `tracedecay_bin()`.
- Apply rustfmt to the missing-CLI regression test.
- Do not build the CLI from the test process or add a dependency.

## Blast Radius

All integration suites that use `tests/common/mod.rs` regain their prebuilt CLI path. Explicit `TRACEDECAY_TEST_BIN` selection and version validation remain unchanged.

## Verification

- Hauler cc-21692 built the profile-sibling CLI with no warnings.
- Hauler cc-21698 passed a root CLI journey with `TRACEDECAY_TEST_BIN` unset at head `7ae2044427d9`.
- Hauler cc-21697 passed the forced missing-binary classification test at the same head.
- `cargo fmt --check -p tracedecay` passed at the same head.
- Hosted Linux root journeys advanced past CLI resolution. Their remaining failures occur in unrelated workflow, evaluator, and runtime assertions.


ScriptedAlchemy:
Superseded by format commit a26fbb7bd4 on #1320.

ScriptedAlchemy:
Reopened: closed by mistake (confused with the Format-only tip fix that landed as a26fbb7 on #1320). This PR restores prebuilt CLI fixture lookup and remains needed.

PR #1321, opened 2026-09-15
fix(workflow): make fan-out rebound projection JSON-safe

### LINKED ISSUE #1283: Work fan-out recovery replays version-pinned child proposals and loops on ProposalMismatch; journey test was vacuously skipped behind the FastEmbed fixture
## Summary

`daemon_suite::advanced_workflow_journey_test::mounted_fan_out_recovers_then_synthesizes_and_hands_off` has never actually run in any gate. Its body was behind `task_session::install_semantic_fixture(&home)`, which returned `None` and skipped the test unless `TRACEDECAY_DISTRIBUTION_FASTEMBED_FIXTURE` was set, and no workflow or `scripts/check-distribution-acceptance.sh` leg ever set it for this test (only for `semantic_activation_journey_test` and the CLI activation test). The vacuous "ok" hid a real Work fan-out recovery bug.

The dense-retrieval retirement (#1260 / #1261) deletes that fixture and its guard, so the body now runs and fails deterministically at:

```
advanced_workflow_journey_test.rs:208: timed out waiting for post-recovery cancellation child
```

## Root cause

Daemon log during the stall (every 5 s):

```
WARN tracedecay_daemon_service::invocation::work::workflow_fan_out::recovery: bounded workflow fan-out recovery page failed problem=Unavailable
```

Temporarily naming the erased errors along that path:

- `workflow_fan_out.rs::apply_workflow_child_product_mutation` → `product.mutations().mutate(...)` returns `WorkProductApplicationErrorV1::InvalidRequest`
- inside `mutation.rs::commit_change`, `snapshot.graph().clone().apply(change)` fails with `WorkProductContractError::ProposalMismatch`
- the change is `WorkProductChangeDraftV1::DecideProposal { proposal: child.proposal.clone(), disposition: Accepted }` issued by `reconcile_workflow_fan_out` for a child whose item has no accepted proposal yet
- `WorkProductGraphV1::validate_proposal` rejects it because `proposal.based_on_version != self.version`

The fan-out plan stores each child's `WorkProposalV1` with the graph version at plan creation. In this journey the first child succeeds (`LinkAcceptedAttempt` / accept advance the graph version), the daemon is killed mid fan-out, and after restart recovery tries to decide the remaining children's stored proposals against the now-newer graph. Every recovery page fails, the loop retries forever, and the "cancel" child is never dispatched.

So: fan-out child proposals are version-pinned at plan time, but recovery replays them after sibling mutations have moved the version. Any fan-out where one child completes before a crash cannot recover.

## Repro

On a tree with the dense-retrieval retirement applied (branch `agent/dense-retire-1261`, or any tree where the `install_semantic_fixture` guard is removed):

```sh
cargo build -p tracedecay-cli --bin tracedecay
TRACEDECAY_TEST_DAEMON_LOG=/tmp/fanout.log \
cargo test -p tracedecay --features test-helpers,test-transport --test daemon_suite -- --exact \
  advanced_workflow_journey_test::mounted_fan_out_recovers_then_synthesizes_and_hands_off
rg 'fan-out recovery page failed' /tmp/fanout.log
```

Reproduces in isolation (not load-dependent); the sibling lifecycle test's failure is a separate, pre-existing issue.

## What this is not

- Not caused by the retirement itself: `crates/tracedecay-domain/src/work_product`, `crates/tracedecay-contracts/src/work_product`, `crates/tracedecay-daemon-service/src/invocation/work`, and `crates/tracedecay-application/src/work` are byte-identical to `dca4de9de1` on the retirement branch.
- Not a test-only problem: the recovery worker in production loops on the same `Unavailable` and the workflow run stays `Running` with an undispatched child.

## Options

1. Recovery rebases the stored child proposal onto the current graph version (`based_on_version = graph.version()`) before `DecideProposal`, keeping the proposal id and command id so idempotent replay still works.
2. Fan-out planning records per-child decisions incrementally at dispatch time instead of storing undecided proposals in the plan, so there is nothing stale to replay.
3. Recovery treats `ProposalMismatch` as a typed terminal state for that child (RecoveryRequired with reason) rather than an infinite `Unavailable` retry.

Whichever is chosen, `DaemonInvocationProblem::Unavailable` collapsing ~40 distinct errors in `workflow_fan_out.rs` made this take instrumentation to see; the recovery warn should carry the underlying error.

The retirement branch leaves the test live and red rather than re-guarding it behind a fixture that no longer exists.

ScriptedAlchemy:
## Summary
- Root site: `WorkflowRunProjection.active_fan_out_attempts` used `BTreeMap<WorkAttemptIdentityV1, …>` composite keys that cannot serialize as JSON object keys after `FanOutChildRetryRebound`.
- The `serde_json::to_value` failure (`key must be a string`) mapped through the response transport to `workflow.transport_unavailable`; cc-21675 reproduced this exact path after the initially empty map became populated.
- Fix: remove the composite-key map from the projection wire and resolve the active fan-out attempt from durable journal history.
- Keep journal reads on deferred transactions (`0fcf6505e9`) and explicitly wait for the Workflow mount after daemon restart. Generated dashboard contracts are synchronized with the projection wire.

## Evidence
- Before: cc-21640 timed out after ~20s waiting for durable `WorkflowGetRun`; measurement observed 351/351 polls as `workflow.transport_unavailable` while daemon recovery settled in milliseconds and `begin_immediate` holds peaked at 115ms.
- Root reproduction: cc-21675 produced JSON `key must be a string` immediately after `FanOutChildRetryRebound`.
- Contract proof: cc-21680 passed the rebound-projection JSON encode/decode assertion.
- Recovery proof: cc-21690 passed durable GetRun, recovery, pause, resume, and cancel after restart. It failed only later at tip code-generation restore (`task_session.rs:289`, 180s).
- The later timeout is classified (A)/(C): separate tip/code-index restore risk, or a journey stage previously masked by the GetRun failure, under a kache store at 410+ GB / 400 GB. This PR does not expand into V16 restoration work.

## Change risk
- Ripwire exact-head PR context at `36959faef1` found 131 transitive dependents / 39 files for the projection module and 11 indexed tests; the wire regression has direct contract coverage plus the production daemon journey.
- Generated schema and TypeScript contracts were regenerated from Rust authority.
- Quality delta reported only test-code duplication for the mount wait helper and growth in an existing contract test; no production complexity or API-surface regression. The repeated mount helpers remain explicit typed operation probes in the shared journey fixture.

## Test plan
- [x] cc-21680: rebound projection remains JSON-transportable
- [x] cc-21690: restart recovery passed durable GetRun / recovery / pause / resume / cancel
- [x] cc-21695: dashboard contract schema export completed
- [x] Ripwire PR context and committed-range quality delta at `36959faef1`

Closes #1283


ScriptedAlchemy:
CI note: prior SDK/hotpath jobs show cancelled (superseded), not product failures. Re-queued main CI. Workflow GetRun transport root is JSON-unsafe rebound map keys (cc-21675); cc-21690 proved past durable GetRun.

Issue #1280, opened 2026-09-13
perf(clones): qualify the 5k-file and 100k-symbol operating envelope


ScriptedAlchemy:
Decision: `docs/plans/tracedecay-v2/00-plan-set-index.md`, rejected decision 11, and `docs/V2-OPERATING-MODEL.md` (Grafeo graph-only). Campaign root: #1260.
Target branch: `codex/tracedecay-total-redesign-plan-reopened` (#707). Depends on: #1268–#1272.

## Owns
On a pinned ~5,000-file corpus record: eligible bytes, source tokens, body count, symbol count, language counts, changed-body count, cold/warm cache state, hardware and OS, elapsed time, peak RSS, disk bytes, postings read, candidates admitted, pairs verified.

Targets: clone indexing adds ≤ 1 GiB peak RSS; clone work adds ≤ 60 s on the declared corpus; one-symbol indexed lookup p95 ≤ 250 ms; one ordinary changed function refreshes within 1 s after extraction; unchanged functions are not renormalized; adversarial workloads stop, page, resume, or return partial coverage rather than growing without bound. Cancellation and restart are exercised.

The graph-publication RSS problem (#852) stays measured independently.

## Acceptance
A simple Linux developer benchmark per the roadmap acceptance rule — raw samples and practical deltas, no gate scaffold.

## Rulings applied (GPT-6 Pro, 2026-09-14)
- Start measurements with the **first usable slice** (exact lookup on full-rebuild generations) and keep them running through #1270–#1273; closure still requires full acceptance.
- **Validate the starting bounded policy** in #1271/#1272 (k/w, 30-token minimum, 0.70 ratio, 16,384 posting rows, 256 candidates, 64 verifications, 2M steps, 200 ms, absolute 1,024 hot-posting threshold) before claiming the performance envelope; those numbers are starting decisions, not measured safe maxima.
- **Corpus (cost and exhaustion):** a thousand-copy exact family without quadratic pair creation; long repetitive bodies; hot-only fingerprints; many plausible candidates; cancellation; every budget boundary with truthful partial results. **Real-world usefulness:** independently reviewed real-repository clone pairs and hard negatives, reported by supported language and declared overlap band.

ScriptedAlchemy:
PR #1327 merged at `8d16fc6e0c` (merge commit) while the drain receipt gate was still red.

[...]

PR #1327, opened 2026-09-15
perf(clones): qualify 100k-symbol operating envelope

ScriptedAlchemy:
## Why

[Issue #1280](https://github.com/ScriptedAlchemy/tracedecay/issues/1280) needs measured clone cost and exhaustion behavior on a pinned developer corpus. This change extends the existing production index benchmark and records the raw Linux samples.

The 101,154-symbol run passes the added-RSS and exact-lookup targets. It fails the 60-second clone-work target and the one-second body-refresh target. The receipt reports those failures without raising a budget.

## Scope

- `tracedecay-index-bench` records file, symbol, body, token, language, disk, RSS, refresh, exact-read, fingerprint, cancellation, and restart facts.
- V14, V15, and V16 artifact selection provides a direct clone-index control and treatment.
- The committed receipt includes the 5,200-file sample, the 101,154-symbol sample, and the 1,000-member exact family.
- MCP redundancy mounting and family-read transport remain out of scope.

## Blast Radius

The change affects only the daemon-free benchmark and its evidence. Production query policy and serving code do not change.

## Verification

- Hauler `cc-21833` and `cc-21834` measured V14 to V16 artifact ingest at 73.238 s to 142.080 s. The 68.842 s delta fails the 60 s target.
- Peak RSS changed from 21,417,521,152 B to 21,641,617,408 B. The 224,096,256 B delta passes the 1 GiB target.
- Warm exact lookup p95 measured 510 µs. One-body refresh measured 8.615 s.
- Hauler `cc-21824` returned 999 peers for a 1,000-member exact family without continuation.
- Hauler `cc-21820` through `cc-21823` passed candidate, posting, work, hot-posting, paging, resume, and cancellation behavior.
- Hauler `cc-21858` completed the final focused compile check with zero errors.


ScriptedAlchemy:
## Drain blocker (exact-head receipt)

Head `573fd74ebc` still ships `benchmark_data/index-bench/clone-envelope-20260915.json` with `harness_commit=e725037b10` (not this head).

Measured bounds **fail** the required 60s clone-work + 1s body-refresh gate:
- `file_5200_clone_v16`: ingest 130.9s, body_refresh **7.77s**
- `symbol_101154_clone_v16`: ingest 142.1s, body_refresh **8.62s**

Hold merge until a receipt on **this** head passes both bounds. Do not treat the staged fingerprint commits as evidence of the bound.

PR #1339, opened 2026-09-15
fix(code-index): seat before the clone backfill; restore clone bodies canonically (#1103)

ScriptedAlchemy:
## Summary

- `indexing_lifecycle_test::mounted_incremental_lifecycle_preserves_only_complete_compatible_generations` **passes at the untouched 45 s bound** (2/2 runs on a 4-vCPU rig, perf profile, hotpath-instrumented CLI). Restart → `code_index_serving_generation_seated`: **89.8 s → 39.0 s / 38.7 s**.
- Dominant wait (measured): the graph seat waited for the text projection task to finish the clone-fingerprint successor (a second full decode of the sealed source into a second 785 MB artifact, ~27 s serial) after exact/lexical owners were already installed. The published pass now stops at owners-ready; the backfill is demand-driven on the retained path, runs without the owner-pass guard (so status stays `fresh`), and the owner probe no longer queues behind a backfill slice.
- **B1 from the #1281 dogfood (one edit wedges a 151-file project) is confirmed and fixed**: a segment carried forward from the parent generation restores its symbol occurrence ids under the current generation; the decoder re-sorted symbols/edges/unresolved references but not `clone_bodies`, so `validate_clone_bodies` refused every incremental generation with ≥2 clone bodies in a carried file (`clone body evidence is not canonically bound to file symbols`) and the publication authority reported corruption on every wake. The base-red unit test `publication_store::partitioned_reclamation_is_bounded_and_preserves_retained_segments` is green with the one-line sort.
- Also: retained-segment reads take a shared store lock instead of a non-blocking exclusive try-lock (a retention tick aborted the whole projection typed; #1226 names it), and the journey accepts the lexical lane's policy-bounded `partial/candidate_sources_pruned` from the current generation (unrelated to timing; the probe query trips the df budget by construction).
- Builds on the lead's wip snapshot `ea04a8f2b` (chunker v5).

## Motivation

#1103 item 4 / #1226; #1281 dogfood B1. Full Hotpath tables, timelines, before/after and the test triage: project store `internal/lifecycle-1103/hotpath.md`; short outcome: `docs/1103-lifecycle-outcome.md`.

## Changes

- `registry.rs` `drive_text_projection`: the published-pass driver breaks once `query_owners_are_ready()`. Pinned by `convergence_park_tests::fresh_graph_activation_starts_while_the_clone_successor_is_pending`.
[...]

Issue #1103, opened 2026-09-08
test-transport daemon_suite: 772-file lifecycle batch writes ~2.4 GB of artifacts and misses the 45 s receipt; hooked saves reconcile only at the 30 s window


ScriptedAlchemy:
The `test-transport`-gated `daemon_suite` (not run in CI) has four deterministic failures on the #707 tip, reproduced with a `--features test-transport` binary on `dd002654e` (hauler cc-10049) and on `b20343a70` (cc-10176):

1. `indexing_lifecycle_test::ignored_dependency_admission_survives_physical_daemon_restart_without_widening` — after the physical daemon restart the terminal query answers `coverage.exact = {status: "stale", generation: …00000002…}` and `lexical` likewise `stale` where the journey requires `complete` (`code_index_journey.rs:421`). Until b20343a70 this test failed earlier on a one-page `tracedecay_retrieve` read; that helper bug is fixed and the real assertion is now reached. **Suspect:** the #1071 change (8051d0117) made restart admission (`activate_retained_generation_from_frontier`) verify the retained generation's sealed file digests; an explicitly admitted ignored dependency whose bytes are not in the sealed manifest (or are sealed under a different sanitization) would make restart mark the seat stale and reconcile instead of serving it as current. Confirm by running this test at f52868364 (pre-#1071) vs 8051d0117.
2. `git_watch_test::linked_worktree_requires_mount_then_serves_only_its_exact_generation` — `production-composition code index did not publish for '<iso>/wt-feat-y' after 20001 ms; composition gate capacity=4, admitted=1, waiting=0; scheduler_state=None; mounted_generation=None` (`git_watch_test.rs:301`). Same symptom as #1097's `linked_worktree_scope_retention_crash_replay_and_pure_inventory_journey` (`linked-b` never publishes), so this is very likely one linked-route defect independent of the semantic model; whichever issue fixes it first should close the other's linked-worktree case.
3. `indexing_lifecycle_test::mounted_incremental_lifecycle_preserves_only_complete_compatible_generations` — `timed out waiting for terminal generation for cancellation_probe_0000_000`; status shows `coverage: partial_refresh_in_progress`, `code_graph_serving: pending`, 129/129 files completed, 49,152 chunks committed (`code_index_journey.rs:413`). Do not raise the wait.
[...]

PR #1349, opened 2026-09-16
fix(code-index-runtime): require lane coverage for freshness ready

ScriptedAlchemy:
## Summary
- Failing unit assertion: freshness ready requires text owners and graph `Ready`.
- Production fix: stop treating graph `Refused`/`Unavailable` as terminal freshness.

## Test plan
- [x] `dashboard_ready_requires_text_and_graph_lane_owners` (red then green)
- [ ] `mounted_incremental_lifecycle_preserves_only_complete_compatible_generations`
- [ ] cancellation lifecycle variant

Closes nothing until lifecycle green on tip; then evaluate #1103 / #1339.


ScriptedAlchemy:
I am taking the separate incremental-index P0 on `agent/incremental-index-wedge`. I read this merged freshness change and will not modify `registry.rs` or `registry/serving_readiness_tests.rs` without posting the exact overlap here first. My current scope is the carried clone-body restore authority plus the existing reconcile error classification and typed status path. I will rerun both lifecycle variants against my exact head and report any inherited red instead of changing their bounds.

ScriptedAlchemy:
Exact merged-head rerun is still red at the unchanged bound. Hauler `cc-22271` ran `indexing_lifecycle_test::mounted_incremental_lifecycle_preserves_only_complete_compatible_generations` and failed after 140.12 s while waiting for `cancellation_probe_0000_000`. Status had 772/772 files committed, generation 3, `rebuild_in_flight=true`, and graph pending. I did not change the timeout or lifecycle files. The one-line clone-carry journey is independently green in `cc-22223`; this remaining red is the adjacent cancellation/throughput defect from #1103.

PR #1365, opened 2026-09-16
fix(code-index): return terminal sync admission

ScriptedAlchemy:
## Summary
- Carries the parked publication-corruption payload through the scheduler admission boundary.
- Refuses hook, branch-publication, and explicit sync work after `CorruptionResetRequired`.
- Makes `tracedecay sync` return `code_index_publication_authority_corrupt` with `retryable=false` and the exact parked reason.
- Regenerates the Rust dashboard schema omitted from the merged #1360.

Tracks #1352. #1360 was merged by another campaign actor before its red sync test had a production fix. This follow-up remains unmerged.

## Evidence
- Historical `8d16fc6e0` reproduction rejected `ci/utils.sh`. Both clone symbol bindings were present, but the pair was in reverse canonical order. The loop produced 153 projection failures in 43 seconds.
- Current 151-file ripgrep journey published generation 2 with `reextracted_files=1` and became fresh in 52 seconds with zero reconcile failures.
- Worker regression changed from 13 attempts after 12 later wakes to 1 attempt with `rebuild_in_flight=false`, `blocked_reason=publication_authority_corrupt`, and `retries_on_wake=false`.
- Sync regression changed from a queued success to a typed non-retryable error with the exact reset reason.

## Verification
- [x] `cc-22290` worker terminal-state test passed.
- [x] `cc-22300` explicit sync terminal-state test passed.
- [x] `cc-22285` exported the Rust dashboard contract schema. The generated TypeScript comparison passed.
- [x] Dashboard typecheck, 1,424 tests, and build passed on the merged product change.
- [ ] Rebased-head CLI build and focused journey chain is running as `cc-22306` through `cc-22309`.
- [ ] The existing mounted lifecycle remains red at the unchanged bound in `cc-22271`. It reached 772 of 772 files, then timed out waiting for `cancellation_probe_0000_000`. This is the adjacent #1103 cancellation and throughput defect.

---

## Order 25c — #1365 terminal-admission HOLD

**Do not merge** head `da2c47d7a7`.

Codex P1 (valid): https://github.com/ScriptedAlchemy/tracedecay/pull/1365#discussion_r4022949014

Root fix: admission consults **canonical terminal convergence park/authority** (or dedicated terminal flag), not observational `build_progress.snapshot()`. No parallel latch.

Tests required:
1. Malformed active publication pointer on first reconcile before any progress snapshot → park set; hook + branch publication + explicit sync return terminal typed corruption (never Accepted/queued).
2. Park visible before progress reason → same.
[...]

PR #1360, opened 2026-09-16
fix(code-index): park corrupt publication authority without retry (#1352)

ScriptedAlchemy:
## Summary
- Closes #1352: corrupt publication authority no longer retries forever.
- Terminal park: `blocked_reason=publication_authority_corrupt`, `retries_on_wake=false`, arrivals consumed.
- Receipts: Hauler `cc-22223` (clone-carry journey), `cc-22238` (isolation terminal).

## Test plan
- [x] isolation corrupt publication terminal
- [x] incremental clone-carry journey
- [ ] tip recheck after land (order 21)
- [ ] post-merge lifecycle + cancellation

Issue #1352, opened 2026-09-16
P0: one source edit corrupts incremental index and retries forever


ScriptedAlchemy:
## Defect

On campaign tip, a fresh 151-file ripgrep index publishes generation 1. Appending one `pub fn` publishes generation 2, then text projection rejects a carried clone body with `clone body evidence is not canonically bound to file symbols`. The graph head is quarantined. Reconcile then retries the terminal `publication authority is corrupt and requires an index reset` outcome about every 0.6 seconds. Status remains stale with `rebuild_in_flight=true`, no `blocked_reason`, and `sync` exits 0 without progress.

Reported dogfood evidence on `8d16fc6e0` counted 624 failures in 6 minutes. This issue requires a fresh isolated reproduction on the current campaign base with the first failing chunk identity inputs named.

## Required result

- Rebind carried clone bodies to the new generation symbol identities at the reuse or decode authority. Keep `validate_clone_bodies` strict.
- Classify publication-authority corruption as terminal in the existing reconcile policy. Stop after the first non-transient failure.
- Set `rebuild_in_flight=false` and a typed `blocked_reason`. Make `sync` report the terminal state.
- Add a production journey for one append and a bounded-attempt terminal-failure test.

## Related work

- #1103 tracks the larger lifecycle throughput suite.
- #1339 contains an unmerged one-line canonical clone-body restore in a broad lifecycle PR. It does not cover terminal retry classification or the stuck status and sync contract.
- #1340 records the dogfood journey that exposed the defect.
- #1349 changed freshness readiness only. Its files remain owned by that lifecycle change.

All runtime work must use an isolated home and data directory. Never use the operator daemon or `~/.tracedecay`.

ScriptedAlchemy:
Closed by #1360 merge `1a6ca405984b` on campaign tip. Receipts: cc-22223, cc-22238, cc-22272.

ScriptedAlchemy:
Reopened after exact-head review. The merged #1360 includes `admin_sync_reports_terminal_publication_corruption_without_queueing`, but it does not include the corresponding `handle_admin_sync` or reconcile-sink change. Hauler `cc-22245` proves the merged shape still returns `{"status":"queued"}` and calls the sink after terminal corruption. I am carrying the missing canonical scheduler-admission fix in a follow-up. #1360 was merged by another campaign actor while this owner was still verifying; the task explicitly required no merge.

ScriptedAlchemy:
[...]

PR #1368, opened 2026-09-16
fix(code-index): return terminal sync admission

ScriptedAlchemy:
## Summary
- Carries the parked publication-corruption payload through the scheduler admission boundary.
- Refuses hook, branch-publication, and explicit sync work after `CorruptionResetRequired`.
- Makes `tracedecay sync` return `code_index_publication_authority_corrupt` with `retryable=false` and the exact parked reason.
- Regenerates the Rust dashboard schema omitted from merged #1360.

Tracks #1352. #1360 was merged by another campaign actor before its red sync test had a production fix. This follow-up remains unmerged.

## Evidence
- Historical `8d16fc6e0` reproduction rejected `ci/utils.sh`. Both clone symbol bindings were present, but the pair was in reverse canonical order. The loop produced 153 projection failures in 43 seconds.
- Current 151-file ripgrep journey published generation 2 with `reextracted_files=1` and became fresh in 52 seconds with zero reconcile failures.
- Worker regression changed from 13 attempts after 12 later wakes to 1 attempt with `rebuild_in_flight=false`, `blocked_reason=publication_authority_corrupt`, and `retries_on_wake=false`.
- Sync regression changed from a queued success to a typed non-retryable error carrying the exact reset reason.

## Verification
- [x] Exact rebased head `da2c47d7a730e9bad38861b64265312bf216ba29`.
- [x] `cc-22306` built the Hotpath and test-transport CLI.
- [x] `cc-22307` passed the explicit sync terminal-state test.
- [x] `cc-22308` passed the worker terminal-state and admission test.
- [x] `cc-22309` passed the one-line append clone-carry daemon journey.
- [x] `cc-22285` exported the Rust dashboard contract schema. The generated TypeScript comparison passed.
- [x] Dashboard typecheck, 1,424 tests, and build passed on the merged product change.
- [ ] The existing mounted lifecycle remains red at the unchanged bound in `cc-22271`. It reached 772 of 772 files, then timed out waiting for `cancellation_probe_0000_000`. This is the adjacent #1103 cancellation and throughput defect.

---

## Closure decision (2026-09-16)

Closed unmerged as a duplicate of draft PR #1365 at the same old head `da2c47d7a7`. This head has confirmed admission defects: cold corruption can still report queued success because admission reads observational progress instead of the canonical convergence park; exact-path hooks erase terminal corruption through a bool facade; text projection can clear/replace the terminal park; branch publication does not recheck the park mid-wait. #1365 remains the sole owner and must resolve those findings before review/merge.

---

## Reopened as sole correction lane

#1365 was closed after this branch was designated its replacement. This PR is reopened **draft** only to receive the corrected implementation. Head `da2c47d7a7` remains blocked and must not merge. Required before ready: canonical convergence-park admission on cold corruption; terminal park cannot be cleared/replaced by text projection; branch publication rechecks terminal park mid-wait; exact-path hooks preserve typed corruption rather than bool→Unavailable; reset/remount recovery proven; both independent reviews rerun.

---

## Order 25c — sole correction lane (owner 8869b79b)

`#1365` stays closed. This DRAFT is the only admission-fix PR.

**Old head `da2c47d7a7` is blocked.** Push a new head on `agent/incremental-index-wedge-sync-terminal-rebased`.

Convergence park is the sole terminal authority (no parallel latch). Required: cold park-without-snapshot; park preservation vs text-projection clear; branch wait-loop mid-park typed/nonretryable; exact-path typed hooks. Dual independent review on the **new** head before merge.

---

## Order 25 new head

- Head `8d44959027` on tip `dd8e56dbf171` replaces blocked `da2c47d7`.
- Typed convergence park; `PublicationAuthorityCorrupt` / `LinkedWorktreeDisabled` / `Unavailable` / `Accepted` merge; `From<bool>` deleted.
- DRAFT until two change-risk reviewers pass on this head.


ScriptedAlchemy (commented):
## Review (draft head `da2c47d7a7`) — withhold approve

Outcome: typed overflow admission is the right shape, but this head still fails the reopen bar listed in the PR body. Do not undraft/merge until park is the sole admission authority and exact-path hooks stop erasing terminal corruption through `bool`.

**Holds (inline):**
- `[P1]` / Blocker — `publication_authority_requires_reset` keys on observational `build_progress.blocked_reason`, not `convergence_park` (cold park-without-snapshot can still admit).
- `[P1]` — `notify_path` / `notify_hook_paths` keep a `bool` facade and collapse terminal corruption to `false`.
- `[P2]` — branch publication only checks admission once before the wait loop; no mid-wait park recheck.
- soft — `"code_index_publication_authority_corrupt"` is an untyped string protocol beside a named `CODE_INDEX_SCHEDULER_UNAVAILABLE` constant.

Regression coverage for overflow→typed corrupt is good; still missing: cold park-without-snapshot refusal, exact-path typed preservation, and branch wait-loop mid-park. Matches the required list in the reopen notes — prefer those tests before ready.

Draft → COMMENT only (no APPROVE).

crates/tracedecay-code-index-runtime/src/code_index_scheduler/registry.rs
```diff
@@ -2377,6 +2384,20 @@ impl CodeIndexSchedulerRegistryV1 {
         self.mounted.lock().await.contains_key(&project_root)
     }
 
+    fn publication_authority_requires_reset(worktree: &MountedCodeIndexWorktreeV1) -> bool {
```
ScriptedAlchemy: [P1] / Blocker — admission authority is still observational progress, not the convergence park.

Could cold `park`-without-`snapshot` still pass this gate and wake? `snapshot().and_then(|s| s.blocked_reason)` is None when progress was never written or was cleared, so `publication_authority_requires_reset` returns false and callers continue to `Accepted` / path enqueue.

The reopen bar names convergence park as the sole terminal authority. Prefer reading `worktree.convergence_park` here (or a helper that treats park presence as reset-required), and only consult progress as a secondary signal — otherwise overflow's typed return and this gate disagree whenever park and progress diverge.

Also: when this returns true but park is `None`, `notify_hook_overflow` maps to `Unavailable` (retryable surface) instead of a typed terminal — that cold hole should stay non-retryable.

crates/tracedecay-code-index-runtime/src/code_index_scheduler/registry.rs
```diff
@@ -2388,6 +2409,9 @@ impl CodeIndexSchedulerRegistryV1 {
             let Some(worktree) = mounted.get(&project_root) else {
                 return false;
             };
+            if Self::publication_authority_requires_reset(worktree) {
```
ScriptedAlchemy: [P1] — exact-path hooks still erase terminal corruption through a `bool` facade.

`false` here is indistinguishable from "not mounted" / canonicalize failure. Callers that treat `!accepted` as unavailable will queue retry or report the wrong route code.

Could `notify_path` / `notify_hook_paths` return `CodeIndexReconcileAdmissionV1` the same way `notify_hook_overflow` now does, and plumb `PublicationAuthorityCorrupt` through the activation/hook sinks? A focused regression that injects a parked worktree and asserts exact-path admission stays typed/non-retryable would lock this.

crates/tracedecay-code-index-runtime/src/code_index_scheduler/registry.rs
```diff
@@ -2417,6 +2441,9 @@ impl CodeIndexSchedulerRegistryV1 {
...
```
ScriptedAlchemy: [P1] — same bool collapse on the primary host after-file-edit path (`notify_hook_paths`).

Overflow became typed; this remains `return false` on reset-required. Prefer one admission type across path / paths / overflow so the MCP hook surface cannot lose `PublicationAuthorityCorrupt`.

crates/tracedecay-code-index-runtime/src/code_index_scheduler/branch_publication.rs
```diff
@@ -443,18 +443,28 @@ impl BranchPublicationContextV1 {
                 ),
             ));
         }
-        if !schedulers
+        match schedulers
             .notify_hook_overflow(canonical_worktree_root)
             .await
```
ScriptedAlchemy, line 449: [P2] — park is checked once at overflow notify, then the wait loop below never re-admits.

If corruption parks mid-wait (or progress/park lands after this match), the loop can idle-timeout or keep polling serving readiness instead of returning typed `code_index_publication_authority_corrupt` / non-retryable.

Could each iteration (or at least before timeout errors) re-read `convergence_park` / admission and fail terminal immediately? A regression that parks after the initial Accepted would prove it.

crates/tracedecay/src/mcp/server.rs
```diff
@@ -118,20 +119,34 @@ impl ServerStats {
...
     LinkedWorktreeDisabled,
+    PublicationAuthorityCorrupt(CodeIndexConvergenceParkedV1),
     Unavailable,
 }
 
+pub(crate) fn code_index_publication_corrupt(
+    parked: CodeIndexConvergenceParkedV1,
+) -> TraceDecayError {
+    TraceDecayError::project_route(
+        "code_index_publication_authority_corrupt",
```
ScriptedAlchemy: soft / discipline — untyped string protocol next to a named constant.

`CODE_INDEX_SCHEDULER_UNAVAILABLE` is pinned; this reason code is a bare `"code_index_publication_authority_corrupt"` in `project_route` (and again in `branch_publication` / `host_outcome`). Prefer a named constant (or typed field) shared by producer and the dispatch test assertion so the cross-boundary contract cannot drift.

crates/tracedecay/src/daemon/project_composition/code_index_activation.rs
```diff
@@ -302,7 +304,10 @@ pub(super) fn code_index_activation_hint_sink(
                     .await
             };
             let overflow_accepted = if batch.overflow {
-                schedulers.notify_hook_overflow(&project_root).await
+                matches!(
```
ScriptedAlchemy: soft — overflow corruption collapses to `false` here, same as Unavailable.

Fine if this sink is fire-and-forget hint-only, but then the MCP reconcile / hook sinks must be the ones that preserve typed terminal (today exact-path still cannot). If any caller of this hint path surfaces admission to operators, match `PublicationAuthorityCorrupt` explicitly instead of `matches!(…, Accepted)`.

crates/tracedecay/src/mcp/server.rs
```diff
@@ -118,20 +119,34 @@ impl ServerStats {
...
+}
+
 impl CodeIndexAdmission {
     pub(crate) fn host_outcome(self) -> HostAdmissionOutcome {
         match self {
             Self::Accepted => HostAdmissionOutcome::replay_completed(true, false),
             Self::LinkedWorktreeDisabled => {
                 HostAdmissionOutcome::degraded("linked_worktree_disabled")
             }
+            Self::PublicationAuthorityCorrupt(_) => {
```
ScriptedAlchemy: [P2] — `PublicationAuthorityCorrupt` maps to `HostAdmissionOutcome::degraded(...)`.

“Degraded” reads as soft/recoverable; this state needs an explicit reset and must not look like ordinary capacity degradation. Prefer a terminal/non-retryable host outcome (or a dedicated token) so hosts do not keep poking the parked worktree.

Related: tip still carries `impl From<bool> for CodeIndexAdmission` just below this match — that converter is the root of exact-path collapses. Delete it; force every sink to return the typed enum explicitly.

crates/tracedecay/src/daemon/project_composition/code_index_activation.rs
```diff
@@ -302,7 +304,10 @@ pub(super) fn code_index_activation_hint_sink(
...
+                    schedulers.notify_hook_overflow(&project_root).await,
+                    CodeIndexReconcileAdmissionV1::Accepted
```
ScriptedAlchemy: [P1] — overflow acceptance is still a bool façade: `matches!(…, Accepted)`.

`PublicationAuthorityCorrupt` and `Unavailable` both become `false` here, then fold into `paths_accepted && overflow_accepted`. The typed enum exists; this arm throws it away. Propagate `CodeIndexReconcileAdmissionV1` (or map Corrupt to a typed sink failure) instead of a bool AND.

crates/tracedecay/src/daemon/project_composition/code_index_activation.rs
```diff
@@ -363,8 +368,16 @@ pub(super) fn code_index_reconcile_sink(
...
+                match schedulers.notify_hook_overflow(&root).await {
+                    CodeIndexReconcileAdmissionV1::Accepted => {
+                        return crate::mcp::server::CodeIndexAdmission::Accepted;
+                    }
+                    CodeIndexReconcileAdmissionV1::PublicationAuthorityCorrupt(parked) => {
+                        return crate::mcp::server::CodeIndexAdmission::PublicationAuthorityCorrupt(
+                            parked,
+                        );
+                    }
+                    CodeIndexReconcileAdmissionV1::Unavailable => {}
```
ScriptedAlchemy: [P2] — after the typed match, `Unavailable` falls through into `match demand` which still ends in `activation.notify_hook_overflow(...).await.into()` / other `.into()` bool converters (outside this hunk).

That is a second overflow notify plus a bool façade on the same wake. If the first match already observed Unavailable/Corrupt, do not re-notify; if activation's overflow is still `bool`, lift it to the same typed admission.

crates/tracedecay/src/mcp/tools/handlers/info/status.rs
```diff
@@ -21,21 +22,27 @@ pub(crate) async fn handle_admin_sync(
...
-            true,
-            "admin sync was not accepted by the code-index scheduler",
-        ));
+    .await;
+    match admission {
+        crate::mcp::server::CodeIndexAdmission::Accepted => {}
+        crate::mcp::server::CodeIndexAdmission::PublicationAuthorityCorrupt(parked) => {
+            return Err(crate::mcp::server::code_index_publication_corrupt(parked));
+        }
+        crate::mcp::server::CodeIndexAdmission::LinkedWorktreeDisabled
```
ScriptedAlchemy: [P2] — `LinkedWorktreeDisabled | Unavailable` collapse to one `code_index_scheduler_unavailable` with `retryable: true`.

Linked-worktree policy refusal is not a transient scheduler blip; operators will retry forever. Keep discriminants (and retryability) distinct the way `PublicationAuthorityCorrupt` already is.

ScriptedAlchemy (commented):
## rspack-team-mode follow-up (new angles only)

Four more inline notes beyond the existing bool-admission / park threads. Still withhold APPROVE until admission is typed end-to-end — collapsing Corrupt via `matches!(… Accepted)` / `From<bool>` keeps the hole open.

crates/tracedecay-code-index-runtime/src/code_index_scheduler/registry.rs
```diff
@@ -2475,9 +2743,41 @@ impl CodeIndexSchedulerRegistryV1 {
     /// signature, then sealed file digests) refreshes the scheduler's cadence
     /// watermark and returns without extraction.
     pub async fn probe_freshness(&self, project_root: &Path) -> bool {
```
ScriptedAlchemy: [P2] probe — residual bool facade (will be in real review)

ScriptedAlchemy (commented):
## rspack-team-mode pending probe — will submit

ScriptedAlchemy (commented):
## rspack-team-mode — tip `21d77340da26` (new head)

Outcome: prior tip `da2c47d7a7` holds look **cleared** on this head — park is the admission authority (`publication_authority_reset` / typed `blocked_reason`), path/overflow/exact-path return `CodeIndexReconcileAdmissionV1`, branch wait re-reads mid-loop, host maps Corrupt via `terminal_unavailable`, and `CODE_INDEX_PUBLICATION_AUTHORITY_CORRUPT` is pinned. Draft → **COMMENT only** (no APPROVE). Do not undraft/merge until CI is green on this tip.

### Prior holds (closed on this tip)
- observational `build_progress.blocked_reason` gate → park-keyed
- bool facade on `notify_path` / `notify_hook_paths` / overflow → typed
- mid-wait branch re-admit → `publication_authority_corruption` each iteration
- degraded mapping / bare reason string → terminal + named constant

### New notes only
Inline: residual bool `probe_freshness` collapse; bare `"code_index_scheduler_unavailable"` still beside the named constant in `hook_writes`. Soft process: dashboard schema regen from the #1360 omission rides the same PR as the admission reopen — prefer landing contract regen separately if it can fail independently.

Still withhold APPROVE on draft.

crates/tracedecay-code-index-runtime/src/code_index_scheduler/registry.rs
ScriptedAlchemy: [P2] — residual bool facade on the ordinary freshness probe.

`probe_freshness_admission` is the typed path (good), but this wrapper still collapses `PublicationAuthorityCorrupt` into `false` via `matches!(… Accepted)`. Any leftover caller that keys retries on `!probe_freshness` will treat terminal corruption as a transient miss.

Could you deprecate or remove the bool API now that the freshness sink already uses admission, or make every production caller take `probe_freshness_admission` only?

crates/tracedecay/src/mcp/server/hook_writes.rs
```diff
@@ -97,6 +97,9 @@ pub(crate) async fn execute_background_refresh_direct(
         super::CodeIndexAdmission::LinkedWorktreeDisabled => {
             Ok(BackgroundRefreshOutcome::LinkedWorktreeDisabled)
         }
+        super::CodeIndexAdmission::PublicationAuthorityCorrupt(parked) => {
+            Err(super::code_index_publication_corrupt(parked))
+        }
         super::CodeIndexAdmission::Unavailable => Err(TraceDecayError::project_route(
             "code_index_scheduler_unavailable",
```
ScriptedAlchemy: soft / discipline — untyped string beside a named constant.

`CODE_INDEX_SCHEDULER_UNAVAILABLE` is pinned in `mcp/server.rs`, and Corrupt now goes through `code_index_publication_corrupt`. This `Unavailable` arm still spells `"code_index_scheduler_unavailable"` literally (same file has other bare copies). Prefer the named constant so the wire token cannot drift from admin-sync / host admission.

crates/tracedecay-code-index-runtime/src/code_index_scheduler/registry.rs
ScriptedAlchemy: Superseded by the tip `21d77340da26` review inline on the same line (real [P2] text). Ignore this probe comment from the pending-review dry run.

crates/tracedecay/src/mcp/server.rs
```diff
@@ -117,41 +123,54 @@ impl ServerStats {
...
+) -> HostAdmissionOutcome {
+    match admission {
+        CodeIndexDemandAdmissionV1::Queued => HostAdmissionOutcome::replay_completed(true, false),
+        CodeIndexDemandAdmissionV1::RefusedByPolicy => {
+            HostAdmissionOutcome::degraded(CODE_INDEX_LINKED_WORKTREE_DISABLED)
+        }
+        CodeIndexDemandAdmissionV1::Terminal(_) => {
+            HostAdmissionOutcome::terminal_unavailable(CODE_INDEX_PUBLICATION_AUTHORITY_CORRUPT)
+        }
+        CodeIndexDemandAdmissionV1::Unavailable(_) => {
```
ScriptedAlchemy: [P2] — typed demand unavailable causes still collapse at the host wire.

`CodeIndexDemandUnavailableV1` now has stable labels (`foreign_root`, `no_proven_change`, `route_retired`, `scheduler_unmounted`), but `code_index_host_outcome` matches `Unavailable(_)` and always emits retryable `CODE_INDEX_SCHEDULER_UNAVAILABLE`. A foreign-root or no-proven-change refusal then looks like scheduler pressure.

Could you map each cause to its own pinned reason (or at least non-retryable where a retry cannot help), so the front-door types survive the host boundary?

crates/tracedecay-code-index-runtime/src/code_index_scheduler/activation.rs
```diff
@@ -339,95 +350,74 @@ impl CodeIndexActivationV1 {
...
+        if !self.accepts_root(project_root) {
+            return CodeIndexDemandAdmissionV1::Unavailable(
+                CodeIndexDemandUnavailableV1::ForeignRoot,
+            );
+        }
+        let overflow = !matches!(demand, CodeIndexDemandV1::HookPaths(_));
+        let rel_paths = match demand {
+            CodeIndexDemandV1::HookPaths(rel_paths) => {
+                if rel_paths.is_empty() {
+                    return CodeIndexDemandAdmissionV1::Queued;
```
ScriptedAlchemy: soft — empty `HookPaths` returns `Queued`.

An empty path batch used to be a no-op refusal; reporting `Queued` can read as "scheduler accepted work" to callers that only check `is_queued()`. Prefer an explicit no-op variant, or keep `Unavailable(NoProvenChange)` / early return without claiming a queue seat.

ScriptedAlchemy (commented):
## rspack-team-mode — tip `b539a34d08a1` (new head)

Outcome: prior tip `21d77340da26` residuals look **cleared** — bool `probe_freshness` façade is gone, demand routes through one `admit` door (`CodeIndexDemandAdmissionV1`), and `hook_writes` uses `CODE_INDEX_SCHEDULER_UNAVAILABLE`. Draft → **COMMENT only** (no APPROVE).

### Prior holds (closed on this tip)
- residual bool freshness probe → removed / typed demand
- bare `"code_index_scheduler_unavailable"` beside named constant → pinned constant

### New notes only
Inline: host boundary still collapses every `Unavailable(_)` cause into one retryable `code_index_scheduler_unavailable`, discarding the typed `CodeIndexDemandUnavailableV1` labels (`foreign_root`, `no_proven_change`, …). Soft: empty `HookPaths` returns `Queued`.

Withhold APPROVE on draft until CI green and the host mapping preserves cause.

ScriptedAlchemy (commented):
## rspack-team-mode — tip `ca2cf026557a` (draft → COMMENT)

Outcome: prior tip `b539a34d08a1` residuals look **cleared** on this head.

### Cleared
- [P2] typed `CodeIndexDemandUnavailableV1` causes no longer collapse at the host wire — `code_index_unavailable_host_outcome` / `code_index_unavailable_error` map each discriminant to its own pinned reason + retry bit.
- soft empty `HookPaths` → `Queued` is gone: empty batches now return `Unavailable(NoProvenChange)` and the regression test pins mount-attempts stay 0.

Draft → **COMMENT only**. Tracks #1352.

PR #1370, opened 2026-09-16
fix(ui-code): keep same-depth Y and 1s active poll

ScriptedAlchemy:
## Summary

Post-#1364 P1 follow-up only (into campaign tip `codex/tracedecay-total-redesign-plan-reopened`, never #707):

- Same-depth cortex regions keep semantic `bandY(depth)` — remove wrap that invented false elevation
- Active-build freshness polling stays at 1s — remove unchanged-read backoff that stretched the UI budget
- Stall copy ("no progress for") retained; grant/retry not reintroduced

## Change-risk (local)

Subtractive revert of the two Codex P1 surfaces only. Residual risks accepted by contract: crowded strata may render denser (semantic Y over wrap legibility); a stuck active build continues 1 Hz polls (honest stall signal over budget stretch). No production callers of removed `activeBuildPollInterval`. Diff is four dashboard files.

## Test plan

- [x] `npm run test -- --run src/workspaces/code/` → 19 files / 129 passed
- [x] cortex crowded-band semantic Y: 20 regions share `bandY` (638)
- [x] browser fixture: Code topology + stalled freshness region renders (`BROWSER_PROOF_OK`)
- [x] no grant/retry reintroduction in diff


dashboard/src/workspaces/code/IndexFreshness.tsx
```diff
@@ -39,53 +39,17 @@ import { elideStart, formatCount, formatMicrosUtc, splitBytes } from '../../ui/f
...
-}
-
-export function activeBuildPollInterval(unchangedReads: number): number {
-  return Math.min(IDLE_POLL_MS, ACTIVE_POLL_MS * 2 ** Math.min(unchangedReads, 5));
-}
-
 export function IndexFreshness() {
   const scope = useScope((s) => s.scope);
-  const cadence = useRef({ signature: '', unchangedReads: 0, dataUpdateCount: -1 });
   const freshness = useQuery({
```
ScriptedAlchemy, line 48: [P2] This restores a hard 1 Hz active poll for as long as the build is open, including when progress is unchanged. The deleted comment cited a measured stuck scheduler ( for ten minutes) as the reason backoff existed. Why remove the bound without an alternate (stall-only interval after N unchanged reads, page-visibility gate, or max active poll budget)? A stuck page now hits  once a second forever.

dashboard/src/workspaces/code/IndexFreshness.dom.test.tsx
```diff
@@ -353,7 +353,7 @@ describe('Code index freshness', () => {
     expect(fetch).toHaveBeenCalledTimes(3);
   });
 
-  it('backs off while an active build reports the same progress, and says how long it has been quiet', async () => {
+  it('keeps polling an unchanged active build each second and reports how long it has been quiet', async () => {
```
ScriptedAlchemy, line 356: [P2] The test now encodes the unbounded 1s cadence as the desired contract. If the product intent is 'always show stall age quickly', keep a short first interval but still back off after unchanged reads so the test does not lock in the expensive steady state.

dashboard/src/workspaces/code/cortexRelief.ts
```diff
@@ -190,42 +185,16 @@ export function buildCortexModel(measurement: StrataMeasurementV1): CortexModel
...
-    return rows;
-  };
-  const bandRows = new Map<number, number>();
-  for (const [depth, band] of byBand) bandRows.set(depth, rowsFor(band.length));
-  const widestRow = [...byBand.entries()].reduce(
-    (max, [depth, band]) => Math.max(max, Math.ceil(band.length / (bandRows.get(depth) ?? 1))),
-    1,
-  );
-  const mostRows = [...bandRows.values()].reduce((max, rows) => Math.max(max, rows), 1);
+  const widestBand = [...byBand.values()].reduce((max, band) => Math.max(max, band.length), 1);
```
ScriptedAlchemy, line 188: [P2] Pinning every same-depth region to one  undoes the measured ripgrep hairline fix (28 regions → ~18px dots). Crowding now shrinks  / radius instead of wrapping rows. Was overlap/legibility re-measured on a shallow wide tree, or only asserted via y-equality in unit tests?

dashboard/src/workspaces/code/cortexRelief.ts
```diff
@@ -237,24 +206,10 @@ export function buildCortexModel(measurement: StrataMeasurementV1): CortexModel
...
-      const row = Math.floor(index / perRow);
-      const rowStart = row * perRow;
-      const rowLength = Math.min(perRow, inBand.length - rowStart);
       placed.set(draft.cluster.directory, {
-        x: PAD.left + ((index - rowStart + 0.5) / rowLength) * usableWidth,
-        y: rows === 1 ? line : bandTop + (row + 0.5) * pitch,
+        x: PAD.left + ((index + 0.5) / inBand.length) * usableWidth,
+        y: bandY(depth),
         radius: MIN_RADIUS + areaScale * Math.sqrt(draft.cluster.file_count),
       });
```
ScriptedAlchemy, line 214: Soft:  still floors at 18 while  is 13, so crowded bands will collide horizontally before the √-area law can shrink bodies. If same-depth-y is non-negotiable, consider a max-per-band cap or explicit overflow affordance rather than silently overlapping.

dashboard/src/workspaces/code/cortexRelief.test.ts
```diff
@@ -119,6 +119,35 @@ describe('elevation', () => {
     expect(model.strata.map((band) => band.depth)).toEqual([0, 4]);
     expect(model.strata.every((band) => band.regions === 1)).toBe(true);
   });
+
+  it('keeps every same-depth region on the stratum line when the band is crowded', () => {
+    const count = 20;
+    const clusters = Array.from({ length: count }, (_, index) =>
+      cluster(`src/mod${index}`, { order: index, file_count: 4 }),
+    );
+    const files = Array.from({ length: count }, (_, index) => file(`src/mod${index}/a.rs`, 1));
```
ScriptedAlchemy, line 128: Soft: these cases prove y-collocation, not label/body collision. A focused assertion on minimum center distance or rendered radius under  would catch the density regression the old wrap path was solving.

ScriptedAlchemy (commented):
rspack-team-mode deep pass on #1370 (not an approve — behavior widen on polling + layout).

Themes: (1) stall poll cadence, (2) crowded-band density vs measured hairline fix tradeoff.

PR #1374, opened 2026-09-16
[HOLD] fix(code-index-runtime): seat the graph before the clone backfill (#1103, #1226)

ScriptedAlchemy:
## Summary

Successor to the closed #1339, rebuilt on tip `aeb2b5c35` (after #1360/#1329) with **only the independent parts** — no clone-body binding hunks (owned by #1360), no chunker wip.

- The graph seat waited for the whole text projection task, which continued past exact/lexical readiness into the clone-fingerprint successor (second full decode of the sealed source into a second artifact, ~27 s serial on the 772-file lifecycle fixture). The published driver now stops at owners-ready; the backfill is demand-driven on the retained path, runs without the owner-pass guard (status stays `fresh`), and owes no successor pass.
- Immutable retained-segment reads take a shared store lock instead of a non-blocking exclusive try-lock (a retention tick aborted the whole projection typed — the lock #1226 names).
- `text_projection_needs_work` and the #1351 clone-index census (`clone_successor_progress`) no longer block on the projection slot behind a running backfill slice.
- The journey accepts the lexical lane's typed policy-bounded `partial/candidate_sources_pruned` from the current generation.

## Measurements (4 vCPU / 15 GiB rig, perf profile, `--features test-transport,hotpath`, isolated fixture home, 45 s bound untouched)

| | tip `aeb2b5c35` | this branch |
| --- | --- | --- |
| journey `mounted_incremental_lifecycle_preserves_only_complete_compatible_generations` at 45 s | FAIL 2/2 (`index_build`/`verification`, graph pending) | **FAIL 2/2** (`verification`, graph pending) |
| restart → `code_index_serving_generation_seated` | 89.8 s / 92 s (base `46698151a`; tip not run to completion) | **62.4 s** (reconcile 12.7 + text 37.7 incl. finalization 6.5 + graph 10.8) |
| same fixes + chunker v5 wip (`ea04a8f2b`, halves this fixture's documents) | — | PASS 2/2 at 45 s, seat 39.0 / 38.7 s (measured on #1339 before it was closed) |

So: this PR removes the dominant wait (~27 s) and the lock abort, but on this rig the remaining serial chain (reconcile ≈ 13 s → text build ≈ 38 s → graph ≈ 11 s) still exceeds 45 s without the volume cut the lead's wip snapshot provides. Whether v5 lands here or separately is the lead's call; #1103 cannot be closed on this PR alone.

## Changes

[...]

Issue #1226, opened 2026-09-11
code-index: restart after SIGTERM mid-refresh rebuilds every file; 772-file batch misses its receipt


ScriptedAlchemy:
Split out of #1103 item 3 (`indexing_lifecycle_test::mounted_incremental_lifecycle_preserves_only_complete_compatible_generations`, test-transport `daemon_suite`), which stays red on the tip after the other three #1103 items landed.

**Observed.** After SIGTERM mid-refresh, restart overflow is a designed `Noop` (revision-7 graph-head recovery on an empty seat) and the successor then rebuilds all **772** files from scratch. On a loaded 96-core host the 45 s receipt bound expires still inside `bulk_commit` at 641/772 (~23 files/s, `code_graph_serving=pending`, hauler cc-15898). On a quieter tip: seal ~16–19 s, then text ~24–27 s, and `GraphSeatGateV1` waits for `published_text_is_ready`, so graph never starts inside the bound. The owner is artifact volume plus a strictly sequential seal → text → graph pipeline, not a hung state machine.

**Not the fix:** raising `RECEIPT_TIMEOUT` (limits are symptoms). Overlapping graph prepare/activate with text after `Published` was tried and reverted: it hits the exclusive flock on the sealed lexical generation store (`sealed lexical source generation store is busy`); with shared reader locks the error goes away but seal alone still exceeds the bound on this host.

**Required.** Instrument with Hotpath and decompose: (a) why a SIGTERM mid-refresh loses the retained generation's progress entirely instead of resuming from the retained head (`4bb8758e3 resume semantics from retained head` exists — why does it not apply here, and is the `Noop` recovery the right answer when a complete retained generation and a partial successor both exist?); (b) `bulk_commit` at 23 files/s for ~67 KB/file is ~1.5 MB/s — compare against what the sealed encoder should cost for its inputs (the #852 measurement puts the pre-seat path at ~8.7× N bytes allocated; the encode itself is much cheaper than this throughput implies — find the serial phase or per-file sync); (c) whether text and graph seating can genuinely overlap once the generation-store lock is a reader lock for the graph side. Ties to #800 (delta-proportional capture and seal) — a one-file sync after restart must not be a 772-file rebuild.

[...]

PR #1377, opened 2026-09-16
fix(code-index-runtime): seat before clone backfill; share owners-ready (#1103)

ScriptedAlchemy:
## Summary
- Route #707 review finding [r4023329939](https://github.com/ScriptedAlchemy/tracedecay/pull/707#discussion_r4023329939) into this lane (never #707): `query_owners_are_ready` is the sole exact/lexical-ready bit at both the published graph seat gate and the full sealed-generation replay skip, via one named helper `exact_and_lexical_ready_for_graph`. Bidirectional test: lexical-incomplete refuses seat/replay; owners-ready with clone backfill unfinished admits both.
- Continues the #1103 / #1226 seat-before-backfill work: published driver stops at owners-ready; clone backfill is demand-driven after the seated generation matches, outside the pass guard; retained-segment reads use a shared store lock; projection probes use `try_lock`.

Successor to closed #1339; supersedes parallel draft #1374 (`agent/1103-seat-before-backfill`) with tip-rebased commits plus the predicate pin and post-seat backfill wake.

## Test plan
- [x] `cargo test -p tracedecay-code-index-runtime --locked query_owners_ready_admits_seat_and_replay_both_directions`
- [x] `cargo test -p tracedecay-code-index-runtime --locked a_publication_seats_its_own_generation_without_waiting_for_a_quiet_tree`
- [ ] Measure `daemon_suite` journey `mounted_incremental_lifecycle_preserves_only_complete_compatible_generations` at the 45 s bound (receipts in PR comments); close #1103/#1226 only with exact green receipts if same root


ScriptedAlchemy (commented):
## rspack-team-mode — tip `0458588b5087`

Outcome: seat-before-clone-backfill with a single `query_owners_are_ready` bit (and `exact_and_lexical_ready_for_graph`) is the right shape for #1103 / #1226. Shared read lock + `try_lock` probe + one inline clone slice are coherent. **COMMENT** (not empty APPROVE) until the swallowed-admission and journey-string items are addressed and CI is green.

Not duplicating the existing Codex [P1] on unbounded `lock_shared()` / [P2] on `drive_retained` after expired proof — those still stand.

### Top new findings
- `[P2]` executor discards `request_query_background_reconcile` admission (`let _ = …`) including terminal Corrupt.
- `[P2]` journey treats `coverage.reason == "candidate_sources_pruned"` as terminal via an untyped string protocol — unexplained acceptance of partial as done.
- soft: one-slice clone warmup returning `Ok(false)` while work remains can read as generation-unavailable to similar/redundancy.

Could you keep seat/backfill atomic, but pin the journey terminal rule to a typed coverage discriminant (or named constant + regression) and propagate reconcile admission on the request path?

crates/tracedecay-code-index-runtime/src/code_index_executor.rs
```diff
@@ -1419,6 +1419,10 @@ where
                     return unavailable(code_search::CodeIndexSearchUnavailableReasonV1::Internal);
                 }
             }
+            // Clone backfill is retained-worker work after the seat. Kick that
+            // wake before any inline slice so a quiet daemon does not strand
+            // the successor on this request thread.
+            let _ = schedulers.request_query_background_reconcile(&scope).await;
```
ScriptedAlchemy: [P2] — swallowed reconcile admission on the similar/redundancy request path.

`request_query_background_reconcile` can refuse with the same typed vocabulary as hooks (`Unavailable` / park / Corrupt once that surface grows). `let _ = …` throws the verdict away, then the request still drives an inline clone slice.

Could you match on the admission (at least fail closed on terminal Corrupt / non-retryable) before `finish_clone_similarity_warmup_for_request`? Same pattern at the sibling call ~L1592.

crates/tracedecay/tests/daemon_suite/code_index_journey.rs
```diff
@@ -18,6 +18,30 @@ use crate::common::{DaemonProcess, tracedecay_command_with_home};
...
+/// the current-generation form is terminal: `generation` names an older
+/// generation whenever the request fell back to one, and that stays warming.
+fn lanes_short_of_terminal(search: &Value) -> Vec<&'static str> {
+    crate::common::incomplete_code_index_query_lanes(search)
+        .into_iter()
+        .filter(|lane| {
+            let coverage = &search["coverage"][*lane];
+            !(*lane == "lexical"
+                && coverage["status"] == "partial"
+                && coverage["reason"] == "candidate_sources_pruned"
```
ScriptedAlchemy: [P2] — untyped string protocol decides "terminal generation".

`coverage["reason"] == "candidate_sources_pruned"` keys correctness on an English/machine string inside JSON. A rename or a partial-from-stale-generation that reuses the same reason would green the journey while search is still warming.

Why not a typed coverage discriminant (or a pinned named constant + comment + regression that stale-generation partial still waits)? Do not merge on unexplained acceptance of partial as done.

crates/tracedecay-code-index-runtime/src/code_index_scheduler/serving.rs
```diff
@@ -1842,8 +1866,8 @@ impl LatestCodeTextGenerationV1 {
                 request_control,
             )?;
             advances += 1;
-            if advances >= TEXT_ARTIFACT_MAXIMUM_OWNER_WARMUP_ADVANCES_V1 {
-                return Ok(false);
+            if advances >= TEXT_ARTIFACT_MAXIMUM_CLONE_WARMUP_ADVANCES_V1 {
+                return Ok(!self.text_projection_needs_work());
```
ScriptedAlchemy: soft — one inline slice then `Ok(!needs_work)` can surface as hard unavailable.

After the single allowed advance, remaining clone work returns `Ok(false)`. Callers that map `false` to `GenerationUnavailable` / lane-unavailable will tell operators "try later / failed" even though the retained worker was just kicked.

Is the contract "at most one slice, then background owns it, request still succeeds with partial clone coverage"? If so, prefer an explicit Warming/Pending outcome over `Ok(false)` that collapses into the unavailable path.

ScriptedAlchemy (commented):
## rspack-team-mode — tip `7ed395cd04e9` (new head)

Outcome: prior tip holds look **cleared** on this head — `request_query_background_reconcile` now fail-closes on `PublicationAuthorityCorrupt`, journey pins `CANDIDATE_SOURCES_PRUNED`, and clone warmup uses `CloneSimilarityWarmupForRequestV1` instead of `Ok(false)`. **COMMENT** (ready for APPROVE once CI is green — not undrafting here).

Not duplicating standing Codex [P1] `lock_shared` / [P2] `drive_retained`.

### New notes only
Inline: journey transport retry still keys on English `"warming in the background"` / `"retired before response completion"` substrings.

ScriptedAlchemy (commented):
## rspack-team-mode — tip `3f8494dece58`

Outcome: prior tip `7ed395cd04e9` [P2] journey English warming substring looks **cleared** — `PROJECT_WARMING_REASON_CODE` + `error_is_project_warming` key the typed route; journey asserts English prose alone must not decide retry.

Seat-before-clone-backfill / `exact_and_lexical_ready_for_graph` shape still looks right for #1103 / #1226.

**COMMENT** (ready PR; no APPROVE this pass).

PR #1396, opened 2026-09-16
perf(query): batch clone table writes, skip redundant reads

ScriptedAlchemy:
## Summary

- Batch `clone_body_payloads` / `clone_occurrences` / `clone_exact_postings` inserts through `MultiRowInsertV1` (same pattern as term/ngram writers).
- Replace per-body read-after-write SELECTs with in-batch digest checks plus batched `IN (...)` verification for cross-batch conflicts.
- Fingerprint staging, V16 schema, triggers, and readers untouched.

## Motivation

Envelope hotpath folds clone-table write amp into `query.artifact.batch.sqlite` (~9–10 s at r4/r8 on tip+#1377). Row-at-a-time inserts + one SELECT per body were the remaining unstaged clone writers after fingerprint staging. Campaign plan: `docs/perf-triple-plan.md` complementary B4.

## Changes

- `builder.rs`: `append_prepared_clone_bodies` + `MultiRowInsertV1::new_with_conflict_clause` + `verify_clone_payload_digests`.

## Test plan

- [x] `cargo test -p tracedecay-query --lib` (255/255)
- [x] Existing `v16_clone_payloads_are_content_addressed_and_postings_page` passes
- [ ] Release `--clone-envelope` ingest before/after (structural win; wall re-measure owed)
- [ ] CI on this PR

<div><a href="https://cursor.com/agents/bc-e2645d18-e0bc-5954-8084-4baad2df9b7c?cursor_ref=pr_footer&cursor_cta=open_in_web"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a>&nbsp;<a href="https://cursor.com/background-agent?bcId=bc-e2645d18-e0bc-5954-8084-4baad2df9b7c&cursor_ref=pr_footer&cursor_cta=open_in_cursor"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;</div>

---

## Coordinator hold — Fable review [f6d91bcf](f6d91bcf-1adb-445a-9980-e69751c1e496)

Core **PASS**, but keep **draft** until all of the following land on a new head, then independent rereview + CI, then merge. Never merge #707.

1. **Direct collision-branch tests** (real schema/triggers): same-batch same digest/different bytes → `Corrupt`; poisoned on-disk digest before next append → `Corrupt` **and** clone occurrence batch rolled back.
2. **Closed typed conflict policy**: replace private raw `conflict_clause: &'static str` suffix with a typed conflict policy/target that can only render append-only `ON CONFLICT(payload_digest) DO NOTHING`; no generic SQL escape hatch.
3. **Docs + non-vacuous integration**: correct inaccurate docs; run the actual `search_quality_suite` exact v16 clone payload test (not only `--lib`).
4. **Perf evidence** before/after baseline vs head at r4+r8, ≥3 runs each: `clone_bodies` span, `batch.sqlite`, ingest wall; measured SQL statement count vs expected collapse; deterministic section digests and table counts identical. **No merge without material measured win.**
5. **Optional** sibling `clone_successor` row-at-a-time writer is separate; do not widen unless the same helper deletes code cleanly and gets its own evidence.


crates/tracedecay-query/src/retrieval/lexical/projection/artifact/builder.rs
```diff
@@ -3751,65 +3776,123 @@ fn sql_blob(value: &[u8]) -> ToSqlOutput<'_> {
...
+        }
+    }
+    let digests: Vec<&str> = staged.keys().copied().collect();
+    for chunk in digests.chunks(PAYLOAD_DIGEST_CONFLICT_CHECK_CHUNK) {
+        checkpoint(control)?;
+        let placeholders = std::iter::repeat_n("?", chunk.len())
+            .collect::<Vec<_>>()
+            .join(", ");
+        let sql = format!(
+            "SELECT payload_digest, payload FROM clone_body_payloads WHERE payload_digest IN ({placeholders})"
```
ScriptedAlchemy, line 3784: soft / hot-path alloc — dynamic `IN ({placeholders})` SQL per chunk.

`prepare_cached` keys on the full SQL string, so every distinct chunk length is a new statement cache entry (and a `format!` + `join` alloc). Prefer a fixed-width chunk (always 512 placeholders, bind NULLs / shorten the bound slice) or a small set of ladder sizes so the cache stays warm across publishes.

ScriptedAlchemy (commented):
## rspack-team-mode — tip `fb2cbfe2b3d3` (draft)

Outcome: batching clone payload/occurrence/posting inserts and replacing per-row `SELECT` with chunked `IN (...)` conflict checks is the right lexical hot-path move. Draft → **COMMENT only**.

Inline: `format!` SQL per chunk still bypasses a stable `prepare_cached` key shape for variable-length `IN` lists.

ScriptedAlchemy (commented):
## rspack-team-mode — tip `da2d1d3b1ed6` (draft → COMMENT / tip-status)

Tip rewritten; batching clone payload/occurrence/posting inserts + chunked digest conflict checks still the right lexical-artifact shape.

**Prior soft still standing (not re-inlined):** dynamic `IN ({placeholders})` SQL per chunk still keys `prepare_cached` on the full string — prefer a fixed-width chunk (or `rarray` / bound table) so the statement cache hits.

Draft → **COMMENT only**.

PR #1445, opened 2026-09-16
perf(clones): seal reused complement without storing rows

ScriptedAlchemy:
## Why

Issue #1280 still fails the one-second body-refresh bound under full-generation publication. PR #1441 omitted reused projection receipt rows. This follow-up is the next parent-delta unit: `ChangedCodeChunkSetV1` authenticates unchanged chunks with `reused_count` and `reused_digest` instead of storing the complement list (plan 40 §6).

## Scope

- Replace `ChangedCodeChunkSetV1.reused: Vec` with `reused_count` + `reused_digest`.
- Seal the complement in `plan_chunk_increment` and bump the changeset digest domain to `tracedecay.changed-code-chunks.v2`.
- Pre-expand projection-profile changes at request construction; publication refuses a lingering reused complement.
- Projection sinks and metrics stop iterating a reused vector.
- Refresh partitioned codec seal digests for the smaller evidence segment.

Out of scope: parent-authenticated `full_replay_digest`, affected-only assemble/materialize, and flat published-generation redesign.

## Tradeoffs

This slice shrinks request and sealed evidence payloads. It does not remove O(generation) assemble or project_publication work, so the 1 s tip bound still needs the remaining parent-referenced generation cutover.

## Blast Radius

Projection request digests and partitioned fixture seals change. Callers that read `changes.reused` must use `reused_count` / `reused_digest`. Profile-change requests must arrive pre-expanded.

## Verification

- `cargo test -p tracedecay-domain --lib code_intelligence::search` — 14 passed
- `cargo test -p tracedecay-code-index --lib receipts` — 7 passed
- `code_index_suite` `projection_receipts` — 6 passed
- `code_index_suite` `chunk_incremental` — 4 passed
- exact `production_orchestration::partitioned_codec_has_stable_bytes_and_round_trips` — passed; evidence segment **8949 → 6837 bytes** (−2112)
- `tracedecay-index-bench --clone-envelope` (release, production+hotpath, default 260-file corpus): `body_refresh.wall_micros=104069` (~104 ms; #1441 reported 164307), `payloads_computed=1`, `payloads_reused=2808`, peak RSS ~919 MiB
- `scripts/check-clone-envelope-receipt.py --receipt benchmark_data/index-bench/clone-envelope-20260915.json` — exit 1 (stale `harness_commit`, failing 60 s / 1 s targets)

Full 5k / 101k tip remeasure and remaining parent-delta authorities stay open on #1280.

PR #1446, opened 2026-09-16
perf(clones): seal full_replay as parent-delta digest

ScriptedAlchemy:
## Why

Issue #1280 still fails the one-second body-refresh bound under full-generation publication. PR #1445 sealed the reused complement as count+digest. This follow-up is the next parent-delta unit: `CodeGenerationSourceCommitmentsV1.full_replay_digest` authenticates source as a parent-delta over the changed-set partitions instead of hashing the full ordered chunk corpus.

## Scope

- Replace corpus-wide `code_source_full_replay_digest` with `code_source_parent_delta_digest`.
- Add `parent_full_replay_digest` on `CodeGenerationSourceCommitmentsV1`.
- Assemble seals commitments from the active parent's full-replay digest plus `ChangedCodeChunkSetV1` without allocating the full `(chunk_id, content_digest)` vec.
- `validate_for_changes` recomputes the parent-delta from the projection changed-set.
- Refresh partitioned codec seal digests for the new commitments wire shape.

Out of scope: affected-only assemble/materialize, flat published-generation redesign, and a HEAD-stamped 5k/101k receipt.

## Tradeoffs

This removes O(generation) full-corpus hashing from assemble and validate. It does not remove O(generation) rematerialize or project_publication, so the 1 s tip bound still needs the remaining parent-referenced generation cutover.

## Blast Radius

Source-commitment digests and partitioned fixture seals change. Callers that assumed `full_replay_digest` was a generation-independent corpus hash must read `parent_full_replay_digest` for the parent binding. Successor noop publishes no longer copy the parent's full-replay digest.

## Verification

- `cargo test -p tracedecay-domain --lib code_intelligence::search` — 14 passed
- exact `production_orchestration::production_owner_publishes_complete_generation_and_restores_it_after_restart` — passed
- exact `production_orchestration::partitioned_codec_has_stable_bytes_and_round_trips` — passed; evidence segment stays **6837 bytes** (digest refreshed for the new field)
- exact `production_orchestration::sealed_manifest_authenticates_source_commitments_and_refuses_missing_history` — passed
- `tracedecay-index-bench --clone-envelope` (release, production+hotpath, default 260-file corpus): `body_refresh.wall_micros=111592` (~112 ms; #1445 reported 104069), `payloads_computed=1`, `payloads_reused=2808`, peak RSS ~916 MiB
- `scripts/check-clone-envelope-receipt.py --receipt benchmark_data/index-bench/clone-envelope-20260915.json` — exit 1 (stale `harness_commit`, failing 60 s / 1 s targets)

Full 5k / 101k tip remeasure and remaining parent-delta authorities stay open on #1280.


crates/tracedecay-domain/src/code_intelligence/search.rs
```diff
@@ -835,65 +835,109 @@ pub fn code_reused_partition_digest(
...
-/// same across generation-id churn.
+/// and all three change partitions. The full-replay digest authenticates the
+/// ordered source as a parent-delta: genesis seals `parent_full_replay_digest =
+/// None` over the changed-set partitions alone; successors bind the parent's
+/// full-replay digest plus the affected partitions and reused complement seal.
 #[derive(Clone, Debug, Serialize, Deserialize, PartialEq, Eq)]
 #[serde(deny_unknown_fields)]
 pub struct CodeGenerationSourceCommitmentsV1 {
     pub incremental_manifest_digest: ManifestDigest,
     pub full_replay_digest: ManifestDigest,
```
ScriptedAlchemy, line 849: soft / naming — `full_replay_digest` now seals a parent-delta (`CODE_SOURCE_PARENT_DELTA_DIGEST_DOMAIN`), not a generation-independent corpus hash.

Keeping the old field name invites the previous reading (“byte-identical source ⇒ same digest across generation-id churn”), which this PR deliberately breaks (and the orchestration test now asserts `assert_ne!` across successors). Prefer `parent_delta_digest` (with a one-release serde alias if anything external already keyed the old name), or a short invariant comment at the field that the name is historical.

ScriptedAlchemy (commented):
## rspack-team-mode — tip `16ca2434d1c7`

Outcome: parent-delta full-replay (`tracedecay.code-source-parent-delta.v1`) with `parent_full_replay_digest` + genesis/`Some` binding checks is the right next seal after #1445 — stops hashing the full corpus on every generation. Partitioned codec goldens refreshed with the domain change (explained).

### Soft
- Wire field is still named `full_replay_digest` while the sealed domain is parent-delta. Prefer renaming (or a doc+alias plan) so consumers do not keep the old “generation-independent corpus hash” reading.

**COMMENT**.

crates/tracedecay-domain/src/code_intelligence/search.rs
ScriptedAlchemy: Kept `full_replay_digest` (stacked #1447/#1450); added invariant that it seals `tracedecay.code-source-parent-delta.v1` and the name is historical. 3363215081

PR #1449, opened 2026-09-16
fix(code-index): yield background-cpu units around chunk fan-outs

ScriptedAlchemy:
## Summary

- Fix a process-wide deadlock in background-CPU admission: a pool worker that fanned chunk work out while holding its own unit could wedge every other admission in the process once a full-width request reached the FIFO head.
- Pin it with a deterministic regression test that fails on the base through a bounded timeout (reporting the admission state) and passes with the fix. The test signals from inside admission (`ProcessBackgroundCpuV1::with_permits_placed`), not from shared counters, so sibling tests in the same process cannot release its gates.
- Prerequisite for #1443, which does not cause the cycle but raises its hit rate.

## Motivation

Verifying #1443 with `cargo test -p tracedecay-code-index-runtime --profile perf --lib -- clone_successor serving` parked the whole test binary at 0.5 % CPU in 4 of 6 runs; gdb showed every thread in `ProcessBackgroundCpuV1::admit_units`. The cycle is in code that lane never touched; it only changed the timing enough to hit it often. The same cycle can occur in production wherever a lexical sorter admission (`with_builder_sorter_cpu_admission`, `indexing_workers() + 1` units clamped to the width) queues while a `collect_bounded_ordered` leaf is mid fan-out.

The gdb-evidenced cycle, in three lines:

1. A rayon worker holds 1 unit inside a `collect_bounded_ordered` leaf (`with_background_cpu_permit`) and, inside it, `chunks::admit_all` runs a nested `par_iter` whose stolen halves call `with_permits` on other workers and queue in `admit_units`; the parent blocks on the join latch still holding its unit.
2. The FIFO head is `with_builder_sorter_cpu_admission` asking for `PRAGMA threads + 1` = width units, which can never fit while that unit is held.
3. `admit_units` is strict FIFO, so the nested halves behind the head are never admitted, and the parent never releases: cycle closed.

## Changes

- `crates/tracedecay-code-index/src/parallelism.rs`: new `with_yielded_background_cpu_permits`, the standalone-safe counterpart of `with_background_cpu_permit` that runs `ProcessBackgroundCpuV1::with_yielded_permits` on the installed authority (no-op without one), the same primitive `parallelism::install` already uses. `install()` docs name it as the sibling primitive for bare `par_iter` fan-outs under a held leaf permit.
- `crates/tracedecay-code-index/src/chunks.rs`: `try_for_each_chunk_ordered` (serves `validate_all` and `rematerialize_for_generation`) and `admit_all` run their `par_iter` inside that yield, so no thread waits on rayon while holding admission. Leaves are still admitted one unit each; the parent's own leaves now acquire like stolen ones instead of reusing the parent's unit.
- `crates/tracedecay-runtime-core/.../background_cpu.rs`: `ProcessBackgroundCpuV1::with_permits_placed(units, placed, op)` — `placed` runs under the state lock right after the request is pushed onto the FIFO (or immediately when a sufficient admission is reused on the calling thread). `with_permits` delegates to it with a no-op; admission logic is unchanged. Pinned by `placed_fires_once_a_blocked_request_holds_its_queue_position` (fires while `waiting=2, active=1`, before admission).
- No change to FIFO policy: with the yield, no holder ever blocks on a request behind it, so head-of-line admission needs no bypass (a bypass would starve wide waiters). Width, budgets, and timeouts are unchanged.
- Regression test `chunks::tests::nested_chunk_fan_out_does_not_wedge_behind_a_full_width_head_waiter`: installs the automatic worker plan, admits a holder on a pool worker, queues a full-width head request that signals via `with_permits_placed` once it holds its queue position, then fans out; leaves are admitted through the injected `admit` closure with the same signal, and the first leaf stays busy until a sibling leaf has taken its own position behind the head. No `waiting_work_units()` reads remain in the control flow. Fails by `recv_timeout` after 10 s with `active_units`/`waiting_work_units`/`leaves_placed`/`leaves_entered` in the message. Skips with a printed reason when the installed width is 1 (1 logical CPU or `TRACEDECAY_INDEX_WORKERS=1`): the scenario needs a second pool worker to steal onto.

Inferred, not traced: other `par_iter` sites (`production/helpers.rs`, `incremental.rs`, `graph_projection/builder.rs`) run top-level under `install`, which already yields; the independent change-risk review enumerated all fan-outs in `code-index`/`code-index-runtime` and found none outside `install` or the new wrapper.

## Test plan

Measured on a shared 4-core VM, `--profile perf`, `CARGO_INCREMENTAL=0`:

| What | Before | After |
| --- | --- | --- |
| `cargo test -p tracedecay-code-index --lib` (whole lib, siblings present) | with `f81df3b34` reverted: `nested_chunk_fan_out… FAILED` at 10 s, `active_units=1 waiting_work_units=11 leaves_placed=63 leaves_entered=60`; 4 sibling tests then park, binary hits outer timeout | 233/233 ok ×3 (4.7 s each) |
| `-p tracedecay-code-index-runtime --lib -- clone_successor serving` on `agent/1280-increment-cost` (`c58fefc46`) | 2 of 6 runs completed, 4 hung (killed at 10–25 min) | 5 of 5 completed (96–119 s); 2 runs hit the pre-existing flake `unchanged_reconcile_does_not_reactivate_the_serving_generation` |
| same filter on merge-base `82cb66a2e` | 3 of 3 completed (one pre-existing flake) | n/a |

`leaves_placed=63 leaves_entered=60` is the wedge itself: the holder ran 60 leaves on its own thread; three stolen leaves placed themselves behind the full-width head and were never admitted while the holder kept its unit.

- [x] `cargo test -p tracedecay-runtime-core --profile perf --lib -- background_cpu`: 6/6 incl. the new hook test
- [x] `cargo test -p tracedecay-code-index --profile perf`: all integration targets green
- [x] `cargo clippy -p tracedecay-code-index -p tracedecay-runtime-core --profile perf --no-deps --all-targets -- -D warnings` clean
- [x] `cargo fmt --check`: touched files clean (pre-existing base diffs in untouched files)
- [x] `npm run lint:commit -- --from origin/codex/tracedecay-total-redesign-plan-reopened --to HEAD` passes
- [ ] `cargo nextest run --workspace --no-fail-fast` — CI

## Checklist

- [ ] `CHANGELOG.md` — release-please generated
- [x] Breaking changes documented — none; `with_permits_placed` is additive, `with_permits` behaviour unchanged

Review threads: shared-counter gate and width-1 assert (Codex P2 ×2) → `89184afd7`, superseding the delta/`Barrier` approach in `109641f0d` (which still read the shared counter).

<div><a href="https://cursor.com/agents/bc-040db512-4012-5b29-bead-9917ce86d94a?cursor_ref=pr_footer&cursor_cta=open_in_web"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a>&nbsp;<a href="https://cursor.com/background-agent?bcId=bc-040db512-4012-5b29-bead-9917ce86d94a&cursor_ref=pr_footer&cursor_cta=open_in_cursor"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;</div>


ScriptedAlchemy (commented):
## rspack-team-mode — tip `89184afd719b`

Outcome: yielding the caller's background-CPU units across nested `par_iter` fan-outs (`with_yielded_background_cpu_permits`) is the right fix for the FIFO wedge behind a full-width head waiter. `with_permits_placed` + the nested chunk admission test pin the production shape without shared-counter races.

**COMMENT**.

PR #1453, opened 2026-09-16
fix(session-runtime): refuse RequireFresh without a refresh worker

ScriptedAlchemy:
## Why

`DaemonSessionRetrievalService` carried its refresh worker as `refresh_status: Option<Arc<dyn SessionProjectionServingStatusPort>>`, and `refresh_not_current()` returned `None` through `as_deref()?` when no worker was mounted. A `RequireFresh` query on such a service was therefore never refused for staleness: absence of the worker read as "current". Every profile-scope service (`new_admitted_profile`, used by retained message search with `catch_up: true` and by the retained LCM route) and the dashboard LCM adapter (`dashboard.rs:260`) were constructed that way (first-principles audit, report 11 finding 2).

The serving vocabulary already had the right word for this state: `SessionProjectionUnavailableReason::WorkerMissing` → `SessionRetrievalUnavailableReason::RefreshWorkerMissing`. It was never reachable because the missing case was an `Option`, not a state.

## What

- `tracedecay_sessions::serving::RefreshWorkerMissing`: a `SessionProjectionServingStatusPort` whose status is `Unavailable { WorkerMissing }` with `blocker: WorkerMissing`.
- `refresh_status` is now a required `Arc<dyn SessionProjectionServingStatusPort>`. Constructors say what they mean: `new(database, root, wake)`, `new_without_refresh_worker(database, root)`, `new_with_serving_port(database, root, port)`. `new_admitted_profile` and the dashboard adapter use `new_without_refresh_worker`; the MCP composition maps its (still `Option`) `project_session_refresh_serving` to `RefreshWorkerMissing` for direct/core servers.
- `projection_serving_status()` always reports a status.

**Behavior change, stated plainly:** a `RequireFresh` query against a service with no wired worker now answers `Unavailable(RefreshWorkerMissing)` instead of a page served as fresh. `AllowStored` reads are unchanged. Wiring the profile refresh wake into the retained profile constructors is a separate follow-up; this PR makes the state truthful first.

## Evidence

Commit order proves the change (`test` first, then `fix`):

1. `3ec3deb` adds `session_retrieval::tests::require_fresh_without_a_refresh_worker_is_refused_as_worker_missing`, written against the pre-change API. **On the base it fails at runtime**: `RequireFresh without a worker must be refused, got Partial { page: SessionRetrievalPageView { results: [ ... ] } }` — the query was served.
2. `c6a217a` lands the port and constructors. `cargo test -p tracedecay-session-runtime --lib session_retrieval`: 31 passed, 0 failed (including the new test). `cargo clippy -p tracedecay-sessions -p tracedecay-session-runtime -p tracedecay-daemon-service --all-targets -- -D warnings`: clean. `cargo check -p tracedecay --lib --tests`: clean (covers `dashboard.rs`, `mcp/server.rs`, and the three daemon test modules that constructed the service). `cargo fmt --check`: clean.

`ripwire --quality-delta=d2c9294089..HEAD`: remaining rows are sub-30-token constructor and struct-literal idioms plus two false dead-code rows (a trait impl and a test); no real duplication added.

## Not changed here

`McpServer.project_session_refresh_serving` is still an `Option` at the composition level (report 02 finding 1, the 55-`Option` server); this PR converts it at the one consumer. The `SessionApplicationRetrievalPortV1::projection_serving_status` trait still returns `Option` for other implementors.

Refs the PR #707 first-principles audit (report 11 finding 2; root cause R2, absence-as-a-variant).

<div><a href="https://cursor.com/agents/bc-ec03ae0c-6cd7-5c15-b223-14cf23b12954?cursor_ref=pr_footer&cursor_cta=open_in_web"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a>&nbsp;<a href="https://cursor.com/background-agent?bcId=bc-ec03ae0c-6cd7-5c15-b223-14cf23b12954&cursor_ref=pr_footer&cursor_cta=open_in_cursor"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;</div>


ScriptedAlchemy (commented):
## rspack-team-mode — tip `e26fa0c8d8be`

Outcome: mounting `RefreshWorkerMissing` when no refresh worker exists (instead of `Option` → “current”) is the right fail-closed fix for `RequireFresh` on profile-scope / LCM adapters. `new_admitted_profile` / `new_without_refresh_worker` + serving-status mapping keep each refusal reason on its own gauge.

**COMMENT**.

PR #1435, opened 2026-09-16
fix(code-index): type NonCanonical causes with reason codes

ScriptedAlchemy:
## Summary
Closes #707 review threads on stringly `NonCanonicalIdentity` dumps (r4023750206) and shares typed taxonomy with chunk increment + profile-shard NonCanonical (r4023750214, r4023750223).

- `NonCanonicalCauseV1` with stable `reason_code` + structured detail keys
- `ChunkingFailureV1` / `ChunkIncrementErrorV1` carry the typed cause
- Profile-shard validation uses closed `ProfileShardNonCanonicalReasonV1`

Rebased onto tip `82cb66a2e511` (`727ed7dbf1aa`).

## Evidence
- `cargo test -p tracedecay-code-index --lib admission_falls_back`
- `cargo test -p tracedecay-code-index --test code_index_suite import_evidence` (11)

Do not merge #707. Campaign base only. FastEmbed stays retired.

<div><a href="https://cursor.com/agents/bc-d2e8ee44-807b-5d2c-ba37-4500ee50d70e?cursor_ref=pr_footer&cursor_cta=open_in_web"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a>&nbsp;<a href="https://cursor.com/background-agent?bcId=bc-d2e8ee44-807b-5d2c-ba37-4500ee50d70e&cursor_ref=pr_footer&cursor_cta=open_in_cursor"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;</div>

PR #1476, opened 2026-09-16
test(code-index): isolate forced worker width

ScriptedAlchemy:
## Summary
- scope the test-only forced indexing width to its invoking test thread
- prevent concurrent code-index unit tests from changing each other's window assertions

## Verification
- code-index, code-extraction, and code-index-retention lib/test sweep passed via cargo-hauler


ScriptedAlchemy:
Superseded by #1435: the forced pool failure/worker-width controls are thread-scoped on tip (`5e82865867`), and the full code-index lib suite passed 243/243.

PR #1386, opened 2026-09-16
perf(code-index-runtime): batch segment dir fsync per publish

ScriptedAlchemy:
## Summary

- Batch `segments_root` directory fsync to once per publish when any new file segment was written (was once per segment).
- Keep per-segment file fsync-before-rename; durability contract unchanged.
- Complementary to #1377 seating and Zack chunker-v5 (untouched).

## Motivation

Lifecycle hotpath: `segment_durable` ~3.3–3.7 s serial on the 772-file restart publish — N directory fsyncs under the exclusive writer lock. Part of campaign ~3× plan (`docs/perf-triple-plan.md` in project store).

## Changes

- `publication_store.rs`: `publish_segment_durable` returns whether a new rename occurred; one `segments_dir_sync` after encode.
- Test: `publishing_many_new_segments_syncs_the_segments_directory_once` (≥8 segments → exactly 1 dir sync).

## Test plan

- [x] `cargo test -p tracedecay-code-index-runtime --features hotpath publication_store` (25/26; one pre-existing unrelated base failure)
- [x] clippy/check on crate with `--features hotpath` clean
- [ ] CI on this PR

<div><a href="https://cursor.com/agents/bc-e2645d18-e0bc-5954-8084-4baad2df9b7c?cursor_ref=pr_footer&cursor_cta=open_in_web"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a>&nbsp;<a href="https://cursor.com/background-agent?bcId=bc-e2645d18-e0bc-5954-8084-4baad2df9b7c&cursor_ref=pr_footer&cursor_cta=open_in_cursor"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;</div>

---

## Coordinator hold — durability and falsifiability

Do not mark ready or merge until [independent review](https://github.com/ScriptedAlchemy/tracedecay/pull/1386) findings are fixed:

[...]

PR #1479, opened 2026-09-16
fix(code-index-runtime): restore shared retained reads

ScriptedAlchemy:
## Summary
- restore the shared generation-store read lock for retained lexical segments
- preserve typed cancellation when retention removes a generation
- repair the hunk lost when PR #1386 merged into the integration train

## Evidence
- base ref test: passes (old non-blocking-reader contract)
- train-tip pre-fix test: fails at the shared-reader parking assertion
- fixed focused test: 5 consecutive passes, plus a post-commit pass
- `ripwire . --quality-delta --legend=compact`: no gating regressions
- `ripwire . --edit-check=read_retained_partitioned_segment`: contract unchanged

## Full suite
`cargo test -p tracedecay-code-index-runtime --lib` reached 474 tests and this regression test passed, but the train has unrelated known-red tests owned by active repair lanes; the run later stalled in `foreign_wake_keeps_pending_arrival_when_query_claim_is_released` and its explicitly stalled cargo-hauler ticket was stopped. No unrelated test was changed or masked.


ScriptedAlchemy:
Superseded by #1458: retained segment reads use the shared lock on tip (`8c78bddae63f`), and the wait/resume regression passed locally.

PR #1562, opened 2026-09-17
fix(code-index): seat text through retryable graph activation

ScriptedAlchemy:
## Summary

- Seats the sealed text generation when native graph activation fails retryably, instead of erasing the prepared serving candidate.
- Does not treat an unfinished clone-fingerprint successor as `published_text_owner_unfinished` once exact and lexical owners are ready.
- Leaves the 90s receipt timeout and the swallowed missing `projects/proj_*` parent alone.
- Investigation only after `b80dd58`, including a mid-run macOS sample (pid 11360, ~26s into the journey). No pipeline change in this PR. Do not merge. Do not raise the timeout.

Draft. Based on current `origin/master` (`dec90f0292`), which already contains #707 and diagnosis PR #1557.

## Seat-swap race

#1557 showed both 90s graph-rebuild attempts failing because search kept serving the predecessor while the replacement text owner and verified graph had already moved. A publish installs that text owner before the serving swap. Two arms then threw the prepared candidate away: retryable graph activation rewrote the pass result to `Ok((Err, None, None))` (`mount.rs`), so the swap never ran for the whole backoff; and an `Unfinished` text projection cleared the same candidate even when the only remaining work was the clone-fingerprint successor. Search therefore stayed on generation N while status and the graph named N+1 (`verified-code-graph-generation-mismatch`).

## Changes

- `serving_generation_after_activation_failure` / `activation_failure_keeps_serving_candidate`: retryable activation still schedules the graph retry and clears `graph_seat_attempted`, but does not wipe the text candidate. The existing swap then installs that generation.
- `text_projection_unfinished_withholds_seat`: `Unfinished` plus ready exact/lexical owners is reclassified as `Finished` before the seat decision, so the successor is woken and the candidate stays.

## Test plan

Ran locally:

`REQUIRE_EXACT_TEST_COUNT=4 scripts/require-exact-test.sh cargo test -p tracedecay-code-index-runtime --lib -- retryable_activation_keeps_the_serving_generation_matched unfinished_clone_fingerprint_successor_is_not_text_projection_unfinished serving_swap_seats_a_generation_whose_publication_moved_while_it_activated graph_rebuild_split_is_not_terminal_freshness`

**Result: 4 passed, 0 failed** (`Finished test profile in 3m 22s`).

- `retryable_activation_keeps_the_serving_generation_matched` — pass
- `unfinished_clone_fingerprint_successor_is_not_text_projection_unfinished` — pass
- `serving_swap_seats_a_generation_whose_publication_moved_while_it_activated` — pass
- `graph_rebuild_split_is_not_terminal_freshness` (#1557 predicate) — pass

The 90s `background_refresh_and_reopen_report_only_servable_generations` journey was not re-run here. Local follow-up: seat matched, search works, clone `2305/2305` completed inside the window, `edge_count` 0, wait failed at ~93.7s. Timeout was not raised.

## Clone-page investigation (`b80dd58`)

The receipt (`graph_rebuild_status_test.rs:26`, `RECEIPT_TIMEOUT = 90s`) needs `code_index_freshness.status == current`. That is text ready **and** graph `Ready` **and** the serving seat matching the advertised generation (`registry.rs:1166-1176`, `1205-1231`). Clone backfill is not an input. `edge_count` 0 with a matched seat means graph never published.

The fixture is `768` files × `128` functions (`graph_rebuild_status_test.rs:232-240`) = **98,304 bodies**. Observed **2305 pages** is 3 pages per generated file plus `src/lib.rs`. Each body is `input + n`, under the 30-token clone floor (`MIN_AUTOMATIC_CLONE_BODY_TOKENS_V1`, `clone_body.rs:13`), so this journey writes **no fingerprint rows**.

### Sample at ~26s (pid 11360)

Hottest worker stacks are not SQLite. They are:

1. `serde` deserialize of `CodeIndexCloneBodyV1` / `CloneBodyPayloadV1` / `ConservativeCloneTokenV1`
2. `clone_payload_digests`
3. `canonical_json::write_value`
4. `admit_validated_file_parts`
5. privacy redaction walks

That is the sealed-file admit that builds every lexical page, including the V14 walk **before** the seat and the clone-successor walk that produces `2305/2305`. An older comment measured segment decode at ~4 ms/file, 3.4 s of a 770-file build, when decode ran on one thread (`partitioned_codec.rs:2712-2718`). Format V3 exists because segments now carry clone-body rows (`partitioned_codec.rs:1351-1354`). The sample says those rows are the CPU at t≈26s, so that 3.4 s is a floor, not the current cost. The successor pays the same admit again.

### Path the sample is in

Per file, on the indexing pool (`decode_segment_window`, `partitioned_codec.rs:2720-2736`, fanned out by `collect_bounded_ordered`):

1. Manifest digest check (`partitioned_codec.rs:1318-1327`).
2. `serde_json::from_slice` of the raw segment (`partitioned_codec.rs:1343-1349`).
3. `canonicalize_json_into` → `write_value` rewrites the whole file JSON, including every clone token (`canonical_json.rs:78-91`, `386-404`; call at `partitioned_codec.rs:1364-1367`).
4. A second `serde_json::from_slice` expands `PersistedFileGenerationArtifactsV2` into `CodeIndexCloneBodyV1` (`partitioned_codec.rs:1384-1388`, `sealed_codec.rs:366`).
5. `admit_validated_file_parts` (`lexical_page_source.rs:3321`) calls `artifacts.validate()` (`3331`), which is `validate_with_clone_payloads(true)` (`artifacts.rs:322-324`). That calls `CloneBodyPayloadV1::validate` (`clones.rs:515-528`) → `clone_payload_digests` (`clones.rs:292-353`): three canonical hashes of the token list (body, conservative, payload), four when rename is complete. Extraction already computed those digests (`clones.rs:412-426`).
6. The same function clones every body and `serialize_page_row`s it again (`lexical_page_source.rs:3368`, `3409-3418`).

The admitted window holds only `workers * 4` files and **drops** the previous window (`lexical_page_source.rs:2211-2216`). After the V14 walk, publish drops the source (`serving.rs:3177-3178`) and the successor calls `open_sealed_source` (`serving.rs:3213`). The second 2305-page walk re-decodes all 768 segments. `observe_clone_page_scratch` then JSON-encodes each body a third time (`serving.rs:3357-3368`) before the per-body SQLite loop.

Privacy is a different stack, on capture, not on admit: `git_tree_capture.rs:398-399` → `sanitize_code_file` (`privacy.rs:27-55`). Rust takes `CodeOrProse`, which copies the file and runs every credential regex (`structured_text.rs:269-277`, `624-632`; `detect.rs:704-714`). The same function then wraps the whole file as `serde_json::Value::String` and `verify_sanitized_json_payload` hashes that copy again (`privacy.rs:50-54`, `detect.rs:574-593`). It is inside the 90s, ahead of page admit. It is not the clone successor.

### What the earlier SQL note still is, and is not

`append_clone_rows` (`clone_successor.rs:433-496`) still does one prepared-each-time `INSERT` / `SELECT` / `INSERT` per body into a hash-keyed `WITHOUT ROWID` table, 37 commits (`serving.rs:135-173`). V16 already batches that (`builder.rs:3846-3901`) and stages fingerprints (`builder.rs:3916-3921`, measured 2.1 GiB → 0.4 GiB). **The t≈26s sample does not show that path.** On this fixture fingerprint inserts are a no-op. Do the batch writer in the follow-up so the successor cannot fall behind V16, but do not expect it to move the stacks the sample recorded.

`2305/2305` is `next_cursor` before `finish` and the full-file hash (`serving.rs:1727-1735`, `clone_successor.rs:193-275`). The retained driver is joined before the next graph retry (`mount.rs:2221-2272`) and holds a 128 MiB reservation the whole time (`serving.rs:2544-2548`). That scheduling still matters after admit gets cheaper. It is not what is hot at 26s.

### What not to do

- Do not raise `RECEIPT_TIMEOUT`.
- Do not raise the work-per-wake hint. It is already 1024; clone commits are capped at 64 pages by the memory ledger.
- Do not parallelize page admit. Decode is already a window fan-out (`partitioned_codec.rs:2712-2718`). The cursor and the SQLite mutation gate stay single-flight.
- Do not drop privacy scanning. The redundant part is the second hash, not the credential scan.

### Follow-up, in sample order

**1. Stop re-hashing clone payloads on admit.** In `admit_validated_file_parts` (`lexical_page_source.rs:3331-3338`) call the existing `validate_reusing_clone_payloads` path (`artifacts.rs:326-328`, `validate_with_clone_payloads(false)`) instead of `validate()`. Segment bytes are already digest-checked. Digests were sealed at `from_extracted`. This removes `clone_payload_digests` from both the V14 walk and the successor walk: 98,304 bodies × 3–4 canonical hashes × 2 opens. Structural checks (order, span, occurrence binding) stay. This is the smallest patch that matches the sample.

**2. Decode a current V3 segment once.** `decode_verified_file_segment` (`partitioned_codec.rs:1335-1388`) parses the segment, rewrites every string through `canonical_json::write_value`, then deserializes the rewrite into clone bodies. The rewrite restores legacy identity markers. A generation this process just sealed is V3 (`encode_file_segment`, `partitioned_codec.rs:1160-1161`) and has no markers. If a scan of the payload finds none, `from_slice` the verified payload once and skip `canonicalize_json_into`. Fail closed back to the rewrite when a marker is present. Removes the `write_value` stack and one of the two deserializes, on both walks.

**3. Do not re-decode the sealed source for the successor.** `serving.rs:3177-3213` drops the source that just admitted 2305 pages and opens a new one, which repeats (1) and (2) for every file. `rewind` (`lexical_page_source.rs:1683`) is not enough: `ensure_admitted_file` clears the window (`2216`). Keep the decoded clone bodies from the first admit (not the chunk/display copies) and let `begin_clone_successor` / `append_pages` consume that list. 98,304 too-small payloads, not a second 768-file JSON parse. Correctness: same ordered pages, same successor cursor and resume checks (`clone_successor.rs:157-167`). This is the clone-backfill-specific cut. Do it after (1) and (2) so the first walk, which is on the path before the seat, is already cheaper.

**4. Drop the third JSON copy on the successor.** Remove `observe_clone_page_scratch` from the admit callback (`serving.rs:3342-3344`, `3357-3374`). It serializes payload and occurrence only to update a progress peak, while `append_clone_rows` serializes them again. Sub-second to a couple of seconds, not the sample’s main frames.

**5. Batch the successor inserts, including fingerprint staging.** Replace `append_clone_rows` (`clone_successor.rs:433-527`) with `verify_clone_payload_digests` + `MultiRowInsertV1` (`builder.rs:3846-3901`, 32 rows/statement). Create `clone_fingerprint_postings_pages` in `reset_clone_tables` and derive it at `finish`, and point `verify_clone_fingerprint_page_rows` (`clone_successor.rs:658-707`) at the staging table until then. Preserves resume. About 0 fingerprint rows on this fixture; the measured 5× write cut is for eligible corpora. Do it in the same writer patch, after the admit cuts.

**6. Defer the retained successor until graph is `Ready` or activation is terminally refused.** Gate `drive_retained` (`mount.rs:597-611`). The join at `mount.rs:2221-2272` blocks the next graph retry on whatever the successor still has left. Freshness does not read clone progress. Keep the one-slice similar-query warmup (`serving.rs:183-186`). This does not remove the t≈26s stacks if those are the first admit. It keeps the second walk off the graph retry after (3) shrinks it.

**7. Privacy, separate small patch, only if still hot after 1–3.** Do not rescan. In `sanitize_code_file` (`privacy.rs:50-54`), stop wrapping the sanitized file as a JSON string solely so `verify_sanitized_json_payload` can hash it again. The regex walk in `raw_only` stays. Not a clone-writer change.

### #1562 vs follow-up

- **Keep in #1562:** the seat-swap only. The sample is admit CPU, not the seat race.
- **Follow-up:** (1) then (2) then (3). That is the sample. (4) and (5) ride along with the successor writer. (6) after the second decode is gone. (7) only if a new sample still shows redaction.
- **Not in either PR:** a higher 90s timeout, a larger batch hint, parallel page cursors.

### Speedup toward the existing 90s receipt

Need about **4s** past 90, and the sample says the time before `2305/2305` is clone-body decode and re-hash, paid twice.

- (1) removes the digest stack on both opens. Several seconds to low tens, not a guess of the whole 26s.
- (2) removes the rewrite stack and one deserialize per file. Same order, on top of (1), on both opens.
- (3) deletes the second open. If the first admit is a large slice of the 90s, cutting the copy is the difference between finishing graph inside the window and missing by ~4s. It does not shorten capture or the first admit.
- (5) is the wrong lever for this sample and this fixture. Keep it so eligible corpora do not regress.
- (6) helps only the graph retry that today waits on the second walk.
- (7) is capture, before either walk. It cannot explain `admit_validated_file_parts` being hot. Treat it as extra seconds, not the fix.

No new Hotpath run. The 27× fingerprint figure remains the comment at `builder.rs:3916-3921`. The 4 ms/file figure remains the comment at `partitioned_codec.rs:2715-2716` and predates clone-body rows.

## Checklist

- [x] No secrets
- [ ] CHANGELOG not updated (internal scheduler seating; no user-facing contract rename)
- [x] No timeout or budget change

<div><a href="https://cursor.com/agents/bc-60796cf9-f832-5ec2-a287-9762d8b1f5c7?cursor_ref=pr_footer&cursor_cta=open_in_web"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a>&nbsp;<a href="https://cursor.com/background-agent?bcId=bc-60796cf9-f832-5ec2-a287-9762d8b1f5c7&cursor_ref=pr_footer&cursor_cta=open_in_cursor"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;</div>


ScriptedAlchemy:
### Related open work (do not widen this PR)

Sibling drafts for the post-seat receipt miss:

- **#1567** — serving-current latency / status census wakes (`GitMetadataFingerprintV1::capture` + `note_wake_if_idle` on the ~16k status polls).
- **#1568** — ranked plan: keep seat-swap here; cut clone successor off the receipt; then admit digest reuse / skip re-canonicalize / no second sealed open.
- **#1570** — share clone-body tokens via `Arc` + DFA-friendly `\b` for privacy (no file overlap with this PR). Strongly complementary to the Mac sample hot stacks.

Recommendation: **merge #1562 as seat-swap only** when ready; do not fold #1570/#1567/#1568 into this branch. No merge from this comment.

ScriptedAlchemy:
Landed through #1877 (6c9af77734), which removes the retryable arm's result rewrite so a sealed generation seats text while native graph activation retries. The starvation this branch targeted was reproduced on master (72 activation attempts, 0 seats) and is covered by `text_seats_while_graph_activation_keeps_failing_retryably`.

PR #1568, opened 2026-09-17
docs(code-index): plan graph-rebuild receipt speed

ScriptedAlchemy:
Draft. Do not merge. Do not raise the 90s receipt.

The seat-swap commit versus master is #1562 (`b80dd58`). This PR does not change that behavior. Interim pstack reorders the cuts. Full text: `docs/plans/graph-rebuild-receipt-speed.md`.

## Outcome

90 seconds is not too short for the predicate. The window is spent twice. The wait helper busy-polls status (~16k calls) and each poll recaptures a fingerprint. The workers spend CPU on clone-body JSON, canonical rewrite, and token deserialize. Edge projection is not the burn. `edge_count` 0 with 98305 symbols is the no-call fixture. Cut that work and make the readiness check cheap.

`seated_generation_age_seconds` is seal age (`status.rs:296`), not time since the serving swap. At ~89s the generation has existed for almost the whole receipt and line 176 still has not returned. Seat match and search hits are not the predicate.

## Pstack, mapped to tip

Poll. The wait is `yield_now` with no sleep (`graph_rebuild_status_test.rs:171`), on 4 worker threads (`:248`). ~16k full `tracedecay_status` calls fit in 90s. Compact status still builds branch diagnostics before the include flag (`status.rs:331-335`) and opens gix via `current_branch` (`branch.rs:107-117`). The named capture is `GitMetadataFingerprintV1::capture` (`identity.rs:166-176`): gix open, HEAD read, hash of loose refs, used by `ready_without_stat` (`reconcile.rs:540-562`). `dashboard_freshness` must not open git (`serving_reads.rs:243-245`). A readiness poll returns the last ladder.

Workers. Not `collect_edge_evidence`. Every callable is tokenized before the 30-token exclusion (`clone_body.rs:157-185`, exclusion at `:229`). Tokens are cloned into the payload (`clones.rs:413-443`), serde-encoded, and rewritten by `canonicalize_json_into` (`partitioned_codec.rs:1230-1236`, named in `canonical_json.rs:654`). The successor deserializes them again (`clone_census.rs:46-49`, `clone_successor.rs:440-458`).

## Ranked changes

Do not amend #1562. Do not touch graph publication on this evidence.

[...]

PR #1567, opened 2026-09-17
docs(code-index): serving-current latency after seat keep

ScriptedAlchemy:
## Summary

Investigation only. Draft. Do not merge. Does not change the 90s receipt timeout.

Stacks on #1562 (`b80dd58`). The new commits are the note in `docs/code-index-serving-current-latency.md`.

Local receipt: after the seat-candidate keep, the serving seat matches and search can hit inside ~90s, but `wait_for_current_generation` still hits `RECEIPT_TIMEOUT`. Timeout status has `seated_generation_age_seconds` ≈ 89. A mid-run sample puts the test thread in that helper at `graph_rebuild_status_test.rs:174` (the `timeout(...).await` around the loop at `148-172`), inside status/tool MCP calls. The prior fail did ~16k status polls. Search is not that park: it runs only after status already looks current (`153-160`).

Age is seal time (`status.rs:293-296`), so the replacement sealed about a second into the wait. Estimates are structural, not a profiled re-run.

## Wait helper vs production current

The helper (`graph_rebuild_status_test.rs:140-179`) matches production `current` (`status.rs:451-489`): coverage `complete`, staleness `fresh`, graph `Ready`, seat match (`registry.rs:1205-1231`). Search only needs the seated exact/lexical owners. Do not treat seat match plus hits as `current`, and do not raise `RECEIPT_TIMEOUT` (`graph_rebuild_status_test.rs:26`). Do not add a sleep to thin the polls.

## Status poll hits capture and project

`tracedecay_status` always snapshots the generation census (`crates/tracedecay/src/mcp/tools/handlers/dispatch_groups.rs:533-541`, read at `status.rs:257`) even though this test already disables branch diagnostics, storage health, session ingest, and staleness (`graph_rebuild_status_test.rs:103-109`). The census reader is `ProjectCodeGraphServingAuthorityV1::project` (`project_reads.rs:86-139`).

`project` tries the ready-decoded seat first (`serving_reads.rs:965-988`). That calls `ready_without_stat` (`serving_reads.rs:928`) → `GitMetadataFingerprintV1::capture` (`reconcile.rs:540-561`, `identity.rs:167-176`) and, on a miss, `note_wake_if_idle` (`serving_reads.rs:924-930`). The next arm, `retained_text_owner_freshness_for_scope`, captures again (`serving_reads.rs:1291-1308`). Only then does it use the O(1) seat (`serving_reads.rs:1342-1364`), which neither captures nor wakes.

[...]

PR #1570, opened 2026-09-17
perf(index): share clone-body tokens; scan secrets with the DFA

ScriptedAlchemy:
## Why

The installed beta.40 daemon was pegging ~18 cores for 80+ minutes while warming 12 projects and still had not sealed a fresh generation. Attacking the premise that index-worker CPU is productive work:

**Premise written down:** "more index workers and fewer blocking waits seal the warm generation sooner." The bounded-lock work, the parent-delta optimization, and the worker-width sizing all assume it.

**Census** (`perf record -g` on the live daemon, bucketed per `tracedecay-index` thread; rerunnable script in the PR discussion): every one of 19–20 active workers had the same shape — 16–50% kernel spinlock/futex, 10–25% malloc, **3–14% in extraction code**. Even across actors, so nothing assigns a thread a bad role; the cost is a resource every worker shares.

Across cost centres, over a 5 s sample of the index workers:

| share | where |
|---|---|
| 42% | regex `PikeVM::search_imp` / `is_word_unicode` (the sanitizer) |
| 20% | `__pv_queued_spin_lock_slowpath` / futex under `__lll_lock_wait_private` (glibc malloc arena) |
| 16% | `_int_malloc` / `free` |
| 9% | TraceDecay extraction |
| 0.3% | tree-sitter |

Callers of the allocation storm: `<Vec<ConservativeCloneTokenV1> as Clone>::clone` (31% of cycles in a 10 s window) and its inner `to_vec<u8>` `String` clones (36%). Every token owns two `String`s and every payload deep-copied the extracted body's token stream.

The PikeVM appears because gitleaks patterns use `\b`, Rust compiles it as a Unicode boundary, and regex's lazy DFA quits on the first non-ASCII byte in the haystack — so every file with an em-dash or emoji in a comment was scanned by the slowest engine once per rule whose keywords appeared.

## What changes (removing the shared cost, not compensating for it)

- `ExtractedCloneBodyV1` and `CloneBodyPayloadV1` hold token streams as `Arc<[ConservativeCloneTokenV1]>`; the payload shares the extracted body's stream instead of cloning it. Wire shape (a serde sequence) and digests are unchanged.
- `re2_compatible_regex` rewrites `\b`/`\B` to `(?-u:\b)`/`(?-u:\B)` — RE2's actual meaning (gitleaks is RE2), and DFA-eligible. `\b` inside a character class stays a backspace.

No allocator swap and no worker cap: both would mask the contention while leaving the copies in place.

## Evidence

- `tracedecay-privacy` full suite (incl. every catalogue rule compiling under the translation), `tracedecay-code-extraction` `main` tests, `tracedecay-code-index` `clones::` tests: green.
- Before/after sample on the same daemon will be attached after the beta carrying this ships.


ScriptedAlchemy:
### Cross-link from the graph-rebuild receipt investigation

This PR attacks the same shared cost our mid-run Mac sample saw on the #1562 journey (`b80dd58`, pid 11360): clone-token `String`/`Vec` churn and privacy regex work.

How it fits the current open set:

- **#1562** — seat-swap only (keep it that way). No file overlap with this PR.
- **#1567** — status-census / fingerprint wake on the wait loop. Orthogonal to Arc-sharing tokens; land independently.
- **#1568** — ranked receipt-speed plan (defer clone successor off the receipt, then admit cuts). #1570 is complementary: it shrinks CPU per body on both the first admit and any successor walk, even if successor stays off the receipt.

Suggestion: land **#1570** as its own follow-up (or ahead of admit-path micro-cuts). It does not conflict with #1562/#1567 file sets. Please attach the promised before/after daemon sample when you have it — that will tell us whether Arc+`(?-u:\b)` alone recovers enough of the ~4s receipt miss, or whether we still need #1568 item “no second `open_sealed_source`”.

No merge requested from this comment.

PR #1577, opened 2026-09-18
perf(clones): decode clone tokens in one pass; validate each payload once

ScriptedAlchemy:
Two measured, wire-preserving cuts to the clone-body codec that the
`background_refresh_and_reopen_report_only_servable_generations` journey drives
98,304 times per generation. These compose with **#1570** (merged) rather than
overlapping it: #1570 removed the *extraction-side* token copy by sharing
`Arc<[ConservativeCloneTokenV1]>` between an extracted body and its payload;
this removes the *decode-side* serde buffer and the *census-side* repeated
payload verification. The token element type is untouched by #1570, so the two
changes layer cleanly — the merge commit here resolves their only overlap (one
import block).

## 1. Clone tokens decode without serde's tagged buffer

`ConservativeCloneTokenV1` is internally tagged, so serde's derived
`Deserialize` buffers every token object into an owned
`serde::__private::de::Content` map before it can dispatch on `kind`. #1570's
`Arc<[_]>` sharing is extraction-side only: on the restore path every sealed
segment still decodes its own token objects, and a 768-file generated corpus
decodes 3.5 M of them per generation, plus again on every clone-census read and
clone-successor pass.

`Serialize` stays derived, so the wire form is byte-identical and no sealed
format revision moves. `Deserialize` is hand-written as a single-pass visitor
keeping every refusal the derive made: unknown member, duplicated member,
missing `kind`, unknown tag, and a member that does not belong to the tagged
variant. `clone_body_tokens.rs` covers the wire form, member-order
independence, and all ten refusals.

Measured on `pub fn refresh_probe_NNNN_NNN(input: u32) -> u32 { input + N }`
× 128 (one fixture file, 2,304 tokens), release profile, **both arms on top of
#1570**:

| | #1570 only | #1570 + this PR |
| --- | --- | --- |
| token stream decode | 658 µs/file (286 ns/token) | 543 µs/file (236 ns/token) |
| token stream encode | 150 µs/file | 146 µs/file (unchanged) |

## 2. The clone census validates each stored payload once

`clone_body_payloads` is keyed by payload digest, so one row backs every
occurrence that shares that body. `read_clone_index_census` joined the two
tables and called `CloneBodyPayloadV1::validate()` — four canonical SHA-256
passes over the token streams — once per *occurrence*. The generated corpus
stores 98,304 occurrences over 128 distinct payloads, so the census performed
98,304 validations to discharge 128 obligations: about 1.2 s of
single-threaded work, inside the `worktree_freshness` read that holds the
scheduler lock and that the journey polls in a spin loop.

The payload table is now validated once up front; only the rename coverages
the per-occurrence counters distinguish are retained, and occurrences are
classified against that verified index. Both prior refusals are preserved and
two are added:

- a payload whose blob `payload_digest` disagrees with its stored row key;
- an occurrence whose payload row is absent, which the inner join previously
  dropped from `source_bodies` silently.

## Reviewing this diff

Only three files are this change:

- `crates/tracedecay-code-extraction/src/clone_body.rs`
- `crates/tracedecay-code-extraction/tests/main/clone_body_tokens.rs`
- `crates/tracedecay-query/src/retrieval/lexical/projection/artifact/clone_census.rs`

The three `code_index_scheduler/registry*` files in the diff are inherited from
`b80dd58` — the tip of `cursor/seat-through-retryable-activation-f5c7`, which
this work started from and which is under review as #1562 (and extended by
#1567). They are not part of this change and disappear from the diff once #1562
lands.

## Verification (composed tree, `master` + this PR)

- `cargo test --release -p tracedecay-code-extraction --test main` — 585 passed, 0 failed (includes the new wire-form and refusal coverage).
- `cargo test --release -p tracedecay-code-index --lib` — 255 passed, 0 failed (includes `streaming_file_segment_bytes_match_the_value_encoder`, the segment byte-parity guard).
- `cargo test --release -p tracedecay-code-index --test code_index_suite -- partitioned sealed clone` — 41 passed, 0 failed (includes `partitioned_codec_has_stable_bytes_and_round_trips`, `sealed_restore_reencodes_identically_at_serial_and_parallel_widths`, `carried_forward_clone_bodies_admit_through_the_reused_sealed_segment`).
- `cargo test --release -p tracedecay-query --test search_quality_suite -- clone` — `v16_clone_payloads_are_content_addressed_and_postings_page` passed.
- `cargo clippy --release -p tracedecay-code-extraction -p tracedecay-query --all-targets` — clean.
- `cargo fmt --all -- --check` — the four reported diffs are in `code_index_generations/locking.rs` and `lifecycle_lease.rs`, untouched here and already drifted on `master`.
- Pre-#1570 run of `cargo test --release -p tracedecay-code-index-runtime --lib -- clone` gave 13 passed / 2 failed; both failures (`query_admission_serves_v14_while_clone_successor_is_pending`, `expired_source_proof_reschedules_pending_clone_backfill`) reproduced identically with the changed files reverted to `b80dd58`, so they are not introduced here.

No sealed generation format revision, segment digest, or clone payload digest
changes, and `RECEIPT_TIMEOUT` is untouched. Nothing merged.

<div><a href="https://cursor.com/agents/bc-10a8eb74-e61b-50a2-8a44-395c44fdfa02?cursor_ref=pr_footer&cursor_cta=open_in_web"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a>&nbsp;<a href="https://cursor.com/background-agent?bcId=bc-10a8eb74-e61b-50a2-8a44-395c44fdfa02&cursor_ref=pr_footer&cursor_cta=open_in_cursor"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;</div>


ScriptedAlchemy:
Superseded by #1882, which ports the measured clone-census commit (, 2.282 s to 0.310 s over 98,304 occurrences). The hand-written  commit () was dropped: 157 lines of byte-compatibility maintenance for a claimed 17% on one decode step that the census port already removes from its hottest path. The seat-through-retryable-activation fix shared by this branch is #1877.

ScriptedAlchemy:
Correction to the close note, which lost its code spans: superseded by #1882, which ports the measured clone-census commit `ea06d59df9` (2.282 s to 0.310 s over 98,304 occurrences). The hand-written `Deserialize` commit `a0eaa9300e` was dropped: 157 lines of byte-compatibility maintenance for a claimed 17% on one decode step that the census port already removes from its hottest path. The seat-through-retryable-activation fix shared by this branch is #1877.

PR #1580, opened 2026-09-18
perf(code-index): stop the readiness waits competing with the rebuild

ScriptedAlchemy:
Investigation of the graph-rebuild receipt's 90s timeout, focused on the status-polling wait. Stacked on `cursor/seat-through-retryable-activation-f5c7` (#1562); reviewing `master...` will show that branch's commits too.

## The 90s timeout was not waiting on anything

Reproduced locally (perf profile, 4 cores). At the moment the deadline fired:

| status field | value |
|---|---|
| `code_index_freshness.status` | `current` |
| `coverage` / `staleness_state` | `complete` / `fresh` |
| `source_reference` / `source_revision` | `refs/heads/main` / the expected refreshed revision |
| `latest_generation_id` | `generation.v1.51e0facb.00000002.3c00fba0…` |
| `code_index_freshness_warning` | absent |
| `retrieval_serving.seated_generation_age_seconds` | 89 |

Every conjunct on the status side held, on the right revision, for the whole 90 seconds. The search response carried that same `code_generation` — inside `preview`, because `limit: 3` rendered **18 084 characters against the 15 000-character response frame** (`tracedecay-mcp-catalog`), so MCP replaced the body with a retrieval handle:

```
["handle","original_chars","preview","preview_chars","retrieve_…","truncated"]
```

`results` and `code_generation` are absent at the top level. The predicate read them as missing, could not distinguish that from a warming generation, and spun to its deadline. `seated_generation_age_seconds ≈ 89` is exactly that: the seat was installed one second in and never moved — **the seat fix in #1562 worked, and the receipt could not see it.**

What blows the frame is ~5 KiB of ranking provenance per candidate, not result bodies, so `limit: 20 → 3` in 7bdc33d3 was a guess at how many candidates fit and landed short.

## Changes

1. **`test(tracedecay)`** — `daemon_suite/git_watch_test.rs` had already hit this wall and solved it properly, by paging the stored original back through `tracedecay_retrieve`. That resolver moves to the shared `tests/common/` surface and both suites read every tool answer through it, so page size stops being load-bearing. Both suites' duplicate `tool` / `tool_payload` helpers go with it (**−113/+16 lines**).
2. **`test(transport)`** — both waits paced at 25 ms instead of `yield_now`. The spin re-entered `tracedecay_status` thousands of times a second, and each call runs the generation census ready-probe, a scheduler freshness read, and branch diagnostics — 1–3 Git opens, 2 blocking-pool offloads, 2–6 registry mutex acquisitions — on the runtime running the reconcile it waits for.
3. **`perf(code-index)`** — `GitMetadataFingerprintV1::capture` re-derived the git-dir and common-dir through a fresh `gix::open` on every sample, against a doc comment calling that cost fixed and cheap. runtime-core already owns a revalidating topology memo for the same question. **Capture 75.2 µs → 10.0 µs**; this is the tier-1 ladder on every query admission, and search runs 2–3 captures per call.
4. **`perf(daemon)`** — the production composition's publication wait re-ran that probe every 10 ms for up to 20 s, leaving a pending arrival the worker yields its graph prepare to. The registry already publishes the edge it wants (`subscribe_serving_generation_changes`), so the wait parks on it, with a 100 ms floor for the terminal answers that install no seat.

## Measured

`graph_rebuild_status_test::background_refresh_and_reopen_report_only_servable_generations`, perf profile, 4 cores:

| | outcome | fails at |
|---|---|---|
| baseline | 93.4 s FAIL | status-polling deadline, line 176 |
| + untruncated read | 101.6 s FAIL | harness reopen budget, line 296 |
| + paced / event-driven waits | **95.9 s FAIL** | harness reopen budget, line 296 |
| same, harness budget scaffolded wide | **164.7 s → 155.3 s PASS** | — |

Fingerprint, one-ref fixture, 2000 warm iterations: capture **75.2 µs → 10.0 µs**; `gix::open` 72.7 µs; topology memo 2.0 µs.

Publication waits with the event-driven wait: 0.45 s initial, 50.1 s reopen (from 55.1 s), 14.5 s unchanged reopen (from 15.4 s).

**No timeout was raised.** The 90 s `RECEIPT_TIMEOUT` and the harness's 20 s budget are untouched; the wide budget above was investigation scaffolding and is not in the diff.

## What this does not fix

With the waits corrected, the journey's next binding constraint is the publication cost itself, and it is not a polling problem:

- **50 s to republish after adding one file** to a 769-file / 98 304-symbol tree, against a 20 s budget. State at timeout: `Indexing` / `PartialRefreshInProgress`, `clone_index: "the sealed lexical artifact is still building"` — progressing, just whole-corpus.
- **14.5 s for a completely unchanged reopen**, where recovery exists specifically to avoid replaying what already serves. Close enough to 20 s to be a flake source on a loaded runner.

That belongs to the rebuild-cost lane and the fix is not a wider budget. Until it lands this test still fails — but at a truthful, self-describing point rather than a 90 s deadline on an already-current generation.

## Also found, not changed here

- The same status-then-search spin exists in `daemon_suite/git_watch_test.rs:238` and `mcp_suite/.../graph_readiness.rs:42`; both compare `code_generation` across two separate tool calls.
- `tracedecay_status` computes the generation census unconditionally, including for `admission_only`, and `build_branch_diagnostics` resolves the live branch through a Git open plus a `branch-meta` read and several path stats on every call — none of it gated by the `include_*` flags.
- `serving_generation_after_activation_failure` (added in b80dd58, #1562) is dead outside `cfg(test)` and warns on every build.

## Verification

- `daemon_suite::git_watch_test` — 5 passed, confirming the shared resolver on the suite that owned it.
- `code_index_scheduler::identity::tests` — 5 passed, including both fingerprint change-detection tests and the in-place loose-ref rewrite.
- `code_index_scheduler::*` — 8 failures, **all pre-existing on #1562**: 7 of the 8 fail identically with this branch's `identity.rs` reverted to b80dd58, and the eighth (`concurrent_query_admissions_claim_one_pending_wake_before_worker_coalescing`) is flaky on base at 3/8 versus 1/5 here.
- `cargo fmt`, `cargo check --tests`, commitlint clean.

<div><a href="https://cursor.com/agents/bc-4cb06845-3a4d-5f7d-b18e-70d0d1b3b707?cursor_ref=pr_footer&cursor_cta=open_in_web"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a>&nbsp;<a href="https://cursor.com/background-agent?bcId=bc-4cb06845-3a4d-5f7d-b18e-70d0d1b3b707&cursor_ref=pr_footer&cursor_cta=open_in_cursor"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;</div>


ScriptedAlchemy:
#1882 ports `e4290b05e8` (git metadata through retained topology, 127.9 µs to 27.7 µs per capture). `7da32a15ae`, the event-driven wait in `wait_for_production_composition_code_index`, is not on master and stays owed here; its only measurement is the 165 s graph-rebuild journey suite. The shared seat fix is #1877.

PR #1797, opened 2026-09-19
fix: live-daemon retry storm, clone-scan stall, partial-generation zero, update identity

ScriptedAlchemy:
Root-cause fixes from Zack's live daemon (beta.46 of 77c55040d6, then beta.47) that ran eleven hours without sealing a code-index generation while printing 71k warnings, plus two defects found while hardening today's flaky tests. Each commit carries its evidence; summary:

- `cc10394232` fix(sessions): the session temporal refresh worker treated every storage-adapter error as retryable, so a SQLite constraint abort (`invalid session refresh progress`, code 19/1811) and `exact SQL query materialization exceeded its limit` were resubmitted about once a second per store for the whole process lifetime (`retry_attempt=25107`). Deterministic refusals are now typed and retire the refresh through the existing terminal-attempt machinery; the producer side also admitted an equal `committed_through` where the durable guard requires a strict advance, which is how the refused row was made. Tests in runtime-core, the worker and the store contract.
- `8b755233b8` perf(code-index): the thread dump of the stuck daemon showed two blocking-pool threads inside `verify_clone_fingerprint_page_rows` full-scanning `clone_exact_postings`/`clone_fingerprint_postings` (WITHOUT ROWID, keyed by class/language) with `WHERE symbol_occurrence_id = ?` twice per clone body, quadratic in the repository's postings. That is what pinned two cores at 100% and what the code-index mount was waiting on. Both tables are now read once per page and bucketed by occurrence; comparisons stay byte-identical.
- `740052a1ce` fix(sessions): `tracedecay_message_search` returned `complete_zero` from a generation whose relation receipt had not been applied yet (`Partial { generation_lag: 1 }`); the execution-error path mapped `Empty` to an authoritative zero. It now answers `partial` with the lag, which the MCP surface already renders and the LCM consumers already handle. 25/42 → 0/42 under six-way contention.
- `5e03dfa9c5` fix(daemon): `tracedecay update` failed its own readiness wait against the binary it had just installed ("protocol identity mismatch ... 0.1.0-beta.47+84598a0b… vs 0.1.0-beta.47") and every connection logged a version skew: the release path reports the bare tag, the daemon its build version, and three raw string comparisons treated a less specific identity as a different one. One canonical semver-aware comparison is now used at all three sites.

Left open, with evidence in the lane report: two ingest owners (live ingest and the project catch-up sweep) advance the same Claude observation cursor with legitimately different coverage reasons and the loser is refused as `observation_cursor_advance_collision` (3,042 + 638 warnings). The sanctioned no-op fix contradicts two contract tests whose names state the opposite intent, so it needs an owner decision: treat a same-range/different-reason advance as a no-op, or give each range one owner.


ScriptedAlchemy:
Runner-queue note (for anyone watching): the account allows 2 concurrent CI runs and the queue reached 18 dispatches, almost all on `cursor/*` branches whose bases predate this PR. I cancelled the queued (not running) cursor runs so this PR's run 35427443207 starts next; the cancelled branches are listed in my session and will be re-dispatched after they are merged with master, which they need anyway. Sorry for the churn; the alternative was a multi-hour wait for every merge.

PR #1794, opened 2026-09-19
test(ci): close three master-tip readiness races

ScriptedAlchemy:
## Summary

Rebased onto green master tip `9520d75f3612016653d90f667c57bd5a7514e454` (42 commits). Head is `30dbebab5e9f8c6600a0e2afdb48b71b2fe81556`. `.github/workflows/ci.yml` is untouched: no `pull_request` trigger, hotpath stays crates.io.

- HTTP, MCP, and SDK reset-required checks ride out a still-warming project open (`unavailable` / `after_delay` / `pre_admission`) for the same 15s grace the CLI already uses, then assert `reset_required`.
- Test daemon shutdown signals the child's process group so helper processes cannot keep the listen socket alive after the leader is reaped. Master already waits up to 10s for that socket to stop accepting.
- Packaged-host advisory ingest keeps polling through `accepted_for_replay`; only `committed` or `exact_duplicate` proves the transcript.
- Master's generation-store lock already closes the ignored-dependency coalesced-publication race. The extra background-admission hold this branch used to layer on top is gone, so `flight_tests.rs` matches master.

## Test plan

- [x] `cargo clippy -p tracedecay-cli --tests --locked -- -D warnings`
- [x] `cargo clippy -p tracedecay --features test-helpers --test runtime_acceptance_suite --test daemon_suite --locked -- -D warnings`
- [x] `cargo clippy -p tracedecay --features test-transport --test transport_acceptance_suite --locked -- -D warnings`
- [x] `cli_non_interactive_test::branch_search_serves_a_committed_generation_behind_dirty_worktree_state` — 1 passed
- [x] `advisory_runtime_acceptance::packaged_host_ingest_delivers_a_registered_advisory_cycle` — 1 passed
- [x] `code_index_ignored_dependencies_test::flight_tests::coalesced_publication_failure_preserves_the_scheduler_error_family` — 1 passed
- [x] `typed_terminal_restart_acceptance::transport_boundaries::reset_required_survives_http_mcp_and_rust_sdk_across_restart` — 1 passed
- [ ] Push CI does not run on this branch. Dispatch `CI` (`workflow_dispatch`) on `30dbebab5e9f8c6600a0e2afdb48b71b2fe81556`.

## Checklist

- [x] No secrets, credentials, or `.env` files included
- [ ] `CHANGELOG.md` updated (test harness only; no user-facing change)
- [ ] Breaking changes documented (none)

[...]

PR #1831, opened 2026-09-19
fix(daemon): drop listen sockets held by helper children

ScriptedAlchemy:
## Summary

- A daemon helper still between `fork` and `exec` no longer keeps the listen socket after the parent is killed: the child asks the kernel to die with that parent.
- The test stop signals the process group `process_group(0)` created, instead of only the leader pid.
- Did not absorb #1797 (unrelated live-daemon fixes) or re-land #1794's already-merged waits. The process-group stop from #1794 is the part that belongs here, and it is no longer the only closure.

## Motivation

Tip `3f71bccb2dba` still only waited for a predecessor socket to stop accepting. That wait treats the symptom. `fork` duplicates the listen descriptor; `FD_CLOEXEC` drops it at `exec`, not before. A pid-only `SIGKILL` of the leader leaves that child accepting. Local census: after the leader is killed, connect still succeeds until the process group is signalled.

## Changes

- `crates/tracedecay-runtime-core/src/process_tree.rs`: `arm_helper_parent_death` (`PR_SET_PDEATHSIG`) and `signal_process_group`.
- `run_foreground_unix` arms parent-death before any helper is forked.
- Test daemon teardown signals the harness-created process group, then reaps the leader.

## Test plan

- [x] `cargo test -p tracedecay-runtime-core --lib process_tree -- --test-threads=1` — 2 passed. The parent-death test fails closed if a pid-only kill leaves the socket up. The group test first shows the socket still accepts after a pid-only kill, then fails that connect after the group signal.
- [x] `cargo clippy -p tracedecay-runtime-core --lib --tests --offline -- -D warnings` — exit 0
- [x] `cargo check -p tracedecay --lib` — exit 0
- [ ] Full workspace nextest is the push CI lane, not this slice

## Checklist

- [ ] `CHANGELOG.md` updated (process teardown; no command or wire change)
- [x] Breaking changes documented (none)

SHA: `ffafc6fcc5f83be6531f40823662d1476d37a345`

<div><a href="https://cursor.com/agents/bc-d96d17e1-771e-5bfc-853c-39215bae501e?cursor_ref=pr_footer&cursor_cta=open_in_web"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a>&nbsp;<a href="https://cursor.com/background-agent?bcId=bc-d96d17e1-771e-5bfc-853c-39215bae501e&cursor_ref=pr_footer&cursor_cta=open_in_cursor"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;</div>


ScriptedAlchemy:
Review lane verdict: REJECT as a merge candidate for now. PR_SET_PDEATHSIG armed via pthread_atfork is Linux-only and the macOS CI lane builds the daemon bootstrap; the harness half (group SIGKILL in terminate_and_reap) is landing via #1833 in batch #1848. If the production root-cause half is still wanted, please rebase onto master after #1848 lands, drop the tests/common/mod.rs hunk, and gate the PDEATHSIG path on cfg(target_os = "linux") with a no-op elsewhere.

ScriptedAlchemy:
Closing after a second review with an experiment.

The production half cannot fire. `arm_helper_parent_death` installs `PR_SET_PDEATHSIG` from a `pthread_atfork` child handler, and glibc runs that handler only on the `fork` path. Rust's `Command` uses `posix_spawn` unless `pre_exec` (or uid, gid, chroot, setsid) is set, and no production spawn in this repo sets any of those. A probe against glibc 2.39, with the handler reporting through a pipe, ran the handler 0 times for a plain spawn and for a `process_group(0)` spawn, and 1 time only under `pre_exec`, which is the arrangement the PR's own test manufactures.

The window it targets is already closed. The listen socket is close-on-exec, so any child drops it at `exec`. The one long-lived production child, the work-attempt provider in `work_attempt_exec.rs`, deliberately runs in its own process group with `kill_on_drop`; a working parent-death signal would kill in-flight work attempts whenever the daemon exits. The change also turns a hook failure into a fatal `TraceDecayError::Config` on the daemon's startup route.

The harness half already landed via #1833 in #1848: `terminate_and_reap` sends the group SIGKILL and keeps the bounded vacate poll this diff would have removed.

If the predecessor-socket symptom recurs on a loaded runner, fork inheritance is not the mechanism, and it needs its own investigation.

PR #1865, opened 2026-09-19
fix: clear the deterministic master CI failures

ScriptedAlchemy:
## Summary
- The npm publication checker required a GitHub `release` trigger after the workflow moved to a master `workflow_dispatch`, so the policy test never mutated the file and the checker rejected the live workflow.
- Hold the reconcile worker before asserting that an expired source proof declines, so a fast renewal is not a fail-on-flake miss.
- Report non-contract clone-warmup faults as `generation_unverified` instead of a permanent `search_failed` while a generation is still retiring.
- Surface the remote-node provision error (status stays 409) so the SDK enrollment failure is no longer an empty conflict.

## Test plan
- [x] `python3 scripts/test-check-sdk-publish-workflow.py`
- [x] `python3 scripts/check-sdk-publish-workflow.py`
- [x] `cargo check -p tracedecay-code-index-runtime --lib`
- [ ] CI on this PR, especially SDK conformance `production-router` and Test Linux runtime/root-transport


ScriptedAlchemy:
Reviewed and synced onto master 140b1de929 (fast-forward pushes, no force):

- Kept: the npm publication checker fix (reproduced both the checker and its test failing on master) and the remote-node provision error body (route is behind local admission plus bearer auth; no test pinned the empty 409).
- Dropped as duplicate: the reconcile-worker fence, byte-identical to cbb13c08e2 from #1864; reconcile.rs now matches master.
- Reverted (ab0bd0652d): the clone-warmup catch-all. Widening Err(AuthorityUnavailable) to Err(_) would answer generation_unverified retryable=true for CapabilityManifestRejected, GenerationMismatch, IncompatibleProjection and Contract, which advance_text_serving_for_request marks permanently failed on the same call (serving.rs:2589), so callers would retry forever. The clone-lane refusal it targeted was the head-open race fixed in #1863.
- Added (656cca971c): test_rejects_restoring_the_release_trigger, since the new no-release-trigger rule had no test; verified it is the only failing test when the checker is weakened.

CI dispatched on 656cca971c; merging on green.

PR #1762, opened 2026-09-18
test(mcp): prove tracedecay_affected behavior

ScriptedAlchemy:
## Summary

- `tracedecay_affected` is the git-family MCP tool (`def_affected` → `dispatch_tool` → `handle_affected`) that walks reverse file dependencies and returns the tests that cover the changed files.
- The old `test_affected` check only looked for the words `affected_tests` and `count`. The traversal unit test pinned frontier batch shapes and a length. Both pass without observing which tests are returned. They now assert the literal sets, ranks, distances, depth cutoff, custom glob, missing `files` rejection, and an unpublished path.
- A changed path the index never published was failing inside adjacency (`code graph adjacency requires at least one seed`) instead of returning no neighbors. File-neighbor lookup now treats a symbol-less path as an empty neighbor set.

## Motivation

The tool's user-visible result is which tests to run. A test that only checks key names cannot fail when that set is wrong. The unpublished-path failure leaked an internal adjacency contract into the MCP response.

## Changes

- `crates/tracedecay/tests/mcp_suite/mcp_handler_test/graph_query_test.rs` calls the production MCP server the same way a host does (`tools/call` via `ProductionProjectCompositionHarnessV1`) and asserts literal JSON.
- `crates/tracedecay-mcp/src/handlers/git/affected.rs` unit tests assert hop distances and stable ranking instead of database-read batches.
- `crates/tracedecay-graph-query/src/queries.rs` returns an empty neighbor list when a logical path has no symbols.

## Test plan

- [x] `cargo test -p tracedecay-mcp --lib handlers::git::affected::tests` — 2 passed
- [x] `REQUIRE_EXACT_TEST_COUNT=3 scripts/require-exact-test.sh cargo test -p tracedecay --features test-transport --test mcp_suite affected -- --skip run_affected --test-threads=1` — 3 passed, 0 failed
- [ ] `cargo nextest run --workspace` not run (scoped to this tool)
- [ ] `CHANGELOG.md` not updated (test and a typed empty-neighbor correction, no new user-facing flag)

Observed `tools/call tracedecay_affected` results on the daemon fixture (`files: ["src/daemon.rs"]`):

- depth 5: `tests/daemon_direct_test.rs` (distance 1, direct), `tests/mcp_near_test.rs` (distance 2, near), `tests/serve_transitive_test.rs` (distance 3, transitive); recommended tests are the first two
- depth 1: only `tests/daemon_direct_test.rs`
[...]

PR #1877, opened 2026-09-20
fix(code-index): seat text through retryable graph activation

ScriptedAlchemy:
Text serving never took its seat while native graph activation kept failing retryably. The retryable arm in `mount.rs` rewrote the pass result to `Ok((Err(error), None, None))`, which fails the serving swap's own `Ok((Ok(_), Some(latest), _))` guard, so the swap never ran. The terminal-failure arm directly below already keeps the seat and marks the graph unavailable, so a retry was seating strictly less than a terminal verdict. Search held its predecessor while a complete generation sat on disk, which is the graph-seating starvation recorded on 2026-08-28 and observed again yesterday on a 22-minute scratch-daemon journey where the code graph never seated.

The test drives the real worker loop with an activation that keeps failing retryably and reads the real serving slot. On master it fails after 72 activation attempts and zero seats in 30 s; with the fix it seats on the first pass. This ports the intent of #1562 (b80dd58dfe) as a deletion rather than its always-true predicate. Its second half, coercing `Unfinished` to `Finished` when exact and lexical owners are ready, is a no-op against the producer that emits `Unfinished` exactly when those owners are not ready, and it stays out.

`file_neighbors` in the graph query now answers an empty vector for an unpublished path instead of letting `entity_ids` refuse an empty seed list as a contract violation that escaped as `code-graph-invalid-request`. The guard sits in the shared function because three callers reach it with unvalidated paths, and its siblings `incoming_edges` and `edges_among` already carry it. This ports the guard from #1762; that PR's deletion of the per-frontier read-count test is rejected, since the replacement asserts distances only and would miss an N+1 regression.

The convergence-park invariant tests pass. The code-index runtime lib suite shows pre-existing parallel-run contention failures shared with master at the same commit; all ten suspects pass serially on this branch.


ScriptedAlchemy:
Run 35487807710 went red on `failed_cold_mount_graph_replay_preserves_retained_text_generation` (FLKY-FL 2/2). That test asserted the complete serving slot stays empty after a permanently refused graph replay, which only held because a retryable activation failure starved the seat, the defect this branch fixes. The read-only publication error is a `GraphActivation` error with no conflict verdict, so it takes the retry arm on every attempt, and with this branch the sealed generation seats while native graph keeps retrying. The old assertion raced that seat.

Commit 6c9af77734 moves the contract onto the seated generation: it is the retained generation, its graph readiness is not `Ready`, and `interactive_graph_store()` is an error. All seven graph-activation tests pass under the root-lib lens and the root crate's clippy is clean. CI re-dispatched at 6c9af77734.

PR #1878, opened 2026-09-20
test(code-index): fence shutdown and observability reconcile tests

ScriptedAlchemy:
Two reconcile tests failed at CI shape, the literal nextest command inside a 4-CPU cgroup, at 4 failures in 16 iterations. Both were test-arrangement defects; production is unchanged.

`shutdown_signals_code_index_worker_without_taking_busy_scheduler_lock` asserted that shutdown returned within 250 ms while a test thread held the scheduler mutex. Shutdown never touches that mutex: it sets the shutting-down flag, wakes the worker, and joins it, and the worker polls `try_lock` under `lock_scheduler_unless_shutting_down`, so neither side can park behind a holder. The 250 ms measured the worker's cancellation-observation latency on a starved host (514 of the 515 ms in the capture were the worker join). The holder now parks on a release channel and the test asserts program order: shutdown returns before the release is sent while the holder is still parked. With the worker changed to a blocking lock, shutdown's join never returns and the test times out, so the arrangement detects the defect it names.

`blocked_observability_store_does_not_hold_reconcile_readiness` holds the store's only write transaction across a whole reconcile, then required the bounded observability producer to flush cleanly. Those cannot both hold: the producer's batch write blocks behind the transaction, misses its 2 s persistence budget, drops the batch, and latches that denial for the producer's life. That is the intended design, telemetry is denied and the product path is not, which is what the readiness assertion proves. Teardown now accepts `Ok` or exactly the persistence-deadline domain; every other producer fault still panics, and a mutation reporting the denial under another domain fails.

After the change, 20 CI-shaped iterations pass for both tests. Under this box's ambient lane load, eight other reconcile tests with fixed wall-clock budgets still fail intermittently; they are listed in the report and untouched here.


ScriptedAlchemy:
Run 35487831813 went red on `a_disproving_exact_source_probe_withdraws_the_busy_read_witness` (FLKY-FL 2/2), a test this PR did not touch. The witness read `None` while the admission permit was held and the pass counter had settled, which is the graph-tail gap already documented on `hold_scheduler_for_root` and fenced for its two siblings. Commit 1a3c163c34 holds the scheduler mutex from before the drift until the read path has answered.

It did not reproduce locally in eight quota-scoped nextest runs (two-core loop 0/60, reconcile module 0/3, whole runtime library 0/5 at CPUQuota=400%). The fenced module passes 3/3 under the same recipe and package clippy is clean. CI re-dispatched at 1a3c163c34.

PR #1885, opened 2026-09-20
test(sessions): keep git auto-maintenance out of the fixture

ScriptedAlchemy:
`provenance_capture_copies_no_git_objects` went red on PR #1877's run (FLKY-FL 2/2 in the core-storage partition) with `.git/objects/maintenance.lock` present in the before fingerprint and absent in the after one. Git 2.47+ detaches `maintenance run --auto` after a commit and holds that lock while it runs, so the fixture's own commit, not the capture, moved the object store. This is the mechanism #1869 fixed for the runtime-core fixture; the sessions fixture now disables auto maintenance and gc for every git call the same way.

The 20 provenance tests pass locally and the crate's clippy is clean.


ScriptedAlchemy:
Folded into #1886 as its second commit so one CI cycle covers both test-fixture fixes; this branch's own base was the broken master head.

PR #1931, opened 2026-09-21
chore(deps): move gix to 0.87.1 so bisync leaves the tree

ScriptedAlchemy:
Every published `bisync` release is yanked. The workspace reached it through `gix = "=0.86.0"`, whose `gix-protocol 0.64.0` hard-requires `bisync ^0.3.0`. That is why a downstream crate could not resolve our published dependency set offline, and why `scripts/check-distribution-acceptance.sh` had to hand its fresh consumer the workspace lockfile.

`gix-protocol 0.65.1` dropped `bisync` outright, so moving to `gix = "=0.87.1"` removes the only edge that reached it.

```
$ cargo tree -i bisync --workspace
error: package ID specification `bisync` did not match any packages
```

The lockfile loses both `bisync` and `bisync_macros`; 45 `gix-*` packages move with the pin.

## What changed

The pin lives in nine manifests rather than one workspace entry, each with its own feature list, so all nine move together and keep the features they had. No other exact pin moved. The `grafeo-* = "=0.5.42"` pins are unrelated and untouched.

One API change needed adapting. gix 0.87 adds `PartialEq<str>`, `PartialEq<&str>` and `PartialEq<String>` for `Id` and `oid` (`gix/src/id.rs` `impl_partial_eq_text!`, delegating to `oid::eq_str`). That makes the existing `id.to_string() != text` comparisons allocate for nothing and trips `clippy::cmp_owned`. Both sides compare the same full lowercase hex, so dropping the `to_string()` preserves behavior. Three sites: `candidate_output.rs:563`, `git_tree_capture.rs:336` and `:462`.

The consumer-side lockfile seed is removed. It existed only for the yank, and it hid the property that step exists to prove. The extracted-package seed higher up stays, because its reason is independent: each extracted archive is its own Cargo root that must build against the release resolution.

## Behavior change that needs a decision

gix 0.87 widened which environment variables a default-permission `gix::open` / `gix::discover` honors. `Options::default()` gained `use_repository_local_environment: true`, and `GIT_INDEX_FILE` is now a config key with an environment override. `Repository::index_path()` changed from computing `git_dir/index` to returning a value resolved at open time:

```rust
// 0.86.0
pub fn index_path(&self) -> PathBuf { self.git_dir().join("index") }
// 0.87.1
pub fn index_path(&self) -> PathBuf { self.index_path.clone() }
```

`GIT_INDEX_FILE` appears nowhere in the 0.86.0 source. `GIT_WORK_TREE` was likewise discarded in 0.86.0, because env-sourced config sections carry `path: None` and the worktree filter required a path under `git_dir`; 0.87 accepts them directly. `open/permissions.rs` is byte-identical between the two versions, so `Permissions::secure()` always allowed `git_prefix`. What changed is what that unlocks.

This is reachable. `repository_open_options()` in `crates/tracedecay-runtime-core/src/git_repository.rs:960` denies `git_prefix` and documents the intent ("rejecting `GIT_*` redirection from the daemon environment"), but it guards a single `gix::open_opts` call site. The other ~48 `gix::open` / `gix::discover` call sites use default permissions. TraceDecay also ships an optional global git `post-commit` hook that runs `tracedecay sync` (`crates/tracedecay-agent-hosts/src/agents/git_post_commit_hook.rs`), and git sets `GIT_INDEX_FILE` and `GIT_DIR` for hook processes.

Nothing in this PR changes that, deliberately. Deciding which call sites should reject `GIT_*` redirection is a threat-model call, not a mechanical port, so it is reported rather than approximated.

## Gates

| Gate | Result |
|---|---|
| `cargo build --workspace --locked` | pass, 9m18s |
| `cargo clippy --workspace --all-targets --locked -- -D warnings` | pass |
| `cargo fmt --all -- --check` | pass |
| `cargo test -p tracedecay-code-index --lib` | 261 passed, 0 failed |
| `cargo test -p tracedecay-code-index-runtime --lib --features test-helpers,test-transport -- --test-threads=4` | 504 passed, 0 failed, 2 ignored |
| `cargo test --lib` for the other touched crates | query 100, code-index-retention 266 (+1 ignored), source-edit 86, sessions 451, runtime-core 575, agent-hosts 512, application 468, all passed |
| `cargo test -p tracedecay --test mcp_suite --features tracedecay/test-transport` | 565 of 571 passed; 6 lane-environment failures, detailed below |

The mcp_suite run was split into three chunks to fit the lane's per-call time budget: 122 non-handler tests (pass), 109 `graph_analysis`/`graph_query` handler tests (pass), and the remaining 340 handler tests (334 pass, 6 fail).

All six failing files contain zero `gix` references, and each passes in isolation:

- `affected_tests_behavior_test` and the four `feedback_*` tests spawn `cargo`/`rustc` inside a sandboxed `HOME`, where rustup has no default toolchain: `error: rustup could not choose a version of cargo to run`. The five pass single-threaded in this lane.
- `retrieve_truncation_test` asserts on MCP response pagination and passes 8/8 in isolation.

`ci.yml` and `distribution-acceptance.yml` are dispatched on this branch; the clean-environment run is the authoritative result. The second workflow is the one that proves the fresh-consumer path now works without the seed.


ScriptedAlchemy:
Superseded by #1939 (gix-open single authority + gix 0.87.1). Merged as 7c8ad164c8291eec605d1d1e3dba10195c85d477.

